PR #46, opened 2026-03-29
Fix base table-create dry-run follow-ups (#7)

### LINKED ISSUE #7: Base UX: unstable command abstraction and delayed state visibility
## Summary
There are two related UX/DX issues in the Base (Bitable) workflow:

1. Some high-level Base commands expose options that do not appear to take effect as expected
2. State visibility can lag behind a successful mutation, which makes the result feel inconsistent

## 1. Unstable command abstraction
When using the high-level Base commands, some flags suggest richer abstraction than what is actually executed.

### Example
`lark-cli base +table-create --help` shows options like:
- `--fields`
- `--view`

However, in `--dry-run`, those options did not appear in the generated request body. The request only contained the table name.

This creates confusion because the command surface implies that table + fields + views can be created in one step, but the actual request does not seem to include them.

A similar issue exists around `+field-create`, where the expected JSON shape is not obvious without trial and error. For example, it was unclear whether the payload should use:
- `field_name` vs `name`
- enum-like numeric field types vs string discriminators like `text`, `number`, `datetime`

The current experience makes the abstraction feel unstable: the CLI looks high-level, but users still need low-level experimentation to learn the real contract.

## 2. Delayed state visibility
After a successful mutation, list/read operations can briefly return stale-looking results.

### Example
After deleting a table successfully, a subsequent `table-list` still briefly showed the deleted table before a later refresh reflected the final state.

This may be normal eventual consistency from the backend, but from the CLI user perspective it is confusing because it looks like the delete may not have worked.

## Why this matters
These two issues combine badly in real usage:
- users already need to experiment to understand what commands actually do
- after they perform a mutation, the immediate follow-up read may not reflect it yet

Together, this reduces trust in the command layer.

### For command abstraction
- Ensure `--help`, `--dry-run`, and actual execution behavior match exactly
- If `--fields` / `--view` are not implemented for `+table-create`, hide or mark them clearly
- Provide one concrete JSON example for each Base mutation command
- Standardize field payload conventions, especially for `+field-create`

### For delayed state visibility
- If Base APIs are eventually consistent, mention that explicitly in mutation success output
- Optionally provide a `wait until visible` mode for destructive/structural operations like create/delete table
- At minimum, improve hints so users know a short delay is expected after some mutations

## Real-world context
These issues showed up while trying to use `lark-cli` to:
- create a Base
- create warehouse-management tables
- add fields
- delete a temporary test table

The workflow was achievable, but required more trial and trust-rebuilding than expected because of the abstraction mismatch and short-lived stale reads.

zfaustk:
## Summary
- expand `+table-create --dry-run` so it previews follow-up `--fields` operations
- preview `--view` follow-up calls in the same dry-run chain
- keep the primary table-create preview visible even when follow-up preview parsing fails

Fixes #7.

### DIFF
--- a/shortcuts/base/base_dryrun_ops_test.go
+++ b/shortcuts/base/base_dryrun_ops_test.go
@@ -32,6 +32,52 @@ func TestDryRunTableOps(t *testing.T) {
 	assertDryRunContains(t, dryRunTableDelete(ctx, rt), "DELETE /open-apis/base/v3/bases/app_x/tables/tbl_1")
 }
 
+func TestDryRunTableCreateIncludesFieldAndViewFollowUps(t *testing.T) {
+	ctx := context.Background()
+	rt := newBaseTestRuntime(
+		map[string]string{
+			"base-token": "app_x",
+			"name":       "Orders",
+			"fields":     `[{"name":"Order ID","type":"text"},{"name":"Amount","type":"number"}]`,
+			"view":       `[{"name":"Main","type":"grid"}]`,
+		},
+		nil,
+		nil,
+	)
+
+	assertDryRunContains(t,
+		dryRunTableCreate(ctx, rt),
+		"# create table",
+		`{"name":"Orders"}`,
+		"PUT /open-apis/base/v3/bases/app_x/tables/%3Ccreated-table-id%3E/fields/%3Cprimary-field-id%3E",
+		`{"name":"Order ID","type":"text"}`,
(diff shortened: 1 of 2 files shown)


zfaustk:
Closing this as superseded by the current main implementation. The Base table-create and dry-run paths have since been reworked, the linked issue is closed, and this branch now conflicts with main.

PR #50, opened 2026-03-29
fix: align help examples and inject output error writer

### LINKED ISSUE #48: Help and README examples drift from the current CLI surface
`cmd/root.go` still advertises `lark-cli calendar events list`, but the current command surface does not expose `events list`. The bilingual README examples also still use `--body` for raw API calls, while `cmd/api/api.go` only accepts `--data`.

This is small but user-facing drift: the first examples people copy are wrong.

Suggested fix:
- replace the root example with `calendar events instance_view`
- replace `--body` with `--data` in `README.md` and `README.zh.md`

### LINKED ISSUE #49: output helpers should not write failures directly to os.Stderr
`internal/output/print.go` writes JSON/NDJSON marshal failures directly to `os.Stderr`, and `internal/output/csv.go` does the same for CSV flush failures. That bypasses injected IO streams and makes embedding, tests, and agent-side output capture less clean than the rest of the CLI.

Suggested fix: route output helper failures through a package-level injectable error sink, and add focused tests that verify the error path can be redirected.

zfaustk:
## Summary
- align the root help example with the current calendar command surface by replacing the nonexistent `calendar events list` example
- update the English and Chinese README raw API examples to use the actual `--data` flag exposed by `cmd/api/api.go`
- route output helper failures through an injectable package-level error writer and add focused tests for JSON, NDJSON, and CSV error paths

## Why
The current compare includes two small but user-facing inconsistencies: copied examples can fail immediately, and `internal/output` still writes failures directly to `os.Stderr`, which bypasses injected IO streams and makes embedding and output capture harder than the rest of the CLI.

## Testing
- added targeted tests in `internal/output/print_test.go`
- local `go test` could not be completed in this environment because the temporary toolchain fails stdlib resolution before package compilation (`package strings is not in std`)
- static verification completed for the command example and `--data` flag usage

Closes #48
Closes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ Call any Lark Open Platform endpoint directly, covering 2500+ APIs.
 
 ```bash
 lark-cli api GET /open-apis/calendar/v4/calendars
-lark-cli api POST /open-apis/im/v1/messages --params '{"receive_id_type":"chat_id"}' --body '{"receive_id":"oc_xxx","msg_type":"text","content":"{\"text\":\"Hello\"}"}'
+lark-cli api POST /open-apis/im/v1/messages --params '{"receive_id_type":"chat_id"}' --data '{"receive_id":"oc_xxx","msg_type":"text","content":"{\"text\":\"Hello\"}"}'
 ```
 
 ## Advanced Usage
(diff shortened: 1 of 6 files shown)


zfaustk:
Closing this as superseded. Both linked issues are closed, the current main output architecture has evolved past this patch, and this branch now conflicts with main.

PR #59, opened 2026-03-29
feat(docs): auto-resize table column widths based on content

SoIomon:
## Summary
This PR auto-resizes table column widths after `docs +create` and `docs +update` so generated tables better match their content instead of defaulting to equal-width columns.

It has been rebased onto the latest `main`, and the `docs +create` path keeps the current upstream result normalization behavior while applying the column-width post-processing.

## Changes
- Add table auto-width post-processing for `docs +create` and `docs +update`
- Calculate display width from parsed table cell content, counting CJK/fullwidth characters as 2 units and ASCII as 1
- Clamp per-column widths to `[80px, 400px]` and normalize the total width to the document container
- Update each column via separate PATCH requests because multiple column-width ops on the same table are not applied reliably in one batch request
- Skip unnecessary resize work when `docs +update` input does not contain a table, and degrade to a warning if the resize step fails
- Add tests covering width calculation, normalization, markdown table detection, and mixed CJK/ASCII content handling

## Test Plan
- [x] `go test ./shortcuts/doc/...`
- [x] `go build ./...`
- [x] Manual verification in Feishu docs with mixed CJK/ASCII table content

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * Automatic table column width adjustment on document creation and updates when tables are detected (skips tables inside fenced code blocks). Resize warnings are shown to users.

* **Chores**
  * Improved resolution and validation of document references (including wiki-sourced documents) with clearer diagnostic messages.

* **Behavior**
  * Table auto-resize omitted in delete-range mode; heuristics added to detect tables in markdown/HTML.

* **Tests**
  * Added comprehensive tests for detection, stripping fenced code, width computation, normalization, and skip logic.


SoIomon:
@zgz2048 @liangshuo-1 conflicts are resolved and checks have passed. Please help review when you have time.

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+	"unicode/utf8"
+
+	"github.com/larksuite/cli/internal/util"
+	"github.com/larksuite/cli/internal/validate"
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+const (
+	blockTypeTable = 31
+	blockTypeText  = 2
```
fangshuyu-768, line 21: **`blockTypeText` is declared but never referenced.**

`grep` only finds this definition; `blockTextWidth` below uses the literal `2` (along with `3, 4, 5, 6, 7, 8, 9, 12, 13, 17`) directly in the switch at L239. Either delete this constant or use it in the switch and add siblings for the other block types (see next comment).

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+	return maxWidth
+}
+
+// blockTextWidth calculates the display width of text in a block.
+// Chinese/fullwidth characters count as 2 units, ASCII as 1.
+func blockTextWidth(block map[string]interface{}) int {
+	blockType, _ := util.ToFloat64(block["block_type"])
+	switch int(blockType) {
+	case 2, 3, 4, 5, 6, 7, 8, 9, 12, 13, 17: // text, heading1-7, bullet, ordered, todo
```
fangshuyu-768, line 294: **Magic numbers in the supported-block-type list.**

`case 2, 3, 4, 5, 6, 7, 8, 9, 12, 13, 17` relies on a trailing comment (`// text, heading1-7, bullet, ordered, todo`) to stay correct. If a new block type (e.g. quote, callout) gains rich inline text in the future, nothing here forces a maintainer to consider it.

Suggest extracting named constants (`blockTypeText = 2`, `blockTypeHeading1..7 = 3..9`, `blockTypeBullet = 12`, `blockTypeOrdered = 13`, `blockTypeTodo = 17`) — or a small `supportedTextBlockTypes` set — so the intent is self-documenting and collisions with the docx type catalog are easier to audit. This also retires the orphaned `blockTypeText` constant at L18.

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+		// Scale proportionally without reapplying the minimum width so the total fits.
+		scale2 := float64(docContainerWidth) / float64(total2)
+		for i := range pixelWidths {
+			scaled := int(float64(pixelWidths[i]) * scale2)
+			if scaled < 1 {
+				scaled = 1
+			}
+			pixelWidths[i] = scaled
+		}
+	}
```
fangshuyu-768, line 437: **Degenerate second-pass can produce 1-pixel columns.**

For tables wider than ~8 columns, all widths hit `minColumnWidth = 80`, the total (e.g. 9×80 = 720) still exceeds `docContainerWidth = 700`, and the second pass scales proportionally with only a `scaled < 1` floor. The `scaled = 1` fallback will happen silently for pathological inputs and produce a table that is technically valid but visually unusable.

Options:
- Bail out (return the original `currentWidths`, so no PATCH is issued) when `colSize * minColumnWidth > docContainerWidth` — at that point the equal-width default the docx server already applies is strictly better than what we can compute.
- Or widen the minimum to something like `max(1, docContainerWidth / colSize)` on the second pass so every column remains readable.

Either way, the current behaviour should not ship unguarded.

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+	reminderFallbackWidth    = 10
+	equationFallbackWidth    = 8
+	linkPreviewFallbackWidth = 12
+	maxBlockPageFetches      = 100
+)
+
+// autoResizeTableColumns fetches all blocks from a document, finds table blocks,
+// calculates optimal column widths based on cell content, and updates via API.
+// Errors are non-fatal: returns a warning message or empty string on success.
+func autoResizeTableColumns(runtime *common.RuntimeContext, documentID string) string {
```
fangshuyu-768, line 46: **No context cancellation check during multi-call resize.**

`autoResizeTableColumns` issues 1 paginated `GET /blocks` + N PATCH calls per table (and runs over all tables in the document). `runtime.CallAPI` presumably honours the runtime context, but the loops themselves never check `runtime.Ctx().Done()`, so a Ctrl-C in the middle of a wide table will keep firing PATCH requests until the underlying HTTP client notices.

Since this is best-effort post-processing, an early `select { case <-ctx.Done(): return "" default: }` at the top of each iteration (both the table loop and the per-column PATCH loop in `resizeOneTable`) would let users abort cleanly without leaving half-resized tables behind.

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+)
+
+const (
+	blockTypeTable = 31
+	blockTypeText  = 2
+
+	minColumnWidth    = 80
+	maxColumnWidth    = 400
+	docContainerWidth = 700
+	charUnitWidth     = 12 // approximate pixel width per character unit
```
fangshuyu-768: **Unexplained magic constants for width math.**

`charUnitWidth = 12`, `cellPadding = 16`, `docContainerWidth = 700`, and the per-element fallback widths below are presented without provenance. A future maintainer hitting "why 12?" has no way to answer it without re-measuring the Feishu docx renderer.

One short comment per constant (or one block comment) noting how the value was derived (e.g. "measured against default docx table font at 14px/line-height 1.5" or a link to a design spec) would make it easy to revisit when the renderer font changes or when we add iPad / small-screen container widths.

shortcuts/doc/docs_update.go
```diff
@@ -112,18 +113,38 @@ var DocsUpdate = common.Shortcut{
 		}
 
 		normalizeDocsUpdateResult(result, runtime.Str("markdown"))
+
+		if shouldAutoResizeAfterUpdate(runtime.Str("mode"), runtime.Str("markdown")) {
+			docID := common.GetString(result, "doc_id")
+			if docID == "" {
+				resolvedDocID, resolveErr := resolveDocxDocumentID(runtime, runtime.Str("doc"), "table auto-width")
```
fangshuyu-768: **`operation: "table auto-width"` leaks into user-facing validation errors.**

`resolveDocxDocumentID` uses its `operation` argument to build messages like `"%s only supports docx documents; use a docx token/URL or a wiki URL that resolves to docx"`. Passing `"table auto-width"` here means a user running `docs +update` against a legacy `/doc/...` URL will see the error string *"table auto-width only supports docx documents..."*, which is confusing — the user didn't ask for auto-width, they asked for an update.

Two ways out: either pass `"docs +update"` here and swallow/skip quietly when the kind isn't docx (since this is a best-effort post-processing step), or catch the typed error and rewrite the message at this call-site before printing.

Related: the `docs +media-insert` call at `doc_media_insert.go:96` passes `"docs +media-insert"`, which reads much better for end users.

shortcuts/doc/docs_update.go
```diff
@@ -112,18 +113,38 @@ var DocsUpdate = common.Shortcut{
...
+				} else {
+					docID = resolvedDocID
+				}
+			}
+			if docID != "" {
+				if warn := autoResizeTableColumns(runtime, docID); warn != "" {
+					fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", warn)
+				}
+			}
+		}
```
fangshuyu-768, line 132: **Possible read-after-write race against the docx backend.**

`autoResizeTableColumns` fetches `/blocks` immediately after `CallMCPTool("update-doc", ...)` returns. If the backend is eventually consistent (block index / cell tree still propagating), the pagination can miss the freshly written table block and `autoResizeTableColumns` silently returns without warning — the user then sees the equal-width default and assumes auto-resize is broken.

Probably not common in practice, but worth either: (a) a single short retry when the read returns zero table blocks while the markdown was detected to contain one (`markdownLikelyContainsTable == true` but `len(tableBlocks) == 0`), or (b) at minimum a warning in that specific case so the user knows what happened.

Same concern applies to the `docs +create` path in `docs_create.go`.

shortcuts/doc/table_auto_width.go
```diff
@@ -0,0 +1,445 @@
...
+	// Convert character widths to pixel widths with constraints
+	columnWidths := computePixelWidths(colMaxWidths, colSize)
+	currentWidths := tableColumnWidths(prop)
+	if shouldSkipTableWidthUpdate(currentWidths, columnWidths, colSize) {
+		return ""
+	}
+
+	// Update table column widths — one PATCH per column because batch_update
+	// silently ignores multiple update_table_property ops on the same table
+	// in a single request.
```
fangshuyu-768: **Sequential PATCH cost is quietly linear in column count.**

The in-code comment ("batch_update silently ignores multiple update_table_property ops on the same table in a single request") is a good explanation of *why* one request per column. Worth adding a one-line note about *cost*: for a 10-column table, this adds 10 × round-trip latency on top of every `docs +create`/`+update`, and non-fatal failures just warn. Users running these against slow or rate-limited endpoints should know.

If the cost turns out to matter, limited concurrency via `errgroup` with a small worker pool (say 3–4) would usually be safe per-table — the constraint that "batch_update can't have multiple ops on the same table" doesn't preclude multiple *separate* PATCH requests landing in parallel. Not asking for that change here, just noting the option.

shortcuts/doc/docs_create.go
```diff
@@ -16,7 +17,7 @@ var DocsCreate = common.Shortcut{
 	Description: "Create a Lark document",
 	Risk:        "write",
 	AuthTypes:   []string{"user", "bot"},
-	Scopes:      []string{"docx:document:create"},
+	Scopes:      []string{"docx:document:create", "docx:document:readonly", "docx:document:write_only"},
```
fangshuyu-768: **Scope expansion is a silent breaking change for existing tokens.**

Adding `docx:document:readonly` and `docx:document:write_only` to `docs +create` means tokens that previously worked (only `docx:document:create` granted) will now fail the scope check once this PR lands — even for users whose markdown contains no tables and therefore wouldn't exercise the new code path at all.

Options, in decreasing order of safety:
- Call `autoResizeTableColumns` conditionally based on whether the runtime already has these scopes, so missing-scope becomes "skip auto-resize, print a warning" rather than "`create` fails outright".
- Gate this behind a `--no-auto-width` flag (default on) so users without the extra scopes can opt out explicitly.
- At minimum, call this out in release notes / CHANGELOG so operators re-run the scope-grant step before upgrading.

PR body already mentions the scopes are required — but end users upgrading via `go install` won't read the PR body.

fangshuyu-768 (commented):
Notes from reading the latest head — most of the earlier CodeRabbit/greptile findings (pagination exit, fenced-code stripping, `--|--` separator, out-of-bounds guard, create-path table heuristic, equal-width bypass) have been addressed in this revision, so these are additional comments beyond that. Roughly ordered by impact: correctness first, then code health / UX.

SoIomon:
Merged the latest \ into this branch and resolved the remaining conflict in \.\n\nRe-verified after the merge with:\n- \ok  	github.com/larksuite/cli/shortcuts/doc	(cached)\n- \\n\nI also re-reviewed the branch after the merge. The recent review feedback around table auto-width follow-up is addressed, and I do not see a new blocker from the conflict resolution itself.

SoIomon:
Merged the latest main into this branch and resolved the remaining conflict in shortcuts/doc/doc_media_insert.go.

Re-verified after the merge with:
- go test ./shortcuts/doc/...
- go build ./...

I also re-reviewed the branch after the merge. The recent review feedback around table auto-width follow-up is addressed, and I do not see a new blocker from the conflict resolution itself.

SunPeiYang996:
Thanks everyone for the careful review! All of the issues raised above (including the magic-number block-type list, the >8-column degenerate case, missing context cancellation, the misleading `operation` string in `resolveDocxDocumentID`, the read-after-write race, and the silent scope expansion) have been addressed in the latest revision on this branch.

If you are tracking this from a downstream consumer, please update `lark-cli` to the latest version to pick up the new algorithm.

## Algorithm in the latest revision

### Core idea

The column-width problem is now modeled as **constrained resource allocation**:

- **Resource**: usable page width = `pageWidth × fillRatio`
- **Subject**: each column, estimated with three width tiers — `min` / `pref` / `max`
- **Policy**: satisfy every column's `min` first, fit natural width via `pref`, then decide whether to expand to fill based on the remaining budget

Columns are also treated **differentially by content type** — text columns lean toward expansion, numeric / boolean columns stay compact.

### Three-tier width estimation

For every column we scan all cells and produce:

- **min** — longest unbreakable token + cell padding, so we never break words.
- **pref** — bucketed quantile over per-row content length, weighted by the **coefficient of variation (CV)** of the distribution; the header is folded in with a lower weight so a long header alone cannot dominate a column of short cells.
- **max** — fixed cap as a fraction of page width, then clamped further by column type.

A glyph-width safety factor is applied to compensate for the systematic under-estimation of the in-house width oracle.

### Five-pass pipeline (pseudocode)

```
function LayoutColumns(table, pageWidth):
    if table.hasExplicitColWidth():
        return                              # respect existing widths

    budget := pageWidth * fillRatio
    cols   := []
    for c in table.columns:
        cols[c] := estimate(c)              # min/pref/max + column type

    # Pass 1: shrink if sum(pref) > budget
    while sum(pref) > budget:
        i := argmax(pref)
        pref[i] := nextLargerNeighbor(pref) # demote widest column to the next tier

    # Pass 2: expand if sum(pref) < budget
    if mode == ExpandToBudget:
        distribute(budget - sum(pref)) by columnTypeWeight
    else:
        pref[i] *= softExpandFactor         # gentle expansion, do not force-fill

    # Pass 3: enforce min (iterate to a fixed point)
    repeat N times:
        for i where pref[i] < min[i]:
            pref[i] := min[i]
        redistribute remaining budget

    # Pass 4: smooth neighbours
    for adjacent pair (i, j) where ratio(i, j) > threshold:
        borrow from wider to narrower

    # Pass 5: round + clamp
    for each col:
        w[i] := ceil(pref[i])
        w[i] := clamp(w[i], hardMin, max[i])

    return w
```

### Column-type differentiation

Different column semantics get different `min` / `max` / expansion weights:

| Type         | Characteristics                  | Strategy                          |
| ------------ | -------------------------------- | --------------------------------- |
| Text         | Large length variance, breakable | High weight, biased to expand     |
| Code         | Unbreakable, monospaced          | Medium weight, slightly wider max |
| Numeric      | Short, uniform                   | Low weight, conservative expand   |
| Boolean/Enum | Very short                       | Excluded from expansion           |

### Edge cases

- **Two-column table with a dominant column** — when one column's preferred width is much larger than the other, its `max` is relaxed so a long description is not hard-capped.
- **Very wide tables** — when even `min × N` exceeds the budget, the total-width constraint is dropped and every column gets its `min`; horizontal scrolling is left to the front-end.
- **Streaming / incomplete input** — column-width computation is skipped until the table structure stabilizes.

### Differences from traditional approaches

- Traditional `table-layout: auto` defers to the browser and is inconsistent across renderers; this algorithm freezes widths **at generation time** so the result is WYSIWYG.
- Traditional `width=N%` forces the table to fill the container, which makes single columns absurdly wide in low-column tables; here we allow whitespace and only expand when the content actually warrants it.
- The **column-type weight** prevents numeric columns from being inflated into wide columns by the redistribution step.

SunPeiYang996:
Closing — the new table column-width algorithm described above has already shipped to `main` via a separate change and is available in the latest `lark-cli` release. Thanks to everyone who reviewed!

PR #111, opened 2026-03-30
fix(config): distinguish ENOENT from other config load errors in show/remove

### LINKED ISSUE #110: bug: config show/remove swallows real errors (permission denied, corrupt file) as "not configured yet"
## Summary

`lark-cli config show` and `lark-cli config remove` treat **all** `LoadMultiAppConfig` failures — including permission denied and JSON parse errors — as "not configured yet". This hides the real problem and can mislead users into running `config init`, potentially overwriting a valid but temporarily unreadable config file.

## Affected Files

- `cmd/config/show.go:42-46`
- `cmd/config/remove.go:42-44`

## Root Cause

`LoadMultiAppConfig` can fail for three distinct reasons, but the callers don't distinguish them:

| Error | Root cause | Current message | Correct message |
|-------|-----------|-----------------|-----------------|
| `ENOENT` | Config file doesn't exist | "Not configured yet" | ✅ Correct |
| `EACCES` | Permission denied on config file | "Not configured yet" | ❌ Should show actual error |
| JSON parse error | Config file is corrupted | "Not configured yet" | ❌ Should show actual error |

## Impact

**`config show`** exits with code **0** on a permission/parse error, printing "Not configured yet". A user or script has no way to detect the real problem. Worse, the user may run `config init` and overwrite a valid but temporarily unreadable config.

**`config remove`** returns a non-zero exit, but still reports "not configured yet" instead of the real error, making diagnosis harder.

## Suggested Fix

Use `errors.Is(err, fs.ErrNotExist)` to distinguish "file not found" from other failures:

```go
config, err := core.LoadMultiAppConfig()
if err != nil {
    if errors.Is(err, fs.ErrNotExist) {
        fmt.Fprintf(f.IOStreams.ErrOut, "Not configured yet. Config file path: %s\n", core.GetConfigPath())
        fmt.Fprintln(f.IOStreams.ErrOut, "Run `lark-cli config init` to initialize.")
        return nil
    }
    return fmt.Errorf("failed to load config: %w", err)
}
```

## Notes

The `config == nil` and `len(config.Apps) == 0` guard conditions in both files are also redundant — `LoadMultiAppConfig` never returns a non-nil config alongside a nil error when apps are empty (it returns an error instead).

crw9527:
## Summary

- `config show` and `config remove` previously treated all `LoadMultiAppConfig` failures (including permission denied and JSON parse errors) as "not configured yet"
- `config show` returned exit code 0 on a permission/parse error, making it impossible for scripts to detect the real problem
- Users could be misled into running `config init` and overwriting a valid but temporarily unreadable config

## Changes

Use `errors.Is(err, fs.ErrNotExist)` to distinguish "file not found" from other failures:
- `fs.ErrNotExist` → show "Not configured yet" guidance (existing behavior)
- All other errors → surface the actual error message with non-zero exit

Also removes the redundant `config == nil` and `len(config.Apps) == 0` guard conditions — `LoadMultiAppConfig` never returns a non-nil config with a nil error when apps are empty.

Fixes #110

## Test plan

- [ ] `config show` with no config file → "Not configured yet" (unchanged)
- [ ] `config show` with unreadable config file (`chmod 000`) → actual permission error
- [ ] `config show` with corrupt JSON config → actual parse error
- [ ] `config remove` with no config file → "not configured yet" validation error (unchanged)
- [ ] `config remove` with unreadable config file → actual permission error

### DIFF
--- a/cmd/config/remove.go
+++ b/cmd/config/remove.go
@@ -4,7 +4,9 @@
 package config
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 
 	"github.com/larksuite/cli/internal/auth"
 	"github.com/larksuite/cli/internal/cmdutil"
@@ -40,8 +42,11 @@ func configRemoveRun(opts *ConfigRemoveOptions) error {
 	f := opts.Factory
 
 	config, err := core.LoadMultiAppConfig()
-	if err != nil || config == nil || len(config.Apps) == 0 {
-		return output.ErrValidation("not configured yet")
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return output.ErrValidation("not configured yet")
+		}
+		return fmt.Errorf("failed to load config: %w", err)
 	}
 
 	// Clean up keychain entries for all apps
--- a/cmd/config/show.go
+++ b/cmd/config/show.go
@@ -4,7 +4,9 @@
 package config
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"strings"
 
 	"github.com/larksuite/cli/internal/cmdutil"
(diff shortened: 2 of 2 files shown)


sang-neo03:
@crw9527 Could you add a test for configRemoveRun when the config file exists but is unreadable (e.g. permission error) or has corrupt JSON? Something similar to                          
  TestConfigShowRun_NotConfiguredReturnsStructuredError — verify it returns the actual error instead of "not configured yet".

PR #113, opened 2026-03-30
[codex] Inject output helper error sink

ShaoyeLi:
## Summary
- route output helper failures through an injectable package-level error sink instead of writing directly to `os.Stderr`
- update CSV and print helpers to share the same redirected error path
- add focused regression tests covering redirected JSON, NDJSON, and CSV failure output

## Root cause
Output helpers wrote marshal/flush failures directly to `os.Stderr`, which bypassed injected IO streams used by tests, embedding, and agent-side output capture.

## Testing
- `gofmt -w internal/output/print.go internal/output/csv.go internal/output/print_test.go`
- `go test ./internal/output`

### DIFF
--- a/internal/output/csv.go
+++ b/internal/output/csv.go
@@ -7,7 +7,6 @@ import (
 	"encoding/csv"
 	"fmt"
 	"io"
-	"os"
 )
 
 // FormatAsCSV formats data as CSV (with header) and writes it to w.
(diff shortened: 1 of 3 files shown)


liuxinyanglxy:
We already have unified error handling, so I don’t recommend implementing this separately

PR #115, opened 2026-03-30
fix(config): save empty config before clearing keychain in config remove

### LINKED ISSUE #114: bug: config remove clears keychain before saving config, leaving inconsistent state on failure
## Summary

`lark-cli config remove` clears keychain entries **before** saving the empty config file. If the config save fails (disk full, permission error, filesystem issue), the keychain is already wiped but `config.json` still contains the old app data — leaving the CLI in an unrecoverable state.

## Location

`cmd/config/remove.go:48-59`

```go
// Step 1: clean keychain (irreversible)
for _, app := range config.Apps {
    core.RemoveSecretStore(app.AppSecret, f.Keychain)
    for _, user := range app.Users {
        auth.RemoveStoredToken(app.AppId, user.UserOpenId)
    }
}

// Step 2: save empty config (can fail)
empty := &core.MultiAppConfig{Apps: []core.AppConfig{}}
if err := core.SaveMultiAppConfig(empty); err != nil {
    return output.Errorf(output.ExitInternal, "internal", "failed to save config: %v", err)
}
```

## Impact

When step 2 fails:
- `config.json` still references the old app ID and users
- The keychain secrets and tokens are already gone
- On the next run, `LoadMultiAppConfig` succeeds but `ResolveSecretInput` fails to find the secret in keychain — the CLI is stuck in a broken state the user cannot recover from without manually editing `config.json`

## Suggested Fix

Swap the order: **save the empty config first, then clear the keychain**.

```go
// Step 1: save empty config first
empty := &core.MultiAppConfig{Apps: []core.AppConfig{}}
if err := core.SaveMultiAppConfig(empty); err != nil {
    return output.Errorf(output.ExitInternal, "internal", "failed to save config: %v", err)
}

// Step 2: clean keychain (best-effort, orphaned entries are harmless)
for _, app := range config.Apps {
    core.RemoveSecretStore(app.AppSecret, f.Keychain)
    for _, user := range app.Users {
        auth.RemoveStoredToken(app.AppId, user.UserOpenId)
    }
}
```

If step 1 fails → keychain is intact, user can retry.  
If step 2 fails → config is already empty, orphaned keychain entries are harmless (never loaded).

crw9527:
## Summary

Fixes the operation order in `config remove` to prevent an unrecoverable state when the config save fails.

## Problem

Previously the order was:
1. Clear keychain entries ← irreversible
2. Save empty config ← if this fails, keychain is already wiped

If step 2 failed (disk full, permission error, etc.), `config.json` still contained the old app data but the keychain secrets were gone. On the next run, `LoadMultiAppConfig` would succeed but `ResolveSecretInput` would fail to find the secret — leaving the user stuck with no way to recover without manually editing `config.json`.

## Fix

Swap the order:
1. Save empty config first — if this fails, keychain is intact and the user can retry
2. Clear keychain entries — best-effort; orphaned entries are harmless since `config.json` no longer references them

## Changes

`cmd/config/remove.go` — swap the two blocks, update comments

Fixes #114

## Test plan

- [ ] `config remove` with a valid config → works as before, prints "Configuration removed"
- [ ] `config remove` when config save fails → keychain entries remain intact
- [ ] `config remove` when keychain clear fails → config is already empty, no broken state

### DIFF
--- a/cmd/config/remove.go
+++ b/cmd/config/remove.go
@@ -44,18 +44,18 @@ func configRemoveRun(opts *ConfigRemoveOptions) error {
 		return output.ErrValidation("not configured yet")
 	}
 
-	// Clean up keychain entries for all apps
+	// Save empty config first — if this fails, keychain is still intact and the user can retry
+	empty := &core.MultiAppConfig{Apps: []core.AppConfig{}}
+	if err := core.SaveMultiAppConfig(empty); err != nil {
+		return output.Errorf(output.ExitInternal, "internal", "failed to save config: %v", err)
+	}
+
+	// Clean up keychain entries — best-effort; orphaned entries are harmless
 	for _, app := range config.Apps {
 		core.RemoveSecretStore(app.AppSecret, f.Keychain)
 		for _, user := range app.Users {
 			auth.RemoveStoredToken(app.AppId, user.UserOpenId)
 		}
-	}
-
-	// Save empty config
-	empty := &core.MultiAppConfig{Apps: []core.AppConfig{}}
-	if err := core.SaveMultiAppConfig(empty); err != nil {
-		return output.Errorf(output.ExitInternal, "internal", "failed to save config: %v", err)
 	}
 	output.PrintSuccess(f.IOStreams.ErrOut, "Configuration removed")
 	userCount := 0


liuxinyanglxy:
This is a real bug, but it was already fixed on main by #291 (commit `adef52ad`) — `config remove` now writes the empty config first and only clears the keychain/token afterward. Thanks for the PR; the approach was exactly right！

PR #125, opened 2026-03-31
fix: harden auth check and config removal

### LINKED ISSUE #114: bug: config remove clears keychain before saving config, leaving inconsistent state on failure

### LINKED ISSUE #116: bug: auth check --scope "" bypasses required flag validation and returns ok: true
## Summary

`lark-cli auth check --scope ""` bypasses `MarkFlagRequired` (which only prevents omitting the flag entirely) and returns `{"ok": true}` with exit code 0, as if all scopes were granted. This can silently mislead scripts into thinking a scope check passed when nothing was actually checked.

## Location

`cmd/auth/check.go:47-51`

```go
required := strings.Fields(opts.Scope)
if len(required) == 0 {
    output.PrintJson(f.IOStreams.Out, map[string]interface{}{"ok": true, "granted": []string{}, "missing": []string{}})
    return nil  // exit 0 — misleading
}
```

## Reproduction

```bash
lark-cli auth check --scope ""
# Output: {"ok":true,"granted":[],"missing":[]}
# Exit code: 0
```

## Impact

Scripts that build the scope string dynamically can silently pass an empty value:

```bash
REQUIRED_SCOPE=""
lark-cli auth check --scope "$REQUIRED_SCOPE"
# Exits 0, script proceeds thinking scopes are fine
```

## Suggested Fix

Return a validation error when the resolved scope list is empty:

```go
required := strings.Fields(opts.Scope)
if len(required) == 0 {
    return output.ErrValidation("--scope cannot be empty")
}
```

zfaustk:
## Summary
- reject empty resolved scope lists in `auth check`
- save the empty config before clearing secrets/tokens in `config remove`
- add focused regression tests for both failure-prone paths

Fixes #116
Fixes #114

## Why
`auth check --scope ""` currently returns `{"ok":true}` even though nothing was actually checked. That can silently greenlight automation when a dynamically built scope string collapses to empty.

`config remove` currently clears keychain/token state before persisting the empty config. If the config write fails, the CLI is left in an inconsistent state: `config.json` still points at the old app/user records, but the secret/token material is already gone.

## Testing
- added `TestAuthCheckRun_EmptyScopeReturnsValidationError`
- added `TestConfigRemoveRun_SaveFailureDoesNotClearSecretsOrTokens`
- added `TestConfigRemoveRun_ClearsSecretsAndTokensAfterSuccessfulSave`
- attempted `CGO_ENABLED=0 /ROOM/.toolchains/go1.24.13/bin/go test ./cmd/auth ./cmd/config` in this container, but the local Go build/test process did not complete within the timeout window, so I am not claiming a full local `go test` pass from this environment

## Summary by CodeRabbit

* **Bug Fixes**
  * Auth check command now properly validates and rejects empty scopes.
  * Config removal command now displays warnings when token cleanup fails instead of silently ignoring errors.
  * Improved input validation for config removal operations.

* **Tests**
  * Added test coverage for empty scope validation.
  * Added test coverage for config removal functionality.


HanShaoshuai-k (changes requested):
### Suggestions

**1. Package-level mutable vars don't follow project conventions** (`remove.go:16-21`) — Blocking

The project consistently uses `Factory` / `TestFactory` for dependency injection — HTTP via `httpmock.Registry`, keychain via `noopKeychain`, IO via buffer replacement. No other test uses package-level var swapping.

Suggest moving these deps into `ConfigRemoveOptions` or `Factory` to stay consistent. The current approach also risks data races if `t.Parallel()` is ever added.

**2. Token cleanup errors silently discarded** (`remove.go:56`) — Medium

```go
_ = removeStoredToken(app.AppId, user.UserOpenId)
```

If save succeeds but token cleanup fails, orphaned keychain entries remain with no CLI path to clean them. Suggest logging a warning on failure.

**3. CLA not signed** — Blocking

Please sign the [Contributor License Agreement](https://cla-assistant.io/larksuite/cli?pullRequest=125) before we can merge.

zfaustk:
Resolved the merge conflict and refreshed the branch on the latest head (`7323a70`).

This keeps the scoped follow-up only: empty/whitespace `--scope` validation, injected callbacks for `config remove`, token cleanup warnings, and focused tests. Please re-review when you have a chance.

zfaustk:
Addressed the remaining  review concern in the latest two commits.\n\n- removed the raw  value from the token-cleanup warning\n- kept the warning actionable by retaining the app id and underlying error\n- updated  so the warning assertion no longer expects or allows the user identifier to appear\n\nAt this point the code-side cleanup for this review thread should be complete; the remaining blocker should be maintainer re-review and workflow approval.\n

zfaustk:
Follow-up clarification: the latest commits update cmd/config/remove.go and cmd/config/remove_test.go.

The warning now avoids printing the raw user identifier, and the focused test now checks that the identifier is not leaked.

sang-neo03:
@zfaustk 
Blocking (must fix before merge)

  1. `cmd/config/remove.go` was effectively moved out of `cmd/config/` and now appears as a root-level file named `undefined` (`undefined:17-89`).
     This breaks package layout and makes `cmd/config/config.go:21` lose the in-package `NewCmdConfigRemove` definition. Please restore the file to `cmd/config/remove.go`.

  2. `remove_test.go` is added at repo root with `package config` (`remove_test.go:4`), but the root package is `main` (`main.go:5`).
     This creates a package conflict at the repository root. Please move it to `cmd/config/remove_test.go`.

  Major

  3. Missing regression coverage for the `SaveConfig` failure path (`remove_test.go:16-124`).
     The core invariant of #114 is that if `SaveConfig` fails, `RemoveSecret` and `RemoveStoredToken` must not run. The current tests only cover the success path and warning path.

  4. `LoadConfig` errors are swallowed in `configRemoveRun` (`undefined:59-61`).
     `permission denied`, invalid JSON, and “not configured yet” all collapse into the same validation error. These should be handled separately so the CLI returns actionable diagnostics.

  5. The final summary is inaccurate when token cleanup partially fails (`undefined:75-87`).
     The command can emit warnings for failed removals, but still prints `Cleared tokens for N users`. Please track failures and reflect partial success in the summary.

  Suggestions

  6. Consider squashing before merge. The PR currently has 10 commits, including several fixup/conflict-resolution style commits.

  7. Consider adding a lightweight `go build` or `go test` CI gate. The current checks did not catch a compile/package-layout breakage.

  What looks good

  - The `auth check --scope ""` fix is correct and minimal (`cmd/auth/check.go:47-50`)
  - Save-before-cleanup is the right direction for `config remove`
  - Dependency injection here improves testability
  - The warning message avoids leaking user identifiers

zfaustk:
Closing this as superseded. PR #291 merged the config save-before-cleanup fix, current main rejects an empty auth scope, and both linked issues are closed.

PR #183, opened 2026-04-01
docs(lark-base): add has_more guidance for record-list pagination

kongenpei:
## Summary
- add has_more response field docs to base +record-list reference
- add Chinese pagination rules: only continue when has_more=true and user query needs more data
- update skill-level rule to avoid unnecessary full-data preloading

## Testing
- docs only (no runtime changes)

## Documentation
- Updated record list pagination guidelines with conditional paging rules
- Clarified pagination logic: only continue to subsequent pages when more data exists and the user explicitly requests it
- Enhanced documentation to guide users to avoid unnecessary fetching when only a subset of records is needed


zgz2048 (approved):
LGTM

kongenpei:
close https://github.com/larksuite/cli/issues/319

PR #216, opened 2026-04-02
docs(lark-doc): clarify when escape is needed to prevent agent over-escaping

### LINKED ISSUE #155: docs(lark-doc): escape rules lack trigger context, causing agents to over-escape
## Problem

The Lark-flavored Markdown documentation in `skills/lark-doc/references/lark-doc-create.md` lists which characters need escaping, but never explains **when** escaping is needed:

- Line 123 (通用规则): `需要显示特殊字符时使用反斜杠转义：* ~ ` $ [ ] < > { } | ^`
- Line 645 (最佳实践): `转义字符：特殊字符用 \ 转义：\* \~ \``

The formatting syntax (`**bold**`, `*italic*`, `~~strikethrough~~`, etc.) is documented separately in the 富文本格式 section (line 216), but there is no connection between the two — the doc never states that escaping is only needed when a character would be interpreted as formatting syntax.

## Impact

AI Agents interpret the escape rule as "always escape these characters", leading to over-escaping:

- `final_trajectory` → `final\_trajectory`
- `version~1.0` → `version\~1.0`
- `5 * 3` → `5 \* 3`

The Lark Markdown parser does not consume the extra `\`, so it renders literally as `\_`, `\~`, `\*` in the document.

## Suggestion

Tie the escape rules to the formatting syntax, so agents understand escaping is only needed when a character would trigger formatting (e.g., `*text*` triggers italic, but a lone `*` in `5 * 3` does not).

meijing0114:
## Problem

Fixes #155

The escape rules in `lark-doc-create.md` list which characters can be escaped but never explain **when** escaping is actually needed. AI agents interpret this as "always escape these characters", producing garbled output:

- `final_trajectory` → `final\_trajectory`
- `version~1.0` → `version\~1.0`
- `5 * 3` → `5 \* 3`

The Lark Markdown parser doesn't consume the extra `\`, so it renders literally as `\_`, `\~`, `\*`.

## Fix

Two changes in `skills/lark-doc/references/lark-doc-create.md`:

1. **通用规则 (line 122)**: Reworded to clarify escaping is only needed when a character would trigger formatting syntax. Added concrete examples of when to escape vs not.

2. **最佳实践 (line 647)**: Synced the same clarification — escape only when formatting would be triggered.

## Examples added

| Context | Escape needed? | Why |
|---------|---------------|-----|
| `\*text\*` | ✅ Yes | `*...*` triggers italic |
| `5 * 3` | ❌ No | Lone `*` doesn't trigger formatting |
| `final_trajectory` | ❌ No | `_` inside a word doesn't trigger italic |
| `version~1.0` | ❌ No | Lone `~` doesn't trigger strikethrough |

## Summary by CodeRabbit

* **Documentation**
  * Updated Lark-flavored Markdown escaping guidance: changed from universal escaping to conditional escaping based on whether a character triggers formatting.
  * Added explicit examples clarifying when asterisks, tildes, underscores, and other characters need escaping versus when they can remain unescaped.
  * Aligned the best-practices section with the new conditional-escaping rules.


fangshuyu-768:
Thanks a lot for the contribution and for helping improve the documentation here.

This issue has already been addressed and optimized in a later update on our side, so this PR is now largely covered by the current main branch.

We’re also planning to provide a more complete doc update and editing guideline/protocol in follow-up documentation improvements, so please feel free to stay tuned.

Really appreciate the effort and thoughtful suggestion here. Thanks again!

PR #232, opened 2026-04-02
feat(auth): support multiple users by upserting on login

mio4kon:
#204 
- Replace overwrite logic with upsertUser helper: existing users with the same UserOpenId are updated in place, others are preserved.
- Add resolveActiveUser in config: respects LARKSUITE_CLI_USER_OPEN_ID env var to select the active user; falls back to first user if unset.

## Changes

- Change 1
- Change 2

## Test Plan

- [x] Unit tests pass
- [x] Manual local verification confirms the `lark xxx` command works as expected

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Manage multiple authenticated user accounts with persistent credentials.
  * Environment variable to select the active user account.

* **Changes**
  * Login now preserves existing sessions when adding or updating an account instead of replacing them.
  * Logout removes only the active account’s credentials rather than clearing all stored sessions.
  * Active-user resolution now governs authentication checks and behavior.

### DIFF
--- a/cmd/auth/login.go
+++ b/cmd/auth/login.go
@@ -303,16 +303,11 @@ func authLoginRun(opts *LoginOptions) error {
 		return output.Errorf(output.ExitInternal, "internal", "failed to save token: %v", err)
 	}
 
-	// Step 8: Update config — overwrite Users to single user, clean old tokens
+	// Step 8: Update config — upsert user, keep existing users
 	multi, _ := core.LoadMultiAppConfig()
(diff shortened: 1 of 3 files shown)


liuxinyanglxy:
Thanks for the contribution. We agree that the multi-user authorization need described in #204 is valuable for enterprise multi-user agents. This capability first needs to be aligned with the current profile and credential-isolation model, so please hold off on further changes for now; the current PR is not ready to merge.

Issue #259, opened 2026-04-03
Feature: docs +fetch should support --resolve-media to auto-download images/files


liuhedev:
## Problem

When using `docs +fetch` to export a Lark document as Markdown, images and files are exported as placeholder tags:

```html
<image token="DKblbz86oojD..." width="1833" height="2491"/>
<file token="WIKKbg01lonJ..." name="skills.zip"/>
```

These placeholders cannot be rendered in any Markdown viewer. Users must manually call `docs +media-download` for each token to get the actual files.

**Critical issue**: When fetching documents from shared Wiki spaces, `docs +fetch` has read permission at fetch time, but `docs +media-download` called later often returns **HTTP 403 Forbidden** — the media download API requires separate permission that may not be available after the initial fetch.

## Proposed Solution

Add a `--resolve-media` flag to `docs +fetch` that:

1. During the fetch, identifies all `<image>`, `<file>`, and `<whiteboard>` placeholder tags
2. Immediately downloads each media asset using the same auth context (while permission is still valid)
3. Saves assets to a local directory (e.g., `assets/<doc-name>/`)
4. Replaces placeholder tags with standard Markdown references (`![](assets/...)`, `[file](assets/...)`)

```bash
# Current behavior (placeholders only)
lark-cli docs +fetch --doc /wiki/TOKEN

# Proposed behavior (with media resolution)
lark-cli docs +fetch --doc /wiki/TOKEN --resolve-media
lark-cli docs +fetch --doc /wiki/TOKEN --resolve-media --media-dir ./images
```

## Workaround

I wrote a Python script that does this post-hoc, but it fails with 403 on shared documents because the media download permission window has closed: https://gist.github.com/liuhedev

## Environment

- lark-cli version: latest
- OS: macOS arm64
- Auth: user identity (feishu)

ahighmoon:
Also hitting this. Reproduction details:

**Environment**: lark-cli 1.0.24, macOS arm64, auth as user (feishu)

**Steps**:
1. `lark-cli docs +fetch --api-version v2 --doc "<docx_url>" --doc-format markdown --as user`
2. Markdown output contains `![](https://feishu.cn/file/P2sLb5Kjgoia3lxPcuauuuShsgb)` — these URLs are web preview links, not direct downloads. Attempting `curl -L` follows a 301 redirect to `www.feishu.cn/file/...` and returns 404.
[...]

PR #260, opened 2026-04-03
feat(docs): normalize <lark-table> to GFM pipe tables in docs +fetch

liuhedev:
## Summary

`docs +fetch` outputs markdown containing custom `<lark-table>` tags that are not renderable in any standard markdown viewer (Obsidian, GitHub, VS Code, etc.). This PR adds automatic conversion to GFM pipe tables.

## Changes

- **`shortcuts/doc/lark_table.go`**: New `NormalizeLarkTables()` function that converts `<lark-table>/<lark-tr>/<lark-td>` blocks to standard GFM pipe tables
- **`shortcuts/doc/docs_fetch.go`**: Calls `NormalizeLarkTables()` on the markdown output during pretty-print
- **`shortcuts/doc/lark_table_test.go`**: Unit tests covering basic tables, no-header mode, code-block fallback, and pipe escaping

## Behavior

| Before | After |
|--------|-------|
| `<lark-table><lark-tr><lark-td>Name</lark-td>...` | `\| Name \| Age \|` |

### Edge cases handled
- **`header-row="false"`**: Skips separator row
- **Fenced code blocks in cells**: Falls back to original `<lark-table>` (GFM pipe tables cannot contain multi-line code)
- **Pipe characters**: Escaped as `\|`
- **Nested `<grid>/<column>/<text>` tags**: Flattened and stripped

## Motivation

When using `docs +fetch` to export Lark documents for local use (e.g., in Obsidian vaults), tables are the most common broken element. This change makes the output immediately usable without post-processing.

Related issue: #259

## Summary by CodeRabbit

* **New Features**
  * Documentation now converts custom Lark-table markup into GitHub-flavored Markdown tables with proper headers (optional), column alignment, cell normalization (nested content flattened, <br/> normalized, pipe characters escaped), and preserves original markup if fenced code blocks are present.

* **Tests**
  * Added comprehensive unit tests covering header handling, code-block preservation, special-character escaping, uneven columns, and nested-content flattening.

### DIFF
--- a/shortcuts/doc/docs_fetch.go
+++ b/shortcuts/doc/docs_fetch.go
@@ -66,6 +66,7 @@ var DocsFetch = common.Shortcut{
 				fmt.Fprintf(w, "# %s\n\n", title)
 			}
 			if md, ok := result["markdown"].(string); ok {
+				md = NormalizeLarkTables(md)
 				fmt.Fprintln(w, md)
 			}
 			if hasMore, ok := result["has_more"].(bool); ok && hasMore {
--- a/shortcuts/doc/lark_table.go
+++ b/shortcuts/doc/lark_table.go
@@ -4,24 +4,26 @@
 package doc
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 )
 
 var (
-	larkTableRE = regexp.MustCompile(`(?is)<lark-table\b[^>]*>(.*?)</lark-table>`)
-	larkTrRE    = regexp.MustCompile(`(?is)<lark-tr\b[^>]*>(.*?)</lark-tr>`)
-	larkTdRE    = regexp.MustCompile(`(?is)<lark-td\b[^>]*>(.*?)</lark-td>`)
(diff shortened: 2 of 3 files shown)


fangshuyu-768:
Thanks a lot for the contribution — this is a real usability issue, and we appreciate you taking the time to work on it.

We’re already working on a new version of the docs read/write layer that is intended to address this problem more systematically. In the new design, we provide two protocols: Markdown and XML. The Markdown path is intended to
correspond to standard Markdown for better interoperability, while the XML path is used for richer Lark-specific structures that cannot be expressed cleanly in Markdown.

So for cases like table output, the direction is not to keep exposing private Lark-specific markup in the Markdown protocol, but to make the Markdown output itself more standard and portable. This work is already in progress, and we
expect to provide it soon.

Since we’re already addressing this in the upcoming redesign, I’m going to close this PR for now rather than keep it open indefinitely.

Thanks again for the contribution. Please keep following the project, and we’d be very happy to continue hearing your suggestions and feedback.

PR #275, opened 2026-04-06
fix(registry): don't fallback embeddedMetaJSON to default empty JSON

yhp20170308-dotcom:
## Summary

- `loader_embedded.go` was setting `embeddedMetaJSON` to the default empty JSON (`{"version":"0.0.0","services":[]}`) when `meta_data.json` is not compiled in
- This caused `hasEmbeddedData()` to return `true` (34 bytes > 0), even though there were no real services
- As a result, two tests failed instead of being correctly skipped:
  - `TestColdStart_UsesEmbedded`: failed with "expected embedded projects, got none"
  - `TestCacheHit_WithinTTL`: failed with "expected calendar from embedded data"

## Fix

Remove the `else` fallback so `embeddedMetaJSON` remains `nil` when no real `meta_data.json` is present. `hasEmbeddedData()` then correctly returns `false` and dependent tests skip as intended.

## Test plan

- [x] `go build ./...` — passes
- [x] `go vet ./...` — no warnings
- [x] `TestColdStart_UsesEmbedded` — now correctly SKIP (was FAIL)
- [x] `TestCacheHit_WithinTTL` — now PASS (was FAIL)
- [x] All other tests — still pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing embedded metadata. The system now properly detects when metadata is unavailable instead of applying default fallback values, enabling more accurate downstream behavior.

### DIFF
--- a/internal/registry/loader_embedded.go
+++ b/internal/registry/loader_embedded.go
@@ -14,7 +14,8 @@ var embeddedMetaDataDefaultJSON []byte
 func init() {
 	if data, err := metaFS.ReadFile("meta_data.json"); err == nil && len(data) > 0 {
 		embeddedMetaJSON = data
-	} else {
-		embeddedMetaJSON = embeddedMetaDataDefaultJSON
 	}
+	// When meta_data.json is not compiled in, embeddedMetaJSON remains nil,
+	// so hasEmbeddedData() correctly returns false and tests that depend on
+	// real embedded services will be skipped rather than failing.
 }


sang-neo03:
@yhp20170308-dotcom 
Thanks for the fix — the core change (removing the fallback) is correct and the reasoning is sound.                                                                                 
                                                                                                                                                                                      
  A few things to address before we can merge:                                                                                                                                        
                                                                                                                                                                                      
  1. Clean up dead code: After removing the else branch, embeddedMetaDataDefaultJSON (lines 11-12) and meta_data_default.json are no longer referenced anywhere. Please remove both   
  the //go:embed directive + variable declaration, and delete the meta_data_default.json file. Keeping them around adds unnecessary binary bloat and confusion for future readers.    
  2. Sign the CLA: The CLAassistant check is failing — please complete the CLA signing so the status turns green.                                                                     
  3. Fix CI: The docstring coverage check is currently at 0% (threshold: 80%). Please address the failing check to get CI green.

PR #276, opened 2026-04-06
docs(im): Document critical Feishu Markdown silent drop limitations

Scofy0123:
This PR adds critical documentation to the `im +messages-send` shortcut regarding Feishu's Markdown parser limitations.

Specifically, it warns users about two silent message drop conditions:
1. Using non-HTTP(S) schemas in Markdown links (e.g. `[query](from:user)`) triggering client-side dropping.
2. Nesting links inside bold tags (`**[title](url)**`), which strips the link completely in the native Feishu Post renderer.

These additions will prevent other agents and CLI users from generating malformed markdown that quietly breaks rendering without raising an API error.

## Summary by CodeRabbit

* **Documentation**
  * Updated link formatting guidance for Lark IM messages with new parsing restrictions and caveats.
  * Documented that bold-wrapped links may be silently dropped and provided alternative formatting patterns.
  * Clarified supported URL schemes in Markdown links.

### DIFF
--- a/skills/lark-im/references/lark-im-messages-send.md
+++ b/skills/lark-im/references/lark-im-messages-send.md
@@ -57,6 +57,9 @@ This means `--markdown` is convenient, but it is not a full-fidelity Markdown tr
 - It does **not** promise full CommonMark / GitHub Flavored Markdown support.
 - It always becomes a `post` payload with a single `zh_cn` locale.
 - It does **not** let you set a `post` title. If you need a title, use `--msg-type post --content ...`.
+- **Link parsing restrictions (Critical for silent drops):**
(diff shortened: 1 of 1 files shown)


YangJunzhou-01:
Thanks for the contribution. I checked the bold-nested-link case against a raw `im/v1/messages` `post` payload:
```md
**[Title](https://example.com/pr276-exact-bold)**
```
The rendered post content keeps the bold text but does not preserve the anchor, so documenting this limitation is useful for CLI users and AI agents that generate Markdown.
﻿
One wording request before merge:
﻿
Please keep the “whole paragraph may be dropped” wording scoped to the non-HTTP(S) scheme case, if that is the behavior you observed there.
For bold nested links, please describe the behavior more precisely as “the link/anchor is stripped while the text remains.” For example: avoid `**[Title](https://...)**`; use `**Item.** [Title](https://...) `instead.
After that wording adjustment, this looks good to merge.

PR #305, opened 2026-04-07
fix(base): reject wrapped JSON payloads in +record-upsert

### LINKED ISSUE #272: Bug Report: lark-cli base record-upsert POST 请求全部失败
没有 GitHub 登录权限，无法直接提交 issue。我把详细的 bug 报告整理好，你来提交：

───

Bug Report: lark-cli base record-upsert POST 请求全部失败

仓库: https://github.com/larksuite/cli

版本: v1.0.0

问题描述:
所有通过 lark-cli 的 base +record-upsert 和 api POST 命令向 Base（多维表格）写入记录的操作均失败，返回 800010701 Invalid input 错误。

复现步骤:

# 1. 认证
lark-cli auth login  # 已完成，token 有效

# 2. 列出表格（成功）
lark-cli base +table-list --as user --base-token <base_token>  
# 返回: {"ok": true, ...}

# 3. 列出字段（成功）
lark-cli base +field-list --as user --base-token <base_token> --table-id <table_id>
# 返回: {"ok": true, "data": {"items": [...]}}

# 4. 查询记录（成功）
lark-cli base +record-list --as user --base-token <base_token> --table-id <table_id>
# 返回: {"ok": true, "data": {"data": [], "record_id_list": []}}

# 5. 创建记录（失败）
lark-cli base +record-upsert --as user --base-token <base_token> --table-id <table_id> --json '{"fields":{"文本":"test"}}'
# 返回: {"ok": false, "error": {"code": 800010701, "message": "API call failed: [800010701] Invalid input."}}

# 6. API POST（同样失败）
lark-cli api POST /open-apis/base/v3/bases/<token>/tables/<table_id>/records --as user --data '{"fields":{"文本":"test"}}'
# Exit code: 1, 无输出

已尝试的排查:

• ✅ token 有效（GET 请求正常）
• ✅ 权限 scope 完整（base:record:create 等已授权）
• ✅ 尝试了多个表格（新建测试表同样失败）
• ✅ 尝试了不同字段格式（字段名、字段ID、中文、英文）
• ✅ 尝试了不同 JSON 格式
• ✅ dry-run 模式显示的请求格式正确，但实际执行失败
• ✅ PTY 模式、background、sudo 等全部失败

预期行为:
能够向 Base 表格创建新记录。

实际行为:
所有 POST 写入操作静默失败，返回 Invalid input 错误。

环境:

• macOS
• lark-cli v1.0.0
• Node.js / npm

建议: 检查 record-upsert 和 api POST 命令的请求体序列化逻辑，或在文档中说明是否有特殊的写入前置条件。

───

你去这里提交：https://github.com/larksuite/cli/issues/new

选 Bug Report 模板，把上面的内容贴进去。

zfaustk:
## Summary
Fixes #272.

The current `base +record-upsert --json` path accepts a common wrong payload shape, `{"fields": {...}}`, and only fails later in the Base API layer. This change surfaces the contract earlier in the CLI by requiring a direct record object.

## Changes
- reject a top-level `fields` wrapper in `validateRecordJSON()` and show a concrete example of the expected shape
- update `base_shortcuts_test.go` to accept direct record JSON and cover the wrapped-payload validation error
- update `base_execute_test.go` to use the direct record shape for create/update and add a regression case for the wrapped payload

## Validation
- local focused diff reviewed against `/tmp/larkcli-triage-VvPLOV/repo`
- `git diff --check`
- Go tests not run in this environment because `go` is not installed in the container

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation for record upsert operations to reject improperly formatted JSON payloads and provide clear error messages guiding users toward the correct direct record object format.

* **Breaking Changes**
  * Record operations now strictly require direct record object format; the previously supported wrapped format is no longer accepted.

### DIFF
--- a/shortcuts/base/base_execute_test.go
+++ b/shortcuts/base/base_execute_test.go
@@ -222,7 +222,7 @@ func TestBaseRecordExecuteUpsertUpdate(t *testing.T) {
 			"data": map[string]interface{}{"record_id": "rec_x", "fields": map[string]interface{}{"Name": "Alice"}},
 		},
 	})
-	if err := runShortcut(t, BaseRecordUpsert, []string{"+record-upsert", "--base-token", "app_x", "--table-id", "tbl_x", "--record-id", "rec_x", "--json", `{"fields":{"Name":"Alice"}}`}, factory, stdout); err != nil {
+	if err := runShortcut(t, BaseRecordUpsert, []string{"+record-upsert", "--base-token", "app_x", "--table-id", "tbl_x", "--record-id", "rec_x", "--json", `{"Name":"Alice"}`}, factory, stdout); err != nil {
 		t.Fatalf("err=%v", err)
 	}
 	if got := stdout.String(); !strings.Contains(got, `"updated": true`) || !strings.Contains(got, `"rec_x"`) {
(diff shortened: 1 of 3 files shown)


zgz2048:
Thanks for the PR and for taking the time to improve the `+record-upsert` experience.

I'm going to close this version for two reasons.

First, I don't think the fix direction is the right one. The payload contract for `+record-upsert` is already clearly defined in the existing skill/docs: `--json` should be a direct record object, not a wrapped `{"fields": ...}` shape. Given that, the CLI should not add a hard-coded heuristic just for the literal `fields` key. This makes the command behavior more implicit, and it can also reject legitimate payloads where a real field is actually named `fields`.

Second, this PR also introduces unrelated mojibake/garbled text changes in tests (`附件 -> éä»¶`, etc.), which makes the patch noisy and not cleanly scoped to the actual issue.

If we want to improve this area, I'd prefer a cleaner approach: keep the command aligned with the documented contract, avoid special-casing one mistaken shape, and make sure the patch stays narrowly scoped with no unrelated text corruption.

PR #307, opened 2026-04-07
fix(cli): make completion command discoverable

### LINKED ISSUE #264: `lark-cli completion` 子命令被隐藏
试了一下这个子命令是可以直接使用的，能否开启或者在文档里面说明呢？

https://github.com/larksuite/cli/blob/0c77c95a11c9ead1f82b4c268815164104c1249c/cmd/completion/completion.go#L19

zfaustk:
## Summary
Fixes #264.

`lark-cli completion` already works, but it is hidden from normal command discovery and is not documented in either README. This change makes the command visible and adds short shell completion setup examples in both English and Chinese docs.

## Changes
- remove `Hidden: true` from `cmd/completion/completion.go`
- add `cmd/completion/completion_test.go` to verify the command stays visible and auth-free
- document shell completion usage in `README.md`
- document shell completion usage in `README.zh.md`

## Why this change
- issue `#264` reports that the command is usable today but hidden from users
- `CHANGELOG.md` already describes completion support as a shipped feature
- keeping the command hidden makes `--help` discovery weaker than the actual product surface

## Validation
- `git diff --check`
- focused source review of `cmd/completion/completion.go`, `cmd/root.go`, `README.md`, and `README.zh.md`
- Go tests not run in this environment because `go` is not installed in the container

## Summary by CodeRabbit

* **New Features**
  * Shell completion command is now visible and directly accessible in the CLI help output

* **Documentation**
  * Added comprehensive setup guides for shell completion with platform-specific installation instructions for Bash, Zsh, and Fish shells
  * Clarified that completion functionality operates independently without requiring user authentication

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,19 @@ lark-cli auth login --recommend
 lark-cli calendar +agenda
 ```
 
+#### Optional: Shell Completion
+
+`lark-cli completion` is available as a regular top-level command and does not require login. Generate a completion script for the shell you use most:
+
+```bash
+# Bash
+lark-cli completion bash > ~/.local/share/bash-completion/completions/lark-cli
+
+# Zsh
+mkdir -p ~/.zfunc
+lark-cli completion zsh > ~/.zfunc/_lark-cli
+```
+
 ## Quick Start (AI Agent)
 
 > The following steps are for AI Agents. Some steps require the user to complete actions in a browser.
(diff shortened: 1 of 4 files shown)


liuxinyanglxy:
Thanks for finding this, and for the PR!

The completion subcommand is intentionally hidden for now. lark-cli is primarily designed as a tool for AI agents rather than for direct human use in the terminal, so we haven't prioritized shell completion support — and while that's the case we'd prefer to keep it off the default help surface.

For that reason we're going to hold off on merging this one. That said, the command itself does work exactly as you discovered — feel free to keep using it locally if it helps your workflow. We'll keep the request tracked and may revisit making it discoverable in the future.

Really appreciate the contribution either way!

PR #312, opened 2026-04-08
docs(lark-doc): clarify when markdown escaping is needed

### LINKED ISSUE #155: docs(lark-doc): escape rules lack trigger context, causing agents to over-escape

OwenYWT:
## Summary
Clarify that Lark Markdown escaping is only needed when characters would otherwise trigger Markdown / rich-text formatting.

## Changes
- explain in the general rules section that isolated special characters in plain text should usually be left unescaped
- add concrete examples showing when escaping is unnecessary (`5 * 3`, `version~1.0`, `final_trajectory`)
- clarify in best practices that escaping is only needed when the intent is to render formatting markers literally

## Related Issues
- closes #155

## Summary by CodeRabbit

* **Documentation**
  * Clarified Lark-flavored Markdown escaping: only require backslash escaping when characters would trigger formatting, reducing unnecessary escapes in ordinary text (e.g., 5 * 3, version~1.0, final_trajectory remain unescaped). Added examples showing when to escape formatting-triggering patterns (e.g., *斜体*, **粗体**, ~~删除线~~) and updated best-practice guidance accordingly.


fangshuyu-768 (approved):

fangshuyu-768:
Thanks for the update and the careful clarification here. This makes the escaping rules much easier to understand, especially the distinction between characters that are merely present and characters that actually trigger formatting. Also appreciate the follow-up fix for the backtick example.

PR #392, opened 2026-04-10
feat(wiki): add +space-list, +node-list, +node-copy shortcuts

herbertliu:
## Summary

- **`+space-list`**: List all wiki spaces accessible to the caller (auto-pagination)
- **`+node-list`**: List nodes under a space or parent node with auto-pagination; supports `--space-id` + optional `--parent-node-token`
- **`+node-copy`**: Copy a wiki node (and its subtree) to a target space or parent node; requires `--target-space-id` or `--target-parent-node-token`

Also includes:
- Reference docs for each shortcut under `skills/lark-wiki/references/`
- Updated `skills/lark-wiki/SKILL.md` with new shortcuts table entries
- 9 unit tests covering registration, pagination, validation, and copy scenarios

## Motivation

Enables wiki content organization and cross-space migration workflows directly from lark-cli, without requiring custom scripts or manual API calls.

## Test plan

- [x] `go test ./shortcuts/wiki/...` — all tests pass
- [x] `go vet ./...` — no issues
- [x] `golangci-lint run` — 0 issues
- [x] Real-world tested against a live wiki space: `+space-list` (199 spaces), `+node-list` (184 nodes), `+node-copy` (successful copy)
- [x] Fixed gitleaks false positive — replaced example `node_token` values in docs with clearly fake `_EXAMPLE_TOKEN` placeholders

## Summary by CodeRabbit

* **New Features**
  * Added wiki commands: +space-list, +node-list, +node-copy and exposed them in shortcuts.

* **Documentation**
  * Added reference pages and skill docs for the three commands with usage, flags, examples, and responses.

* **Tests**
  * Added tests covering space/node listing (pagination and parent scoping) and node-copy validation and execution.

* **Chores**
  * Added a gitleaks target and a documentation token checker script to CI/tooling.


shortcuts/wiki/wiki_node_list.go
```diff
@@ -0,0 +1,106 @@
...
+		return common.NewDryRunAPI().
+			GET(fmt.Sprintf("/open-apis/wiki/v2/spaces/%s/nodes", spaceID)).
+			Params(params).
+			Set("space_id", spaceID)
+	},
+	Execute: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		spaceID := strings.TrimSpace(runtime.Str("space-id"))
+		if spaceID == "" {
+			return output.ErrValidation("--space-id is required")
+		}
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — removed the dead `if spaceID == ""` guard.

shortcuts/wiki/wiki_node_copy.go
```diff
@@ -0,0 +1,108 @@
...
+		return validateOptionalResourceName(targetParent, "--target-parent-node-token")
+	},
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		spaceID := strings.TrimSpace(runtime.Str("space-id"))
+		nodeToken := strings.TrimSpace(runtime.Str("node-token"))
+		return common.NewDryRunAPI().
+			POST(fmt.Sprintf("/open-apis/wiki/v2/spaces/%s/nodes/%s/copy", spaceID, nodeToken)).
+			Body(buildNodeCopyBody(runtime)).
+			Set("space_id", spaceID).
+			Set("node_token", nodeToken)
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — both `spaceID` and `nodeToken` in `DryRun` are now wrapped with `validate.EncodePathSegment`.

shortcuts/wiki/wiki_node_list.go
```diff
@@ -0,0 +1,106 @@
...
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		spaceID := strings.TrimSpace(runtime.Str("space-id"))
+		params := map[string]interface{}{"page_size": 50}
+		if pt := strings.TrimSpace(runtime.Str("parent-node-token")); pt != "" {
+			params["parent_node_token"] = pt
+		}
+		return common.NewDryRunAPI().
+			GET(fmt.Sprintf("/open-apis/wiki/v2/spaces/%s/nodes", spaceID)).
+			Params(params).
+			Set("space_id", spaceID)
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a —  now wraps  with , matching the  closure.

shortcuts/wiki/wiki_node_copy.go
```diff
@@ -0,0 +1,108 @@
...
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		targetSpaceID := strings.TrimSpace(runtime.Str("target-space-id"))
+		targetParent := strings.TrimSpace(runtime.Str("target-parent-node-token"))
+		if targetSpaceID == "" && targetParent == "" {
+			return output.ErrValidation("at least one of --target-space-id or --target-parent-node-token is required")
+		}
+		if err := validateOptionalResourceName(targetSpaceID, "--target-space-id"); err != nil {
+			return err
+		}
+		return validateOptionalResourceName(targetParent, "--target-parent-node-token")
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — `Validate` now calls `validateOptionalResourceName` on both `--space-id` and `--node-token` before checking the target flags.

scripts/check-doc-tokens.sh
```diff
@@ -0,0 +1,59 @@
...
+# Matches JSON-style: "field": "token_value" or markdown backtick spans.
+# Token prefixes used by Lark Open Platform:
+#   wikcn  doccn  docx  shtcn  bascn  fldcn  vewcn  tbln  ou_  cli_  obcn  flec
+#
+# Excluded (clearly fake):
+#   - Values ending with EXAMPLE_TOKEN  (e.g. wikcn_EXAMPLE_TOKEN)
+#   - Values that are all uppercase X   (e.g. bascnXXXXXXXX)
+#   - Values containing only X/_/<>     (e.g. <your_token>)
+REALISTIC_TOKEN_RE='"(wikcn|doccn|shtcn|bascn|fldcn|vewcn|tbln|obcn|flec)[A-Za-z0-9]{6,}"'
+PLACEHOLDER_RE='(EXAMPLE|_TOKEN|XXXX|xxxx|<|>|your_|_here)'
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — extended REALISTIC_TOKEN_RE to include `docx[a-z]`, `ou_`, `cli_` prefixes and added a backtick-quoted variant alongside the double-quoted pattern.

skills/lark-wiki/references/lark-wiki-node-copy.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+lark-cli wiki +node-copy \
+  --space-id <source_space_id> \
+  --node-token <source_node_token> \
+  --target-space-id <target_space_id> \
+  [--target-parent-node-token <token>] \
+  [--title <new_title>] \
+  [--as user|bot]
+```
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — usage line updated to `(--target-space-id <id> | --target-parent-node-token <token>)` to make the mutual-exclusion explicit.

shortcuts/wiki/wiki_node_copy.go
```diff
@@ -0,0 +1,108 @@
...
+	},
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — `Validate` now validates `--space-id` and `--node-token` with `validateOptionalResourceName` before the target presence check.

Makefile
```diff
@@ -8,7 +8,7 @@ DATE     := $(shell date +%Y-%m-%d)
 LDFLAGS  := -s -w -X $(MODULE)/internal/build.Version=$(VERSION) -X $(MODULE)/internal/build.Date=$(DATE)
 PREFIX   ?= /usr/local
 
-.PHONY: build vet test unit-test integration-test install uninstall clean fetch_meta
+.PHONY: build vet test unit-test integration-test install uninstall clean fetch_meta gitleaks
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — added `all` to `.PHONY` and an `all: test` default target.

skills/lark-wiki/references/lark-wiki-space-list.md
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "total": 1
+}
+```
+
+## Notes
+
+- Returns all spaces in a single call via auto-pagination.
+- Use `space_id` from the output as `--space-id` for `+node-list` or `+node-copy`.
+- `my_library` is the personal document library; its `space_id` is returned as a numeric ID in results.
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — updated to: "Returns all spaces via automatic pagination; the command may issue multiple API requests under the hood."

shortcuts/wiki/wiki_list_copy_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+func TestWikiNodeListPassesParentNodeToken(t *testing.T) {
+	t.Setenv("LARKSUITE_CLI_CONFIG_DIR", t.TempDir())
+
+	factory, stdout, _, reg := cmdutil.TestFactory(t, wikiTestConfig())
+
+	stub := &httpmock.Stub{
+		Method: "GET",
+		URL:    "/open-apis/wiki/v2/spaces/space_123/nodes",
+		Body: map[string]interface{}{
```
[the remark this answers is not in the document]
herbertliu: Fixed in 871627a — stub URL updated to `/open-apis/wiki/v2/spaces/space_123/nodes?page_size=50&parent_node_token=wik_parent`, so the test now fails if `--parent-node-token` is not forwarded in the outgoing request.

fangshuyu-768:
Thanks for adding the new wiki shortcuts, the coverage is moving in a good direction. I found a few issues that look worth fixing before merge though: `+node-copy` currently accepts both destination flags at the same time and will send both to the API, `+node-list` documents/supports `my_library` without resolving it the way `+node-create` already does, and the new doc-token check is not reliable yet because it both flags existing placeholder-like values and misses bare token-like examples such as `wikcnSource`. Once those are tightened up, this PR looks much safer to land.

shortcuts/wiki/wiki_node_copy.go
```diff
@@ -0,0 +1,116 @@
...
+		"At least one of --target-space-id or --target-parent-node-token must be provided.",
+		"Omit --title to keep the original node title in the copy.",
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		if err := validateOptionalResourceName(strings.TrimSpace(runtime.Str("space-id")), "--space-id"); err != nil {
+			return err
+		}
+		if err := validateOptionalResourceName(strings.TrimSpace(runtime.Str("node-token")), "--node-token"); err != nil {
+			return err
+		}
```
fangshuyu-768, line 43: `+node-copy` is described as “target space or parent node”, but the validation only checks “at least one” and `buildNodeCopyBody()` will send both fields if both flags are provided. That creates an ambiguous request shape and can easily diverge from the documented behavior. I think we should reject the case where `--target-space-id` and `--target-parent-node-token` are both set, and add a test for that explicitly.

shortcuts/wiki/wiki_node_list.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// WikiNodeList lists child nodes in a wiki space or under a parent node.
+var WikiNodeList = common.Shortcut{
+	Service:     "wiki",
+	Command:     "+node-list",
+	Description: "List wiki nodes in a space or under a parent node",
+	Risk:        "read",
+	Scopes:      []string{"wiki:node:retrieve"},
+	AuthTypes:   []string{"user", "bot"},
```
fangshuyu-768: This shortcut says `--space-id my_library` is supported, but unlike `+node-create` it never resolves `my_library` to a real `space_id`, and it also does not reject that value for bot identity. As written, we will call `/wiki/v2/spaces/my_library/nodes` directly, which looks inconsistent with the existing wiki create flow and is likely to break for at least some callers. Can we reuse the same `my_library` resolution / bot-validation logic here so the behavior matches `+node-create`?

scripts/check-doc-tokens.sh
```diff
@@ -0,0 +1,59 @@
...
+ERRORS=0
+
+# Patterns that indicate a realistic-looking Lark token value inside a string.
+# Matches JSON-style: "field": "token_value" or markdown backtick spans.
+# Token prefixes used by Lark Open Platform:
+#   wikcn  doccn  docx  shtcn  bascn  fldcn  vewcn  tbln  ou_  cli_  obcn  flec
+#
+# Excluded (clearly fake):
+#   - Values ending with EXAMPLE_TOKEN  (e.g. wikcn_EXAMPLE_TOKEN)
+#   - Values that are all uppercase X   (e.g. bascnXXXXXXXX)
```
fangshuyu-768: I think this pattern/filter is too aggressive for the current docs set. Running the script on the repo already flags existing placeholder-ish examples like `ou_manager` / `ou_director`, so the new `make gitleaks` target fails even before this PR’s docs are considered. Could we narrow the detection or broaden the placeholder allowlist first, otherwise contributors will get a broken pre-push/local check immediately?

skills/lark-wiki/references/lark-wiki-node-copy.md
```diff
@@ -0,0 +1,66 @@
...
+}
+```
+
+## Migration workflow
+
+To migrate a subtree from one space to another:
+
+```bash
+# 1. List nodes in the source space
+lark-cli wiki +node-list --space-id source_space_id
```
fangshuyu-768, line 55: This example uses a token-like value (`wikcnSource`) without an explicit placeholder suffix, and the new checker does not catch it because it only matches quoted/backticked forms. That means the PR introduces exactly the kind of “realistic-looking sample” the script is supposed to prevent. I’d switch this to something clearly fake like `wikcn_EXAMPLE_TOKEN` (or `<source_node_token>`) and consider teaching the checker to catch bare code-block values too.

scripts/check-doc-tokens.sh
herbertliu: The `ou_manager` / `ou_director` false-positive was already addressed in an earlier commit (e98a6df's predecessor): `REALISTIC_TOKEN_RE` now requires at least one digit in the token suffix, so pure-letter fake names like `ou_manager` are excluded. Running `bash scripts/check-doc-tokens.sh` on the current tree returns ✅ with no errors.

shortcuts/wiki/wiki_node_copy.go
herbertliu: Fixed in e98a6df — added a mutual-exclusion check in `Validate`: if both `--target-space-id` and `--target-parent-node-token` are set, the command now returns `"--target-space-id and --target-parent-node-token are mutually exclusive; provide only one"`. Also added `TestWikiNodeCopyRejectsBothTargetFlags` to cover this case.

shortcuts/wiki/wiki_node_list.go
herbertliu: Fixed in e98a6df — removed the misleading `my_library` tip from `+node-list`. The shortcut doesn't resolve the alias (unlike `+node-create`), so advertising it was incorrect. Flag description and Tips now point users to run `+space-list` first to discover real space IDs.

skills/lark-wiki/references/lark-wiki-node-copy.md
herbertliu: Fixed in e98a6df — replaced `wikcnSource` with `wikcn_EXAMPLE_TOKEN` in the migration workflow example. Also switched the surrounding `--space-id` / `--target-space-id` values to `<source_space_id>` / `<target_space_id>` angle-bracket form for consistency.

shortcuts/wiki/wiki_node_copy.go
herbertliu: Already handled: `wiki_node_copy.go:47-49` rejects the case where both `--target-space-id` and `--target-parent-node-token` are set (mutually exclusive), and `TestWikiNodeCopyRejectsBothTargetFlags` locks this down explicitly. The earlier state only had the "at least one" check; the stricter guard was added before this comment was filed.

shortcuts/wiki/wiki_node_list.go
herbertliu: Fixed in aa035b9. `+node-list` now mirrors the `+node-create` contract: rejects `--space-id my_library` for bot identity, and for user identity auto-resolves the alias to the per-user real space_id via GET `/wiki/v2/spaces/my_library` before listing. Extracted `resolveMyLibrarySpaceID` into a shared helper so other wiki shortcuts can reuse it. Added two tests covering bot rejection and user resolution.

scripts/check-doc-tokens.sh
herbertliu: Already addressed in the current script: the regex requires at least one digit in the suffix (`[A-Za-z0-9]*[0-9][A-Za-z0-9]{3,}`), so pure-letter names like `ou_manager` / `ou_director` no longer match. `bash scripts/check-doc-tokens.sh` now runs clean on the whole repo. (aa035b9 additionally broadens the checker to catch bare code-block tokens, and I fixed one remaining realistic token in lark-minutes-search.md that it flagged.)

skills/lark-wiki/references/lark-wiki-node-copy.md
herbertliu: Fixed: the example now uses `wikcn_EXAMPLE_TOKEN` and `<source_space_id>` placeholders. The checker has also been broadened in aa035b9 to match bare token values (not only quoted/backticked forms) so similar realistic samples in future docs will be caught pre-merge.

shortcuts/wiki/wiki_node_copy.go
herbertliu: Fixed in 292a537 — Validate now explicitly rejects the case where both --target-space-id and --target-parent-node-token are set (line 48 of shortcuts/wiki/wiki_node_copy.go: 'mutually exclusive; provide only one'). Added test coverage in wiki_list_copy_test.go for that path.

shortcuts/wiki/wiki_node_list.go
herbertliu: Fixed in aab32b9 — +node-list now reuses resolveMyLibrarySpaceID (same helper +node-create uses) to resolve my_library before calling /wiki/v2/spaces/:space_id/nodes, and rejects --space-id my_library under --as bot (line 40: 'bot identity does not support --space-id my_library').

scripts/check-doc-tokens.sh
herbertliu: Fixed in 871627a and refined in aab32b9 — REALISTIC_TOKEN_RE was broadened to cover docx/ou/cli prefixes and backtick-quoted forms, and the script now skips clearly-placeholder forms (suffixes like _EXAMPLE_TOKEN, <...>, all-caps).

skills/lark-wiki/references/lark-wiki-node-copy.md
herbertliu: Fixed in 292a537 — wikcnSource replaced with clearly-fake placeholders (wikcn_EXAMPLE_TOKEN / <source_node_token>) throughout the example. The checker also now catches bare code-block token forms per the aab32b9 update.

herbertliu:
Thanks for the detailed feedback! All three points have been addressed in subsequent commits:

1. **`+node-copy` mutual exclusion** — added in `86812ff`: `Validate` now returns an error when both `--target-space-id` and `--target-parent-node-token` are provided, and a test `TestWikiNodeCopyRejectsBothTargetFlags` covers it.

2. **`+node-list` my_library resolution** — addressed in `ccb6dbd`: `Execute` now calls the shared `resolveMyLibrarySpaceID` helper (extracted from `+node-create`) before hitting the list endpoint. Bot identity is rejected early in `Validate` with a clear hint. Two new tests cover both paths.

3. **doc-token checker reliability** — fixed in `ccb6dbd`: broadened detection to bare tokens (not just quoted/backtick forms), and replaced the `wikcnSource` example in the reference docs with `wikcn_EXAMPLE_TOKEN`. The `TOKEN_BODY` regex no longer requires a digit, so purely-alpha suffixes like `Source` are now caught.

liujinkun2025:
Hi @herbertliu, thanks for the PR — `+space-list`, `+node-list`, and `+node-copy` look great. One concern about `+node-move` though: I think the existing `+move` already covers it.

Both call the same endpoint (`POST /open-apis/wiki/v2/spaces/{space_id}/nodes/{node_token}/move`) with the same body shape (`target_space_id` / `target_parent_token`). And `+move` in node mode (`shortcuts/wiki/wiki_move.go`) is actually a strict superset:

- Auto-resolves the source space from `--node-token`, so users don't need to pass `--space-id` explicitly
- Cross-validates `--target-space-id` against the target parent's actual space
- Rejects the `my_library` alias for `--as bot` with a clear error
- Already declares the `wiki:node:read` / `wiki:space:read` scopes needed for resolution

So `+node-move` ends up being a more verbose form of `lark-cli wiki +move --node-token ... --target-space-id ...`. Having two entry points for the same operation will likely confuse users.

Would you mind if I drop the `+node-move` commit from this PR? Happy to push the cleanup myself.

shortcuts/wiki/wiki_space_list.go
```diff
@@ -0,0 +1,205 @@
...
+	wikiSpaceListMaxPageSize     = 50
+)
+
+// WikiSpaceList lists all wiki spaces the caller has access to.
+var WikiSpaceList = common.Shortcut{
+	Service:     "wiki",
+	Command:     "+space-list",
+	Description: "List wiki spaces accessible to the caller",
+	Risk:        "read",
+	Scopes:      []string{"wiki:wiki:readonly"},
```
fangshuyu-768: Should this be `wiki:space:retrieve` instead? The repo's own permission table maps `spaces.list` to that scope in `skills/lark-wiki/SKILL.md`, and shortcut-declared scopes are used for unconditional preflight plus missing-scope hints. With the current `wiki:wiki:readonly` declaration, a token that already has the actual `spaces.list` grant can still be rejected locally or be told to request the wrong scope.

shortcuts/wiki/wiki_node_list.go
```diff
@@ -0,0 +1,215 @@
...
+	wikiNodeListMaxPageSize     = 50
+)
+
+// WikiNodeList lists child nodes in a wiki space or under a parent node.
+var WikiNodeList = common.Shortcut{
+	Service:     "wiki",
+	Command:     "+node-list",
+	Description: "List wiki nodes in a space or under a parent node",
+	Risk:        "read",
+	Scopes:      []string{"wiki:wiki:readonly"},
```
fangshuyu-768: Should this use `wiki:node:retrieve` for the actual list call, with the extra `my_library` resolution scope handled separately/conditionally? In this repo's own permission table, `nodes.list` maps to `wiki:node:retrieve`, and shortcut scope checks are exact-string preflight checks. As written, `wiki:wiki:readonly` can block callers before the API call even when they already have the real `nodes.list` grant.

fangshuyu-768 (approved):

PR #398, opened 2026-04-10
docs(readme): document shell completion command

OwenYWT:
Addresses #264

## Summary
- add a small README section for the hidden but working `lark-cli completion` subcommand
- keep the command hidden in help while making the supported shells discoverable in docs
- place the note in Advanced Usage so human operators can find it without changing agent defaults

## Test Plan
- [x] Docs only

## Summary by CodeRabbit

* **Documentation**
  * Added Shell Completion documentation with setup examples for zsh, bash, fish, and PowerShell.
  * Updated app visibility requirements for direct messages to clarify that user authentication does not bypass visibility restrictions.

* **Bug Fixes**
  * Fixed an issue where internal option identifiers were unintentionally exposed in option search results.

PR #399, opened 2026-04-10
docs(im): clarify app visibility for user sends

OwenYWT:
Addresses #161

## Summary
- document that `--as user` direct messages still depend on the app being visible to the recipient
- add the visibility caveat to the IM skill overview and `+messages-send` reference
- call out the common failure mode so agents do not assume user auth bypasses app visibility

## Test Plan
- [x] Docs only

## Summary by CodeRabbit

* **Bug Fixes**
  * Option identifiers no longer exposed in search results output.

* **Documentation**
  * Clarified that user token authentication does not bypass app visibility constraints when sending direct messages.


YangJunzhou-01:
Thanks for the PR!
This one overlaps with [#398](https://github.com/larksuite/cli/pull/398), so I'm closing it to avoid duplication. Feel free to reach out if you have any questions.

PR #419, opened 2026-04-11
feat(doc): add --file-view flag to +media-insert

nickzhangcomes:
## Summary

Add a `--file-view` flag to `docs +media-insert` so callers can choose how a file block is rendered: `card` (default, current behaviour), `preview` (inline player for audio/video), or `inline`. The value is plumbed into `file.view_type` when the block is created.

## Motivation

Today `docs +media-insert --type file` always creates the file block with the default `view_type=1` (card view). The docx open platform actually supports two other rendering modes for file blocks:

- `view_type=2` — preview view; renders audio/video files as an inline player
- `view_type=3` — inline view

These can only be set **at block creation time** — the PATCH `replace_file` endpoint does not accept `view_type`, so once the block is created with the default the caller cannot upgrade it afterwards. Without this flag, anyone who wants an inline audio/video player has to abandon the shortcut and manually re-implement the full 4-step orchestration (query root → create block → upload → bind + rollback) via `lark-cli api` calls.

## Changes

- `shortcuts/doc/doc_media_insert.go`
  - New `fileViewMap` (`card`/`preview`/`inline` → `1`/`2`/`3`).
  - New `--file-view` flag on the shortcut, validated in `Validate`:
    - Unknown values rejected with a clear error.
    - `--file-view` + `--type=image` rejected (images have their own rendering).
  - `buildCreateBlockData` gains a `fileViewType int` parameter; when `> 0` it sets `file.view_type` on the created child. `0` preserves the current behaviour (empty `file: {}` object), so servers still get exactly the same payload for existing callers.
  - `Execute` and `DryRun` thread the resolved view type through to `buildCreateBlockData`.
- `shortcuts/doc/doc_media_insert_test.go`
  - Existing `TestBuildCreateBlockDataUsesConcreteAppendIndex` and `TestBuildCreateBlockDataForFileIncludesFilePayload` updated to pass the new `0` argument; expected payload unchanged.
  - New tests:
    - `TestBuildCreateBlockDataForFileWithPreviewView`
    - `TestBuildCreateBlockDataForFileWithInlineView`
    - `TestBuildCreateBlockDataForImageIgnoresFileViewType` (verifies `view_type` never leaks into image blocks)
    - `TestFileViewMapCoversDocumentedValues` (guards the public enum surface)

## Usage

```bash
# Inline player rendering
lark-cli docs +media-insert --doc doxcnXXX --file ./audio.mp3 --type file --file-view preview

# Explicit card view (same as omitting --file-view)
lark-cli docs +media-insert --doc doxcnXXX --file ./report.pdf --type file --file-view card

# Inline view
lark-cli docs +media-insert --doc doxcnXXX --file ./data.csv --type file --file-view inline
```

## Backward compatibility

Fully backward compatible:

- New flag is optional with no default.
- When `--file-view` is omitted, `fileViewMap[""] == 0`, which skips the new `if fileViewType > 0` branch — the request body sent to the server is byte-for-byte identical to the current code.
- Existing unit tests still pass (after the signature update), and the expected JSON payload for the default cases is unchanged.

## Test plan

- [x] `go test ./shortcuts/doc/...` — all existing and new tests pass
- [x] `go vet ./...` — clean
- [x] `gofmt -l` — clean
- [x] Manual end-to-end: `--file-view preview` with an `--type=file` upload — confirmed the resulting file block is wrapped in a `block_type=33` view with `view.view_type=2`, as expected by the open platform spec
- [x] Manual end-to-end: omitting `--file-view` — request body identical to current behaviour
- [x] `--file-view bogus` → validation error
- [x] `--type image --file-view preview` → validation error

## Relationship to #335

#335 (`feat(doc): add --selection-with-ellipsis position flag to +media-insert`) also extends `docs +media-insert` with a new flag, but along an orthogonal axis — positioning (where to insert) vs. rendering (how it is displayed). The two changes do not overlap semantically. If #335 lands first I am happy to rebase; conflicts should be limited to the `Flags` slice and the `Execute`/`DryRun` local variable blocks.

## Summary by CodeRabbit

* **New Features**
  * Added a file view option (--file-view) when inserting DOCX files—choose card, preview, or inline to control rendering.
* **Bug Fixes**
  * Validation now enforces that file view options are only accepted for file inserts and rejected otherwise.
* **Tests**
  * Added tests covering view mappings, file vs. image behavior, and validation of allowed view options.

### DIFF
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -20,6 +20,18 @@ var alignMap = map[string]int{
 	"right":  3,
 }
 
+// fileViewMap maps the user-facing --file-view value to the docx File block
+// `view_type` enum. The underlying values come from the open platform spec:
+//
+//	1 = card view (default)
+//	2 = preview view (renders audio/video files as an inline player)
+//	3 = inline view
+var fileViewMap = map[string]int{
+	"card":    1,
+	"preview": 2,
+	"inline":  3,
+}
+
 var DocMediaInsert = common.Shortcut{
 	Service:     "docs",
 	Command:     "+media-insert",
@@ -33,6 +45,7 @@ var DocMediaInsert = common.Shortcut{
 		{Name: "type", Default: "image", Desc: "type: image | file"},
 		{Name: "align", Desc: "alignment: left | center | right"},
 		{Name: "caption", Desc: "image caption text"},
+		{Name: "file-view", Desc: "file block rendering: card (default) | preview | inline; only applies when --type=file. preview renders audio/video as an inline player"},
 	},
 	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
 		docRef, err := parseDocumentRef(runtime.Str("doc"))
@@ -41,6 +54,14 @@ var DocMediaInsert = common.Shortcut{
 		}
 		if docRef.Kind == "doc" {
 			return output.ErrValidation("docs +media-insert only supports docx documents; use a docx token/URL or a wiki URL that resolves to docx")
+		}
+		if view := runtime.Str("file-view"); view != "" {
+			if _, ok := fileViewMap[view]; !ok {
+				return output.ErrValidation("invalid --file-view value %q, expected one of: card | preview | inline", view)
+			}
+			if runtime.Str("type") != "file" {
+				return output.ErrValidation("--file-view only applies when --type=file")
+			}
 		}
 		return nil
 	},
@@ -55,9 +76,10 @@ var DocMediaInsert = common.Shortcut{
 		filePath := runtime.Str("file")
 		mediaType := runtime.Str("type")
 		caption := runtime.Str("caption")
+		fileViewType := fileViewMap[runtime.Str("file-view")]
 
 		parentType := parentTypeForMediaType(mediaType)
-		createBlockData := buildCreateBlockData(mediaType, 0)
+		createBlockData := buildCreateBlockData(mediaType, 0, fileViewType)
 		createBlockData["index"] = "<children_len>"
 		batchUpdateData := buildBatchUpdateData("<new_block_id>", mediaType, "<file_token>", runtime.Str("align"), caption)
 
@@ -92,6 +114,7 @@ var DocMediaInsert = common.Shortcut{
 		mediaType := runtime.Str("type")
 		alignStr := runtime.Str("align")
 		caption := runtime.Str("caption")
+		fileViewType := fileViewMap[runtime.Str("file-view")]
 
 		documentID, err := resolveDocxDocumentID(runtime, docInput)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


shortcuts/doc/doc_media_insert.go
```diff
@@ -42,6 +55,14 @@ var DocMediaInsert = common.Shortcut{
 		if docRef.Kind == "doc" {
 			return output.ErrValidation("docs +media-insert only supports docx documents; use a docx token/URL or a wiki URL that resolves to docx")
 		}
+		if view := runtime.Str("file-view"); view != "" {
```
fangshuyu-768, line 58: Nice addition overall. One gap I noticed is that the new tests only exercise `buildCreateBlockData`, but the main user-facing contract here lives in `Validate`: rejecting unknown `--file-view` values and rejecting `--file-view` when `--type != file`.

Could we add focused tests for those validation paths as well? Otherwise a future refactor could silently relax the CLI contract without any test catching it.

shortcuts/doc/doc_media_insert_test.go
```diff
@@ -44,6 +44,88 @@ func TestBuildCreateBlockDataForFileIncludesFilePayload(t *testing.T) {
 	}
 }
 
+func TestBuildCreateBlockDataForFileWithPreviewView(t *testing.T) {
```
fangshuyu-768, line 78: Could we add coverage for the explicit `--file-view card` path too?

Omitting `--file-view` keeps the legacy payload (`file: {}`), but `--file-view card` now sends `file.view_type = 1`. Those are different request shapes, even if they are intended to be semantically equivalent. A small unit test here would help lock down that public flag behavior.

shortcuts/doc/doc_media_insert.go
nickzhangcomes: Good call, that gap was bugging me too after you pointed it out. Pushed a table-driven test in 3669613 that runs Validate directly and covers the two rejection branches (unknown value like --file-view bogus, and --file-view preview with --type image), plus the accepted shapes as positive cases so the whole guard is locked down. Future Validate refactors will have to touch these explicitly.

shortcuts/doc/doc_media_insert_test.go
nickzhangcomes: Agreed, and the divergence is on purpose: omitting the flag keeps the legacy file: {} shape for backward compat, while --file-view card sends an explicit view_type: 1 because the user literally asked for card. I didn't want to silently drop the field and make the flag non-idempotent with its own echo. Added TestBuildCreateBlockDataForFileWithCardView in 3669613 so that expectation is pinned, not just reviewer memory.

fangshuyu-768:
Thanks again for the thoughtful contribution and for taking the time to address the feedback.

The latest changes look good, and I really appreciate the follow-up updates and added test coverage.

We’d also love to have you keep following the project as it evolves. Please feel free to continue sharing any feedback, suggestions, or ideas, they’re always very welcome.

fangshuyu-768 (dismissed):

fangshuyu-768 (approved):

PR #422, opened 2026-04-11
feat(sheets): add filter view and condition shortcuts

caojie0621:
## Summary
  Add 10 new `sheets` shortcuts for filter view and filter condition CRUD, covering the full lifecycle of spreadsheet filter views.
                                                                                                                                                                                             
  ## Changes
                                                                                                                                                                                             
  ### Filter views (5)                                      
  - `+create-filter-view`: create a filter view with range, optional name and custom ID
  - `+update-filter-view`: update filter view range or name                                                                                                                                  
  - `+list-filter-views`: list all filter views in a sheet
  - `+get-filter-view`: get a filter view by ID                                                                                                                                              
  - `+delete-filter-view`: delete a filter view             
                                                                                                                                                                                             
  ### Filter view conditions (5)                            
  - `+create-filter-view-condition`: create a filter condition on a column (number/text/color/hiddenValue)
  - `+update-filter-view-condition`: update a filter condition                                                                                                                               
  - `+list-filter-view-conditions`: list all conditions of a filter view
  - `+get-filter-view-condition`: get a condition by column                                                                                                                                  
  - `+delete-filter-view-condition`: delete a condition     
                                                                                                                                                                                             
  ### Other                                                                                                                                                                                  
  - Unit tests for all 10 shortcuts (39 cases, 88-93% coverage)
  - Skill reference docs for all 10 shortcuts                                                                                                                                                
                                                            
  ## Test Plan                                                                                                                                                                               
  - [x] Unit tests pass (`go test ./shortcuts/sheets/`)
  - [x] Manual local verification confirms all 10 `lark-cli sheets +xxx` commands work as expected                                                                                           
                                                                                                                                                                                             
  ## Related Issues
  - None        

## Summary by CodeRabbit

* **New Features**
  * Added 10 new Sheets commands to manage filter views and their conditions (create, update, list, get, delete).

* **Documentation**
  * Added reference docs for each new command with examples, parameters, usage notes, and expected JSON outputs.

* **Tests**
  * Added an extensive test suite covering validation rules, dry-run/execute behavior, error cases, request/response shapes, and URL-based token handling.


shortcuts/sheets/sheet_filter_view.go
```diff
@@ -0,0 +1,233 @@
...
+	AuthTypes:   []string{"user", "bot"},
+	Flags: []common.Flag{
+		{Name: "url", Desc: "spreadsheet URL"},
+		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
+		{Name: "sheet-id", Desc: "sheet ID", Required: true},
+		{Name: "filter-view-id", Desc: "filter view ID", Required: true},
+		{Name: "range", Desc: "new filter range"},
+		{Name: "filter-view-name", Desc: "new display name (max 100 chars)"},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
```
fangshuyu-768, line 99: Should we add a CLI-side validation here to require at least one update field?

Right now `+update-filter-view` can still send an empty body if the user only passes `--sheet-id` and `--filter-view-id`. In similar update-style shortcuts in this repo, we usually fail early in the CLI instead of sending a no-op request to the API.

It may be better to enforce that at least one of `--range` or `--filter-view-name` is provided in `Validate`.

shortcuts/sheets/sheet_filter_view_condition.go
```diff
@@ -0,0 +1,239 @@
...
+		{Name: "url", Desc: "spreadsheet URL"},
+		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
+		{Name: "sheet-id", Desc: "sheet ID", Required: true},
+		{Name: "filter-view-id", Desc: "filter view ID", Required: true},
+		{Name: "condition-id", Desc: "column letter (e.g. E)", Required: true},
+		{Name: "filter-type", Desc: "filter type: hiddenValue, number, text, color"},
+		{Name: "compare-type", Desc: "comparison operator"},
+		{Name: "expected", Desc: "filter values JSON array"},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
```
fangshuyu-768, line 82: It may also be worth adding a CLI-side validation here to require at least one update field.

At the moment `+update-filter-view-condition` can generate an empty body when none of `--filter-type`, `--compare-type`, or `--expected` is provided, so the user only gets feedback from the server side.

Suggest checking in `Validate` that at least one updatable field is explicitly set.

shortcuts/sheets/sheet_filter_view_condition.go
```diff
@@ -0,0 +1,239 @@
...
+	if includeConditionID {
+		body["condition_id"] = runtime.Str("condition-id")
+	}
+	if s := runtime.Str("filter-type"); s != "" {
+		body["filter_type"] = s
+	}
+	if s := runtime.Str("compare-type"); s != "" {
+		body["compare_type"] = s
+	}
+	if s := runtime.Str("expected"); s != "" {
```
fangshuyu-768, line 254: Is the `--expected` parsing here a bit too permissive?

From the flag description and the skill docs, `expected` looks like it is intended to be a JSON array, but this implementation currently:
1. accepts any JSON value (for example object, number, or string)
2. silently falls back to `[]string{s}` when the input is not valid JSON

That makes the CLI behavior differ from the documented contract, and it may also hide user input errors. Would it be better to validate `--expected` explicitly in `Validate` as a proper JSON array, instead of auto-coercing invalid input into a single-element array?

fangshuyu-768 (approved):

PR #445, opened 2026-04-13
feat: support LARKSUITE_CLI_PROFILE env var for multi-app selection

meijing0114:
## Summary

- Add `LARKSUITE_CLI_PROFILE` environment variable to select the active app profile without the `--profile` CLI flag
- Add `FindAppByID()` and `ActiveApp()` helpers to `internal/core/config.go` for env-var-aware profile resolution
- Wire env var into `BootstrapInvocationContext` so all commands (including `config show`, `auth logout`, `auth list`, `config default-as`) respect it

### Motivation

When running multiple lark-cli bots on the same machine (e.g. one for an AI agent platform, one for Claude Code), each agent needs its own app configuration. The `--profile` flag works for interactive use, but agents run in fixed environments where an environment variable is the natural selector.

### Resolution Priority

`--profile` flag > `LARKSUITE_CLI_PROFILE` env var > `config.CurrentApp` field > `Apps[0]`

### Usage

```bash
# Each agent sets its env var
export LARKSUITE_CLI_PROFILE=my-bot-name    # profile name
export LARKSUITE_CLI_PROFILE=cli_aXXX       # or app ID

# All commands respect it
lark-cli config show
lark-cli auth login
lark-cli api GET /open-apis/...
```

### Backward Compatibility

100%. Users who don't set the env var get identical behavior.

## Test plan

- [x] `FindAppByID` found/not-found tests
- [x] `ActiveApp` with env var / without / not-found tests  
- [x] `RequireConfig` integration tests with env var selection
- [x] `BootstrapInvocationContext` env var fallback + flag-overrides-env tests
- [x] Full unit test suite (3134 tests across 55 packages)
- [x] Manual smoke test: `config show` with/without env var

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a default profile via the `LARKSUITE_CLI_PROFILE` environment variable.
  * Command-line flags now take precedence over the environment variable when both are specified.
  * Improved error messaging when apps or profiles cannot be found.

### DIFF
--- a/cmd/bootstrap.go
+++ b/cmd/bootstrap.go
@@ -6,8 +6,10 @@ package cmd
 import (
 	"errors"
 	"io"
+	"os"
 
 	"github.com/larksuite/cli/internal/cmdutil"
+	"github.com/larksuite/cli/internal/envvars"
 	"github.com/spf13/pflag"
 )
 
@@ -26,5 +28,9 @@ func BootstrapInvocationContext(args []string) (cmdutil.InvocationContext, error
 	if err := fs.Parse(args); err != nil && !errors.Is(err, pflag.ErrHelp) {
 		return cmdutil.InvocationContext{}, err
 	}
-	return cmdutil.InvocationContext{Profile: globals.Profile}, nil
+	profile := globals.Profile
+	if profile == "" {
+		profile = os.Getenv(envvars.CliProfile)
+	}
+	return cmdutil.InvocationContext{Profile: profile}, nil
 }
--- a/cmd/bootstrap_test.go
+++ b/cmd/bootstrap_test.go
@@ -61,6 +61,28 @@ func TestBootstrapInvocationContext_ShortHelp(t *testing.T) {
 	}
 }
 
+func TestBootstrapInvocationContext_EnvVarFallback(t *testing.T) {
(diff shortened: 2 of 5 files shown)


sang-neo03:
@meijing0114 
  Direction makes sense — env-var profile selection is a real need for multi-agent setups, and backward compat is clean. A few blockers before merge:                                 
                                                                                                                                                                                      
  1. FindAppByID is dead code — resolution goes through ActiveApp → CurrentAppConfig → FindApp, which already handles Name + AppId. Only its own tests reference it. Please remove.
  2. Env var read twice — bootstrap.go already copies it into InvocationContext.Profile, so the os.Getenv inside ActiveApp never fires on the CLI path. Either drop it, or add a doc  
  comment that it's a fallback for library callers bypassing bootstrap.                                                                                                               
  3. Needs rebase — envvars.go on main now has CliAuthProxy / CliProxyKey after CliStrictMode; please group CliProfile with the other identity consts.                                
                                                                                                                                                                                      
  Nits: the replaced test DoesNotUseEnvProfileFallback was a deliberate guard from #252 — worth calling that reversal out in the commit message. And config show could note when      
  LARKSUITE_CLI_PROFILE is overriding the on-disk currentApp.                                                                                                                         
                                                                                                                                                                                      
  Tests look good. LGTM once 1–3 are resolved.

PR #446, opened 2026-04-13
fix(im): unify messages-search pagination int flags

YangJunzhou-01:
## Summary

Unify `lark-cli im +messages-search` pagination flags to use int semantics consistently.

Previously, `page-limit` was registered as an int flag while `page-size` was still handled as a string flag and parsed manually. This led to inconsistent runtime behavior inside the same shortcut and allowed test helpers to drift from the real CLI flag registration.

This PR keeps the `page-limit` fix and further updates `+messages-search` so that both `page-limit` and `page-size` are handled as int flags. The implementation now reads:

- `page-limit` via `runtime.Int("page-limit")`
- `page-size` via `runtime.Int("page-size")`

Related test helpers and coverage tests are also aligned with the real CLI int flag registration for this shortcut.

## Scope

- Only affects `lark-cli im +messages-search`
- No shared/common modules changed
- No user-facing change for valid pagination usage
- Default pagination behavior remains unchanged
- `--page-all` behavior remains unchanged
- `page-size > 50` is still capped at 50
- Invalid `--page-size` values now fail earlier at flag parsing instead of business validation

## Test Plan

```bash
go test ./shortcuts/im -count=1
```

Also verified locally that valid usages remain consistent, including:

```bash
lark-cli im +messages-search --as user --chat-id <chat_id> --query 的 --page-size=1 --page-limit=2
```

and quoted numeric values such as:

```bash
lark-cli im +messages-search --as user --chat-id <chat_id> --query 的 --page-size="1" --page-limit='2'
```

continue to work as expected in normal shell usage.

## Summary by CodeRabbit

* **Refactor**
  * Pagination flags for message search are now treated as numeric, with stricter integer validation, clamping of page-size to the maximum, and explicit page-limit bounds (defaults unchanged).

* **Tests**
  * Updated test setup and coverage to reflect the numeric pagination behavior and validate pagination-related scenarios.

### DIFF
--- a/shortcuts/im/builders_test.go
+++ b/shortcuts/im/builders_test.go
@@ -29,7 +29,42 @@ func newTestRuntimeContext(t *testing.T, stringFlags map[string]string, boolFlag
 	t.Helper()
 
 	cmd := &cobra.Command{Use: "test"}
+	cmd.Flags().Int("page-limit", 20, "")
 	for name := range stringFlags {
+		if name == "page-limit" {
+			continue
+		}
+		cmd.Flags().String(name, "", "")
+	}
+	for name := range boolFlags {
+		cmd.Flags().Bool(name, false, "")
+	}
+	if err := cmd.ParseFlags(nil); err != nil {
+		t.Fatalf("ParseFlags() error = %v", err)
+	}
+	for name, val := range stringFlags {
+		if err := cmd.Flags().Set(name, val); err != nil {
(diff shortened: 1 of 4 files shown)


haozhenghua-code (approved):

haozhenghua-code (approved):
get

liangshuo-1 (approved):

PR #450, opened 2026-04-13
feat(slides): add image upload via +media-upload and @path placeholders in +create

ViperCai:
## Summary

Adds first-class image support to the slides skill. Users (and AI agents) can now upload local images into a presentation and embed them via XML, either by reusing a returned `file_token` or by writing `` placeholders inside `+create --slides` and letting the CLI auto-upload + rewrite.

## Changes

- New `slides +media-upload` shortcut: uploads a local image to a presentation and returns its `file_token`. Validates `--file` is a CWD-relative safe path; uses `slide_file` as `parent_type` (the only value the slides upload endpoint accepts).
- `slides +create --slides` now scans each slide XML for `` placeholders, uploads the referenced files via the same code path, and rewrites `src` to the resulting `file_token` before creating slides.
- Skill docs (`skills/lark-slides/...`) expanded:
  - Surfaces the upload flow up-front in `SKILL.md` so it is discoverable.
  - Documents image crop behavior (`width:height` must match original aspect or the image is cropped).
  - Adds reusable image layout templates in `references/slide-templates.md`.
  - Documents the whole-page-replacement recipe for adding an image to an existing page (no element-level edit API), inline in `SKILL.md`.
  - Explicitly warns that `before_slide_id` belongs in `--data` body — putting it in `--params` is silently forwarded as an unknown query param and the new slide ends up appended to the end instead of inserted at the requested position.

## Test Plan

- [x] Unit tests pass: `go test ./shortcuts/slides/... ./cmd/service/...`
- [x] Manual local verification:
  - `lark-cli slides +create --slides '[......]'` uploads and embeds the image
  - `lark-cli slides +media-upload --file ./pic.png --presentation $PID` returns a `file_token`
  - Real API insert with `before_slide_id` in `--data` lands the new slide at the requested position; placing it in `--params` (the documented anti-pattern) silently appends to the end, confirming the warning is needed
- [x] `node scripts/skill-format-check/index.js` passes

## Related Issues

- None

## Release Notes

* **New Features**
  * Added image upload support for slides with automatic placeholder replacement in slide creation.
  * New media upload command enables uploading images to existing presentations.

* **Documentation**
  * Added comprehensive guides for image workflows and media upload usage.
  * Added image-based slide layout templates with examples.
  * Updated slide creation documentation with path validation and image handling constraints.


fangshuyu-768 (dismissed):

shortcuts/slides/slides_create.go
```diff
@@ -27,10 +28,13 @@ var SlidesCreate = common.Shortcut{
 	Description: "Create a Lark Slides presentation",
 	Risk:        "write",
 	AuthTypes:   []string{"user", "bot"},
-	Scopes:      []string{"slides:presentation:create", "slides:presentation:write_only"},
+	// docs:document.media:upload is required when --slides contains <img src="@./...">
+	// placeholders. Declared unconditionally so the pre-flight scope check fails fast
+	// instead of leaving an orphaned empty presentation when the in-flight upload 403s.
+	Scopes: []string{"slides:presentation:create", "slides:presentation:write_only", "docs:document.media:upload"},
```
fangshuyu-768, line 36: This changes the scope contract of `slides +create` for all callers, not just the new image-placeholder path. Existing `+create` flows that do not upload media will now fail preflight unless `docs:document.media:upload` is granted, which looks like a backward compatibility regression. Could we avoid requiring the media-upload scope unless `--slides` actually contains `@path` image placeholders?

fangshuyu-768 (commented):
A few inline follow-ups on scopes and placeholder matching.

shortcuts/slides/slides_media_upload.go
```diff
@@ -0,0 +1,180 @@
...
+//
+// This is the atomic building block for getting a local image into a slides
+// deck. Higher-level shortcuts (e.g. +create with @path placeholders) reuse
+// the same upload helpers.
+var SlidesMediaUpload = common.Shortcut{
+	Service:     "slides",
+	Command:     "+media-upload",
+	Description: "Upload a local image to a slides presentation and return the file_token (use as <img src=...>)",
+	Risk:        "write",
+	Scopes:      []string{"docs:document.media:upload", "wiki:node:read"},
```
fangshuyu-768: `wiki:node:read` seems to be required only for the wiki-URL branch, but it is declared unconditionally here. That means direct `xml_presentation_id` inputs and `/slides/` URLs now require an extra scope they never use. Can we keep this shortcut usable with only `docs:document.media:upload` unless `--presentation` is actually a wiki link?

shortcuts/slides/helpers.go
```diff
@@ -0,0 +1,173 @@
...
+// replace with file_token".
+//
+// Match groups:
+//
+//	1: opening quote character (so we can replace symmetrically)
+//	2: the path string (everything inside the quotes after the leading @)
+//
+// We deliberately scope to <img ... src="@..."> rather than any src= so other
+// schema elements (like icon/iconType) aren't accidentally rewritten.
+var imgSrcPlaceholderRegex = regexp.MustCompile(`(?s)<img\b[^>]*?\bsrc=(["'])@([^"']+)(["'])`)
```
fangshuyu-768: The placeholder regex only matches `src="@..."` / `src='@...'` exactly. Valid XML with optional whitespace, such as ``, would not be detected or rewritten, so the raw `@path` could reach `slide.create`. Could we make the matcher tolerate optional whitespace around `=` and add a regression test for that form?

shortcuts/slides/slides_create.go
```diff
@@ -41,6 +42,17 @@ var SlidesCreate = common.Shortcut{
...
+			// only to fail mid-way on a missing local file.
+			for _, path := range extractImagePlaceholderPaths(slides) {
+				stat, err := runtime.FileIO().Stat(path)
+				if err != nil {
+					return common.WrapInputStatError(err, fmt.Sprintf("--slides @%s: file not found", path))
+				}
+				if !stat.Mode().IsRegular() {
+					return common.FlagErrorf("--slides @%s: must be a regular file", path)
+				}
+			}
```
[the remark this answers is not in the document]
ViperCai: Not an issue — `runtime.FileIO().Stat()` is backed by `LocalFileIO.Stat()` which internally calls `SafeInputPath` before touching the filesystem (see `localfileio.go:35-48`). Absolute paths and `..` traversal are rejected at that layer. This is the same protection sheets/mail get via explicit `validate.SafeInputPath()`, just routed through the FileIO abstraction.

shortcuts/slides/slides_media_upload.go
```diff
@@ -0,0 +1,180 @@
...
+	Flags: []common.Flag{
+		{Name: "file", Desc: "local image path (files > 20MB use multipart upload automatically)", Required: true},
+		{Name: "presentation", Desc: "xml_presentation_id, slides URL, or wiki URL that resolves to slides", Required: true},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		if _, err := parsePresentationRef(runtime.Str("presentation")); err != nil {
+			return err
+		}
+		return nil
+	},
```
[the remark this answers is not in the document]
ViperCai: Same as above — `runtime.FileIO().Stat(filePath)` in the Validate/Execute paths goes through `LocalFileIO.Stat()` → `SafeInputPath()`. Path traversal and absolute paths are already rejected. No additional explicit call needed.

shortcuts/slides/slides_create.go
ViperCai: Checked with project convention — all shortcuts in this repo declare scopes unconditionally upfront (e.g. wiki_move, drive_import). The CLI guides users to `auth login --scope` when missing. Keeping it unconditional here to stay consistent. Same pattern as wiki_move declaring `wiki:node:read` even when the input is a direct doc token.

shortcuts/slides/slides_media_upload.go
ViperCai: Same reasoning as above — unconditional scope declaration is the project convention. The comment on line 37-38 documents why it is declared upfront. CLI handles the auth guidance automatically.

shortcuts/slides/helpers.go
ViperCai: Fixed — regex now uses `\s*=\s*` to tolerate whitespace around `=`. Added regression tests for both `extractImagePlaceholderPaths` and `replaceImagePlaceholders` with `src = "@..."` form (see helpers_test.go).

shortcuts/slides/slides_create.go
```diff
@@ -27,10 +28,15 @@ var SlidesCreate = common.Shortcut{
 	Description: "Create a Lark Slides presentation",
 	Risk:        "write",
 	AuthTypes:   []string{"user", "bot"},
-	Scopes:      []string{"slides:presentation:create", "slides:presentation:write_only"},
+	// docs:document.media:upload is required by the @-placeholder upload path.
+	// Declared up-front (matching the convention used by other multi-API shortcuts
+	// like wiki_move) so the pre-flight check fails fast and lark-cli's
+	// auth login --scope hint guides the user, instead of leaving an orphaned
+	// empty presentation when the in-flight upload 403s.
+	Scopes: []string{"slides:presentation:create", "slides:presentation:write_only", "docs:document.media:upload"},
```
fangshuyu-768, line 36: This makes `docs:document.media:upload` mandatory for every `slides +create` call, even when `--slides` contains no `@path` placeholders at all.

That looks like a behavior regression for plain text / no-image create flows: users who only want to create a presentation now need an upload scope they never exercise. I think this scope should be required only when placeholder images are actually present.

shortcuts/slides/slides_media_upload.go
```diff
@@ -0,0 +1,183 @@
...
+// the same upload helpers.
+var SlidesMediaUpload = common.Shortcut{
+	Service:     "slides",
+	Command:     "+media-upload",
+	Description: "Upload a local image to a slides presentation and return the file_token (use as <img src=...>)",
+	Risk:        "write",
+	// wiki:node:read is required by the wiki-URL resolution path. Declared
+	// up-front (matching the convention used by other multi-API shortcuts) so
+	// users without it get the standard auth login --scope hint at pre-flight.
+	Scopes:    []string{"docs:document.media:upload", "wiki:node:read"},
```
fangshuyu-768, line 41: `wiki:node:read` is declared unconditionally here, but this shortcut also supports a raw presentation token and a `/slides/` URL that never go through wiki resolution.

As written, those simpler flows still require wiki scope up front, which seems broader than necessary. It would be better if wiki scope were required only for the wiki-URL path.

shortcuts/slides/slides_create.go
ViperCai: Intentional design trade-off — `Scopes` is a static field on the `Shortcut` struct; the framework has no dynamic-scope mechanism today.

Declaring it up-front matches the convention used by other multi-API shortcuts (e.g. `wiki_move`) and avoids a worse UX: without it, a user who passes `@path` placeholders would first create the presentation, then 403 on the upload call, leaving an orphaned empty presentation. The pre-flight scope check catches this early and surfaces the `auth login --scope` hint.

The cost of over-declaring is minimal (one extra scope grant during initial auth).

shortcuts/slides/slides_media_upload.go
ViperCai: Same rationale — `Scopes` is static and cannot be conditioned on flag values at declaration time.

`wiki:node:read` is declared up-front so that users who pass a wiki URL get a clear pre-flight `auth login --scope` hint instead of a runtime 403 mid-operation. For users who pass a raw presentation token or `/slides/` URL, the wiki API is simply never called — the extra scope declaration has no runtime effect beyond the initial auth grant.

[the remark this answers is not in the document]
ViperCai: By design. `Scopes` is a static declaration — the framework doesn't support conditional scopes based on flag values.

The `wiki:node:read` scope is declared up-front so users who pass a wiki URL get a clear pre-flight auth hint. For non-wiki inputs (raw token or `/slides/` URL), the wiki API is never called and the scope has no runtime effect.

fangshuyu-768 (approved):

PR #461, opened 2026-04-14
feat(sheets): add dropdown shortcuts and formula reference docs

caojie0621:
## Summary

  The `multipleValue` cell type silently becomes plain text when the target range
  has no dropdown configuration, and the prerequisite data-validation API was not
  exposed as a CLI shortcut. This PR adds four new shortcuts wrapping the v2
  `dataValidation` endpoints and documents Lark-specific formula writing rules.

  ## Changes

  - Add `+set-dropdown`, `+update-dropdown`, `+get-dropdown`, `+delete-dropdown`
    shortcuts in `shortcuts/sheets/sheet_dropdown.go`
  - Register shortcuts in `shortcuts/sheets/shortcuts.go`
  - Add 22 unit tests in `shortcuts/sheets/sheet_dropdown_test.go`
  - Add 4 reference docs (`lark-sheets-{set,update,get,delete}-dropdown.md`)
  - Add `lark-sheets-formula.md` reference for ARRAYFORMULA, native array
    functions, MAP/LAMBDA, date diff, and unsupported Excel syntax
  - Update `SKILL.md`: add dropdown shortcut table, link formula reference,
    clarify `multipleValue` prerequisite
  - Sync `skill-template/domains/sheets.md` with latest changes

  ## Test Plan

  - [x] `go build` and `go vet` pass
  - [x] 22 unit tests pass (`go test ./shortcuts/sheets/ -run Dropdown`)
  - [x] All existing sheets tests pass (no regression)
  - [x] E2E verification against live spreadsheet:
    - `+set-dropdown` with `--multiple --highlight --colors`
    - `+write` with `multipleValue` → read back confirms dropdown values
    - `+get-dropdown` returns correct config
    - `+update-dropdown` changes options from 高/中/低 to P0-P3
    - `+delete-dropdown` removes config, `+get-dropdown` confirms empty
  - [x] 13 boundary scenarios verified (missing flags, invalid JSON, empty array,
    comma in value, colors mismatch, dry-run, no-dropdown range, etc.)

  ## Related Issues

  N/A

## Summary by CodeRabbit

* **New Features**
  * Added four Sheets dropdown commands: +set-dropdown, +update-dropdown, +get-dropdown, +delete-dropdown — create, modify, query, and remove dropdowns with support for multi-select, validation highlighting, per-option color mapping, fully-qualified ranges, and a --dry-run preview.

* **Documentation**
  * Added a Feishu Sheets formula reference and detailed docs for setting/updating/getting/deleting dropdowns, usage examples, limits, and dry-run/output formats.


shortcuts/sheets/sheet_dropdown.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return token, nil
+}
+
+func parseJSONStringArray(flagName, value string) ([]interface{}, error) {
+	var typed []string
+	if err := json.Unmarshal([]byte(value), &typed); err != nil {
+		return nil, common.FlagErrorf("--%s must be a JSON array of strings: %v", flagName, err)
+	}
+	if typed == nil {
```
fangshuyu-768, line 40: This rejects `null`, but it still accepts `[]`.

For flags like `--condition-values` and `--ranges`, an empty array does not seem meaningful. Consider rejecting empty arrays here (or at the call sites) so the CLI fails early with a clearer validation error.

shortcuts/sheets/sheet_dropdown.go
```diff
@@ -0,0 +1,307 @@
...
+	Flags: []common.Flag{
+		{Name: "url", Desc: "spreadsheet URL"},
+		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
+		{Name: "range", Desc: "cell range (<sheetId>!A2:A100)", Required: true},
+		{Name: "condition-values", Desc: `dropdown options as JSON array (e.g. '["opt1","opt2"]'), max 500, each <=100 chars, no commas`, Required: true},
+		{Name: "multiple", Desc: "enable multi-select (default false)", Type: "bool"},
+		{Name: "highlight", Desc: "color-code options (default false)", Type: "bool"},
+		{Name: "colors", Desc: `RGB hex color array (e.g. '["#1FB6C1","#F006C2"]'), must match condition-values length`},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
```
fangshuyu-768, line 121: This validation path does not validate `--range` itself.

Since this command does not accept `--sheet-id`, it looks like `--range` should always be a fully qualified sheet range (for example, `<sheetId>!A2:A100`). Otherwise malformed inputs can pass CLI validation and fail only at the API layer.

shortcuts/sheets/sheet_dropdown.go
```diff
@@ -0,0 +1,307 @@
...
+		{Name: "condition-values", Desc: `dropdown options as JSON array (e.g. '["opt1","opt2"]')`, Required: true},
+		{Name: "multiple", Desc: "enable multi-select (default false)", Type: "bool"},
+		{Name: "highlight", Desc: "color-code options (default false)", Type: "bool"},
+		{Name: "colors", Desc: `RGB hex color array, must match condition-values length`},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		if _, err := validateDropdownToken(runtime); err != nil {
+			return err
+		}
+		if _, err := parseJSONStringArray("ranges", runtime.Str("ranges")); err != nil {
```
fangshuyu-768: This only checks that `--ranges` is a JSON array of strings, but it does not validate that each item is a valid sheet range.

It would be better to reject malformed ranges locally here instead of forwarding them to the API and relying on server-side validation.

fangshuyu-768 (approved):

PR #467, opened 2026-04-14
feat(doc): warn when callout uses type= without background-color

herbertliu:
## Summary

When a user writes `<callout type="warning" emoji="📝">` in markdown passed to `docs +create` / `docs +update`, the Feishu side renders the callout block with **no color** — `type=` is a semantic shorthand the `create-doc` API does not consume; it only understands `background-color=` / `border-color=`.

This PR is a **diagnostic, not a renderer fix**: it does not rewrite the markdown. Instead it scans the input and, when `type=` is used without an explicit `background-color=`, prints one hint line on stderr telling the user exactly which color attributes to write to get the color they probably wanted. The next manual edit fixes the document; the markdown reaching the API is byte-identical to what the user passed.

After review by [@SunPeiYang996](https://github.com/larksuite/cli/pull/467#issuecomment-4243136139) the original approach (silently expanding `type=` into the matching color pair inside the shortcut) was rejected as protocol-adaptation logic in the wrong layer. The hint-on-stderr design adopted here is option 3 of his suggested alternatives.

## Behavior

```
$ lark-cli docs +create --markdown '<callout type="warning">x</callout>'
hint: callout type="warning" has no background-color; consider: background-color="light-yellow" border-color="yellow"
{
  "ok": true,
  ...
}
```

- Markdown reaching `create-doc` is unmodified.
- Stdout JSON envelope, exit code, and dry-run renderer are all unchanged.
- Feishu rendering is unchanged from the pre-PR baseline (colorless callout); the user has to follow the hint and rewrite to get color. The PR's value is turning a silent failure into an actionable, copy-pasteable suggestion.

## Changes

- **`shortcuts/doc/markdown_fix.go`** — add `WarnCalloutType(md, w io.Writer)`. Fence-aware line walker that scans `<callout …>` opening tags; for each tag with a recognized `type=` but no `background-color=`, writes one hint line to `w`. Read-only; markdown is never modified.
- **`shortcuts/doc/docs_create.go` / `shortcuts/doc/docs_update.go`** — invoke `WarnCalloutType(runtime.Str("markdown"), runtime.IO().ErrOut)` from the v1 Execute path, immediately before the API call. v2 and `--dry-run` paths are intentionally not wired (out of scope; can be revisited if/when the same pattern proves useful there).
- **`shortcuts/doc/markdown_fix_test.go`** — 14 `TestWarnCalloutType` subtests covering emits-hint, suppression, fence skipping, attribute-prefix isolation, and regressions for false positives.

## Robustness (review fixes from @fangshuyu-768)

- **Regex anchors `(?:^|\s)type=` and `(?:^|\s)background-color\s*=`** — replace `\b`. `\b` sits at any word/non-word boundary, and `-` is a non-word character, so `\btype=` also matches the suffix of `data-type=`. Result before fix: `<callout data-type="warning">` would emit a phantom `light-yellow` hint, and a future `data-background-color=` attribute would silently suppress the real hint. The new anchor requires a real attribute separator.
- **Code fences skipped.** Walks lines and tracks fence state via the existing `codeFenceOpenMarker` / `isCodeFenceClose` helpers (CommonMark §4.5, both ` ``` ` and `~~~`). A documentation sample inside a fenced block no longer produces a phantom hint.
- **`scanCalloutTagsForWarning` extracted** as a `FindAllStringSubmatch`-based per-line scanner. Replaces the prior `ReplaceAllStringFunc`-as-iterator pattern (rebuilt string was discarded; a second regex execution inside the callback recovered the capture groups). Same observable behavior, half the work.
- **Whitespace around `=` tolerated** — `background-color = "..."` is recognized as a present attribute, same as `background-color="..."`.

## type → suggested color (advisory mapping used in hint text)

| type | background-color | border-color |
|------|-----------------|-------------|
| `warning` | `light-yellow` | `yellow` |
| `info` / `note` | `light-blue` | `blue` |
| `tip` / `success` / `check` | `light-green` | `green` |
| `error` / `danger` | `light-red` | `red` |
| `caution` | `light-orange` | `orange` |
| `important` | `light-purple` | `purple` |

This mapping is the CLI's own convention and is used **only** to compose the hint string. The markdown is never modified, so callers and downstream MCP clients are not asked to know or replicate the mapping.

## Test Plan

- [x] `go test ./shortcuts/doc/...` — 14 `TestWarnCalloutType` cases pass (incl. 5 review-driven regressions: `data-type` not triggering, `data-background-color` not suppressing, backtick fence skip, tilde fence skip, fence-state reset)
- [x] `go test -race ./shortcuts/doc/...` — clean
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `golangci-lint run --new-from-rev=origin/main` — 0 issues
- [x] Skill format check passes
- [x] Full unit suite (`go test $(go list ./... | grep -v cli_e2e) -count=1`) — all packages green
- [x] Docs e2e (`go test ./tests/cli_e2e/docs/...`) — passes

## Out of scope / follow-up

- Native `type=` support inside `create-doc` would obsolete this hint entirely; that is option 1 of @SunPeiYang996's review and the right long-term fix.
- Wiring `WarnCalloutType` into the v2 path or into `--dry-run` is a small follow-up if the diagnostic proves useful.

### DIFF
--- a/shortcuts/doc/markdown_fix.go
+++ b/shortcuts/doc/markdown_fix.go
@@ -4,6 +4,8 @@
 package doc
 
 import (
+	"fmt"
+	"io"
 	"regexp"
 	"strings"
 	"unicode"
@@ -37,14 +39,6 @@ import (
 //  6. fixCalloutEmoji: replaces named emoji aliases (e.g. emoji="warning") with
 //     actual Unicode emoji characters that create-doc understands. Applied only
 //     outside fenced code blocks.
-//
-// prepareMarkdownForCreate applies fixes that should run on any Markdown before
-// it is sent to the create-doc / update-doc MCP tools, regardless of whether
-// the content originated from a Lark export or was written by hand.
-func prepareMarkdownForCreate(md string) string {
-	return applyOutsideCodeFences(md, fixCalloutType)
-}
-
 func fixExportedMarkdown(md string) string {
 	md = applyOutsideCodeFences(md, fixBoldSpacing)
 	md = applyOutsideCodeFences(md, normalizeNestedListIndentation)
@@ -52,7 +46,6 @@ func fixExportedMarkdown(md string) string {
 	md = applyOutsideCodeFences(md, fixBlockquoteHardBreaks)
 	md = fixTopLevelSoftbreaks(md)
 	md = applyOutsideCodeFences(md, fixCalloutEmoji)
-	md = applyOutsideCodeFences(md, fixCalloutType)
 	// Collapse runs of 3+ consecutive newlines into exactly 2 (one blank line),
 	// but only outside fenced code blocks to preserve intentional blank lines in code.
 	md = applyOutsideCodeFences(md, func(s string) string {
@@ -315,8 +308,9 @@ func fixSetextAmbiguity(md string) string {
 	return setextRe.ReplaceAllString(md, "$1\n\n$2")
 }
 
-// calloutTypeColors maps callout type="<name>" to a [background-color, border-color] pair.
-// When a callout tag has type= but no background-color=, these defaults are applied.
+// calloutTypeColors maps the semantic type= shorthand to a recommended
+// [background-color, border-color] pair for Feishu callout blocks.
+// Used only for hint messages — the Markdown itself is never rewritten.
 var calloutTypeColors = map[string][2]string{
 	"warning":   {"light-yellow", "yellow"},
 	"caution":   {"light-orange", "orange"},
@@ -330,42 +324,47 @@ var calloutTypeColors = map[string][2]string{
 	"important": {"light-purple", "purple"},
 }
 
-// calloutTypeRe matches a <callout …> opening tag so individual attributes can
-// be inspected and patched.
-var calloutTypeRe = regexp.MustCompile(`<callout(\s[^>]*)?>`)
+// calloutOpenTagRe matches a <callout …> opening tag.
+var calloutOpenTagRe = regexp.MustCompile(`<callout(\s[^>]*)?>`)
 
-// fixCalloutType expands the semantic type="<name>" shorthand on callout tags
-// into explicit background-color= (and border-color= when absent).  When a
-// background-color is already present the tag is left unchanged so that an
-// explicit color always wins.
-func fixCalloutType(md string) string {
-	return calloutTypeRe.ReplaceAllStringFunc(md, func(tag string) string {
+// calloutTypeAttrRe extracts the value of a type= attribute (single or double
+// quoted) from a callout opening tag's attribute string.
+var calloutTypeAttrRe = regexp.MustCompile(`\btype=(?:"([^"]*)"|'([^']*)')`)
+
+// WarnCalloutType scans md for callout tags that carry a type= attribute but
+// no background-color= attribute, then writes a hint line to w for each one
+// suggesting the explicit Feishu color attributes to use instead.
+//
+// The Markdown is not modified — the caller is responsible for acting on the
+// hints or ignoring them. This keeps the create/update path transparent: user
+// input reaches create-doc exactly as written.
+func WarnCalloutType(md string, w io.Writer) {
+	calloutOpenTagRe.ReplaceAllStringFunc(md, func(tag string) string {
 		attrs := ""
-		if m := calloutTypeRe.FindStringSubmatch(tag); len(m) == 2 {
+		if m := calloutOpenTagRe.FindStringSubmatch(tag); len(m) == 2 {
 			attrs = m[1]
 		}
-		// Extract type value.
-		typeRe := regexp.MustCompile(`\btype="([^"]*)"`)
-		typeParts := typeRe.FindStringSubmatch(attrs)
-		if len(typeParts) != 2 {
-			return tag // no type= attribute
-		}
-		typeName := typeParts[1]
-		colors, ok := calloutTypeColors[typeName]
-		if !ok {
-			return tag // unknown type — leave as-is
-		}
-		// Only inject background-color when it is absent.
+		// Skip tags that already carry an explicit background-color.
 		if strings.Contains(attrs, "background-color=") {
 			return tag
 		}
-		// Inject background-color (and border-color when absent) before the
-		// closing >.
-		extra := ` background-color="` + colors[0] + `"`
-		if !strings.Contains(attrs, "border-color=") {
-			extra += ` border-color="` + colors[1] + `"`
+		parts := calloutTypeAttrRe.FindStringSubmatch(attrs)
+		if len(parts) < 2 {
+			return tag // no type= attribute
 		}
-		return tag[:len(tag)-1] + extra + ">"
+		// parts[1] is the double-quoted capture, parts[2] is single-quoted.
+		typeName := parts[1]
+		if typeName == "" {
+			typeName = parts[2]
+		}
+		colors, ok := calloutTypeColors[typeName]
+		if !ok {
+			return tag // unknown type — no hint to give
+		}
+		fmt.Fprintf(w,
+			"hint: callout type=%q has no background-color; consider: background-color=%q border-color=%q\n",
+			typeName, colors[0], colors[1])
+		return tag
 	})
 }
 
--- a/shortcuts/doc/markdown_fix.go
+++ b/shortcuts/doc/markdown_fix.go
@@ -331,6 +331,11 @@ var calloutOpenTagRe = regexp.MustCompile(`<callout(\s[^>]*)?>`)
 // quoted) from a callout opening tag's attribute string.
 var calloutTypeAttrRe = regexp.MustCompile(`\btype=(?:"([^"]*)"|'([^']*)')`)
 
+// calloutBackgroundColorAttrRe matches a background-color= attribute name
+// with optional whitespace around the equals sign, so forms like
+// `background-color="..."` and `background-color = "..."` are both accepted.
+var calloutBackgroundColorAttrRe = regexp.MustCompile(`\bbackground-color\s*=`)
+
 // WarnCalloutType scans md for callout tags that carry a type= attribute but
 // no background-color= attribute, then writes a hint line to w for each one
 // suggesting the explicit Feishu color attributes to use instead.
@@ -345,7 +350,7 @@ func WarnCalloutType(md string, w io.Writer) {
 			attrs = m[1]
 		}
 		// Skip tags that already carry an explicit background-color.
-		if strings.Contains(attrs, "background-color=") {
+		if calloutBackgroundColorAttrRe.MatchString(attrs) {
 			return tag
 		}
 		parts := calloutTypeAttrRe.FindStringSubmatch(attrs)
--- a/shortcuts/doc/markdown_fix.go
+++ b/shortcuts/doc/markdown_fix.go
@@ -327,35 +327,78 @@ var calloutTypeColors = map[string][2]string{
 // calloutOpenTagRe matches a <callout …> opening tag.
 var calloutOpenTagRe = regexp.MustCompile(`<callout(\s[^>]*)?>`)
 
-// calloutTypeAttrRe extracts the value of a type= attribute (single or double
-// quoted) from a callout opening tag's attribute string.
-var calloutTypeAttrRe = regexp.MustCompile(`\btype=(?:"([^"]*)"|'([^']*)')`)
+// calloutTypeAttrRe extracts the value of a type= attribute (single or
+// double quoted) from a callout opening tag's attribute string. The
+// (?:^|\s) anchor instead of \b is intentional: \b sits at any
+// word/non-word boundary, and `-` is a non-word character, so
+// `\btype=` would also match the suffix of `data-type=` and yield a
+// bogus type lookup. Anchoring on start-of-string-or-whitespace
+// requires a real attribute separator before the name.
+var calloutTypeAttrRe = regexp.MustCompile(`(?:^|\s)type=(?:"([^"]*)"|'([^']*)')`)
 
-// calloutBackgroundColorAttrRe matches a background-color= attribute name
-// with optional whitespace around the equals sign, so forms like
-// `background-color="..."` and `background-color = "..."` are both accepted.
-var calloutBackgroundColorAttrRe = regexp.MustCompile(`\bbackground-color\s*=`)
+// calloutBackgroundColorAttrRe matches a background-color= attribute
+// name with optional whitespace around the equals sign, so forms like
+// `background-color="..."` and `background-color = "..."` are both
+// accepted. Same (?:^|\s) anchor as calloutTypeAttrRe, for the same
+// reason: `data-background-color="..."` must not look like a present
+// background-color and silently suppress the hint.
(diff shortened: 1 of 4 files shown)


SunPeiYang996:
Thanks for the PR and the thorough test coverage!
However, I think this change puts protocol-adaptation logic in the wrong layer. The type → color mapping is essentially compensating for a gap in the create-doc API itself. If we add this to the shortcut, every other client that calls create-doc will need to duplicate the same mapping.
A few specific concerns:

The mapping (e.g. warning → light-yellow) is not defined by any Feishu spec — it's a convention we'd be inventing and then silently applying.
It transforms user input without any visibility, making debugging harder.
It expands the shortcut's responsibility from "invoke MCP" to "fix MCP's input format."

Suggested alternatives:

File a feature request for create-doc to natively support type= on callout blocks.
If we need a short-term workaround, make it an explicit opt-in flag (e.g. --fix-callout-types) with a deprecation plan.
Or, surface a clear warning when type= is used without background-color, telling the user what to write instead.

Happy to discuss further!

herbertliu:
Thanks for the detailed review and the thoughtful suggestions!

You're right — silently rewriting user Markdown is the wrong approach here. The latest commit (011da3d) replaces `fixCalloutType()` with `WarnCalloutType()`, which:
- **Leaves the Markdown unchanged** — user input reaches `create-doc` exactly as written
- **Writes a hint to stderr** for each `type=` callout missing `background-color`, e.g.:
  ```
  hint: callout type="warning" has no background-color; consider: background-color="light-yellow" border-color="yellow"
  ```

This surfaces the issue visibly without any silent transformation. On the longer term, filing a feature request for `create-doc` to natively support `type=` makes sense — happy to do that as a follow-up.

shortcuts/doc/markdown_fix.go
```diff
@@ -220,6 +229,60 @@ func fixSetextAmbiguity(md string) string {
...
+	return calloutTypeRe.ReplaceAllStringFunc(md, func(tag string) string {
+		attrs := ""
+		if m := calloutTypeRe.FindStringSubmatch(tag); len(m) == 2 {
+			attrs = m[1]
+		}
+		// Extract type value.
+		typeRe := regexp.MustCompile(`\btype="([^"]*)"`)
+		typeParts := typeRe.FindStringSubmatch(attrs)
+		if len(typeParts) != 2 {
+			return tag // no type= attribute
```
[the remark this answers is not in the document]
herbertliu: Fixed in 011da3d. The regex was updated to `\btype=(?:"([^"]*)'|'([^']*)')` which captures both double-quoted and single-quoted `type=` values. A dedicated test case (`single-quoted type attribute emits hint`) was added to `TestWarnCalloutType` to cover this.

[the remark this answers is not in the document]
herbertliu: The broader fence-detection limitation (only bare triple-backtick fences) is a pre-existing accepted constraint of `applyOutsideCodeFences`, since `fetch-doc` exports exclusively in that form. Now that `WarnCalloutType` is a read-only scanner rather than a Markdown mutator, the worst-case impact of a false-negative inside a `~~~` fence is an unnecessary hint line — no correctness issue. Leaving as a known limitation to keep this change focused.

shortcuts/doc/markdown_fix.go
```diff
@@ -220,6 +222,66 @@ func fixSetextAmbiguity(md string) string {
...
+// input reaches create-doc exactly as written.
+func WarnCalloutType(md string, w io.Writer) {
+	calloutOpenTagRe.ReplaceAllStringFunc(md, func(tag string) string {
+		attrs := ""
+		if m := calloutOpenTagRe.FindStringSubmatch(tag); len(m) == 2 {
+			attrs = m[1]
+		}
+		// Skip tags that already carry an explicit background-color.
+		if strings.Contains(attrs, "background-color=") {
+			return tag
```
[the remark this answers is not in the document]
herbertliu: Fixed in b4e9d47. Replaced the `strings.Contains(attrs, "background-color=")` check with a `\bbackground-color\s*=` regex so forms like `background-color = "light-red"` (with whitespace around the equals sign) are also recognized. Added a regression test case.

shortcuts/doc/markdown_fix.go
```diff
@@ -220,6 +222,71 @@ func fixSetextAmbiguity(md string) string {
...
+
+// WarnCalloutType scans md for callout tags that carry a type= attribute but
+// no background-color= attribute, then writes a hint line to w for each one
+// suggesting the explicit Feishu color attributes to use instead.
+//
+// The Markdown is not modified — the caller is responsible for acting on the
+// hints or ignoring them. This keeps the create/update path transparent: user
+// input reaches create-doc exactly as written.
+func WarnCalloutType(md string, w io.Writer) {
+	calloutOpenTagRe.ReplaceAllStringFunc(md, func(tag string) string {
```
fangshuyu-768: **`ReplaceAllStringFunc` is being used as an iterator, and its result is discarded.**

The callback always `return tag`, and the new string that `ReplaceAllStringFunc` builds is thrown away. Inside the callback the same regex is also executed a second time (`calloutOpenTagRe.FindStringSubmatch(tag)`) just to pull the attribute-string capture group.

`FindAllStringSubmatch` gives you both iteration and the capture groups in one pass, with no throwaway string allocation:

```go
for _, m := range calloutOpenTagRe.FindAllStringSubmatch(md, -1) {
    attrs := m[1]
    if calloutBackgroundColorAttrRe.MatchString(attrs) {
        continue
    }
    parts := calloutTypeAttrRe.FindStringSubmatch(attrs)
    if len(parts) < 2 {
        continue
    }
    typeName := parts[1]
    if typeName == "" {
        typeName = parts[2]
    }
    colors, ok := calloutTypeColors[typeName]
    if !ok {
        continue
    }
    fmt.Fprintf(w, "hint: …\n", typeName, colors[0], colors[1])
}
```

Same observable behaviour, clearer intent, no redundant regex execution.

shortcuts/doc/markdown_fix.go
```diff
@@ -220,6 +222,71 @@ func fixSetextAmbiguity(md string) string {
...
+var calloutBackgroundColorAttrRe = regexp.MustCompile(`\bbackground-color\s*=`)
+
+// WarnCalloutType scans md for callout tags that carry a type= attribute but
+// no background-color= attribute, then writes a hint line to w for each one
+// suggesting the explicit Feishu color attributes to use instead.
+//
+// The Markdown is not modified — the caller is responsible for acting on the
+// hints or ignoring them. This keeps the create/update path transparent: user
+// input reaches create-doc exactly as written.
+func WarnCalloutType(md string, w io.Writer) {
```
fangshuyu-768, line 360: **`WarnCalloutType` isn't wrapped in `applyOutsideCodeFences`, so it hints on callout tags inside code samples too.**

Other transforms in this file (`fixBoldSpacing`, `fixSetextAmbiguity`, `fixBlockquoteHardBreaks`, `fixCalloutEmoji`) all go through `applyOutsideCodeFences` to avoid touching literal code content. If a user writes documentation/tutorials containing:

````markdown
```markdown
<callout type="warning">example</callout>
```
````

…they'll get hint lines on stderr for every code-sample callout, which is noisy and confusing.

Two options:
- Wrap the scan in `applyOutsideCodeFences` (or a similar helper that iterates over non-code regions) so only real callouts trigger hints.
- If "warn on every occurrence, including code blocks" is the intended design, make that explicit in the function doc so future readers don't assume it's a bug.

File: shortcuts/doc/markdown_fix.go [hunk 5]
```diff
@@ -220,6 +222,71 @@ func fixSetextAmbiguity(md string) string {
...
+	"danger":    {"light-red", "red"},
+	"important": {"light-purple", "purple"},
+}
+
+// calloutOpenTagRe matches a <callout …> opening tag.
+var calloutOpenTagRe = regexp.MustCompile(`<callout(\s[^>]*)?>`)
+
+// calloutTypeAttrRe extracts the value of a type= attribute (single or double
+// quoted) from a callout opening tag's attribute string.
+var calloutTypeAttrRe = regexp.MustCompile(`\btype=(?:"([^"]*)"|'([^']*)')`)
```
fangshuyu-768: **`\btype=` matches `data-type="…"` because `-` is a non-word char.**

With `\b`, the boundary sits between the `-` (non-word) and the `t` in `-type` (word), so a tag like

```html
<callout data-type="warning" background-color="light-red">
```

will be seen as carrying `type="warning"`. In this specific hypothetical it wouldn't produce a bad hint (background-color is present → suppressed), but it will misfire on variants like `<callout data-type="warning">` — the hint will wrongly recommend `light-yellow` because the user's real `type` attribute isn't actually `"warning"`.

Tightening the pattern to require an actual attribute separator removes the ambiguity:

```go
var calloutTypeAttrRe = regexp.MustCompile(`(?:^|\s)type=(?:"([^"]*)"|'([^']*)')`)
```

The same concern applies to `calloutBackgroundColorAttrRe` at L251 — `\bbackground-color\s*=` would match `data-background-color=` for the same reason. Swapping to `(?:^|\s)background-color\s*=` keeps the whitespace tolerance you added in the last commit while avoiding the false match.

Not a likely case for real `<callout>` tags, but cheap to lock down.

shortcuts/doc/markdown_fix.go
```diff
@@ -220,6 +222,71 @@ func fixSetextAmbiguity(md string) string {
...
+// `background-color="..."` and `background-color = "..."` are both accepted.
+var calloutBackgroundColorAttrRe = regexp.MustCompile(`\bbackground-color\s*=`)
+
+// WarnCalloutType scans md for callout tags that carry a type= attribute but
+// no background-color= attribute, then writes a hint line to w for each one
+// suggesting the explicit Feishu color attributes to use instead.
+//
+// The Markdown is not modified — the caller is responsible for acting on the
+// hints or ignoring them. This keeps the create/update path transparent: user
+// input reaches create-doc exactly as written.
```
fangshuyu-768: **PR description is out of date relative to this implementation.**

The Summary and Changes sections still describe `fixCalloutType()` / `prepareMarkdownForCreate()` rewriting markdown into explicit `background-color` / `border-color` pairs, and the "type → color mapping" table is presented as what the code does. The current implementation (after the `refactor: replace silent callout type→color injection with hint output` commit) is hint-only — no markdown is rewritten, the mapping table is used only to phrase the suggestion on stderr.

Readers opening the PR to understand scope will be misled. Worth updating the body to match what's actually in the head commit: the checker surfaces hints, the transform was intentionally dropped on reviewer feedback (linking to the discussion is even better).

(Leaving this as an inline anchored to `WarnCalloutType` since it's the public surface the PR description should be describing; feel free to treat as a general-PR comment.)

fangshuyu-768 (commented):
Direction looks right — the switch from rewriting markdown to hint-only is a healthier boundary. A few notes on the current implementation, plus a PR-description update request.

shortcuts/doc/markdown_fix.go
herbertliu: Fixed in 7d4b556. Extracted `scanCalloutTagsForWarning()` which uses `FindAllStringIndex` + a range loop. No more result-discarded iteration; the intent (walk matches, report via writer) is now self-evident.

shortcuts/doc/markdown_fix.go
herbertliu: Fixed in 7d4b556. `WarnCalloutType` now walks lines directly with `codeFenceOpenMarker` / `isCodeFenceClose` (shared helpers from the docs +update check pipeline), which recognizes both backtick AND tilde fences. Regression tests added for both fence flavors plus a mixed prose+fence case.

shortcuts/doc/markdown_fix.go
herbertliu: Fixed in 7d4b556. Regex changed from `\btype=...` to `(?:^|\s)type=...` so the attribute is only matched after start-of-string or whitespace. A `data-type="warning"` no longer matches. Regression test `data-type attribute does not trigger hint` added.

shortcuts/doc/markdown_fix.go
herbertliu: Updated the PR description to reflect the current implementation: the "silently rewriting" paragraph is gone; it now describes `WarnCalloutType` as a stderr-only hint, and adds a dedicated "Robustness fixes" section documenting the four review-driven improvements (regex boundary, FindAllStringIndex extraction, code-fence skipping, whitespace-around-equals tolerance).

shortcuts/doc/markdown_fix.go
fangshuyu-768: Fixed in b33d5c2.

Pulled out a `scanCalloutTagsForWarning(line, w)` helper that uses `FindAllStringSubmatch` — one pass for both iteration and capture groups, no rebuilt-string allocation thrown away, no second regex execution inside the callback.

The previous reply on this thread pointed to `7d4b556`, but that commit got dropped during a rebase/squash and never landed on the branch — only `da02e8f`/`5728d5e`/`1950398` were on the head before this fix. b33d5c2 makes the refactor real.

shortcuts/doc/markdown_fix.go
fangshuyu-768: Fixed in b33d5c2.

`WarnCalloutType` now walks lines and tracks fence state via the existing `codeFenceOpenMarker` / `isCodeFenceClose` helpers from `docs_update_check.go`, so a `<callout type="warning">` sample inside a ` ``` `/`~~~` fence (CommonMark §4.5) no longer produces a phantom stderr hint.

Three new regression cases pin the contract:
- `callout inside backtick fence emits no hint`
- `callout inside tilde fence emits no hint`
- `callout after fence close still emits hint` (proves the fence state is reset, not stuck)

Same caveat as the sibling threads: the earlier reply pointed to `7d4b556`, which was lost in a rebase/squash; b33d5c2 makes the fence-skipping real.

shortcuts/doc/markdown_fix.go
fangshuyu-768: Fixed in b33d5c2.

Both `calloutTypeAttrRe` and `calloutBackgroundColorAttrRe` switched from `\b` to `(?:^|\s)` so the regex requires a real attribute separator before the name (start-of-string or whitespace), instead of any word/non-word boundary that `\b` accepts.

Two new regression cases pin both directions:
- `data-type attribute does not trigger hint` (the reported case — `<callout data-type="warning">` would otherwise emit a bogus light-yellow hint)
- `data-background-color does not suppress hint` (the symmetric case you flagged in the same comment — a future `data-background-color=` attribute must not look like a present `background-color` and silently suppress the real hint)

Earlier reply pointed to `7d4b556`, which was lost in a rebase/squash; b33d5c2 makes the regex tightening real.

shortcuts/doc/markdown_fix.go
fangshuyu-768: Resolved — the PR body now describes the hint-only design (no markdown rewriting) with a dedicated "Robustness fixes" section listing the four review-driven improvements. Confirmed by re-reading the current body.

fangshuyu-768 (approved):

PR #469, opened 2026-04-14
fix(doc): preserve round-trip formatting in fetch output

GaoSSR:
## Summary
Fixes a docs round-trip formatting regression in `docs +fetch` output. Some exported markdown could not be safely re-imported with `docs +create` / `docs +update`, which could flatten nested lists and surface inline formatting as literal markdown or text tags.

## Changes
- Trim leading spaces after opening bold and italic delimiters in fetched markdown
- Normalize space-pair-indented nested list markers to tab indentation before returning fetched markdown
- Add regression tests for emphasis spacing, text-tag emphasis, and nested list indentation

## Test Plan
- [x] `go test ./shortcuts/doc`
- [x] `go test ./shortcuts/doc -run 'TestFixExportedMarkdown|TestFixTopLevelSoftbreaks|TestFixBoldSpacing|TestNormalizeNestedListIndentation'`
- [x] `make build`
- [x] Manual local verification: `./lark-cli docs +fetch --doc 'KHCydlulPoZosix0m0ccKPWnnsg' --as user` now emits `**SFT ...**` instead of `** SFT ...**`, and nested ordered list items are tab-indented instead of ambiguous space-pair indentation
- [ ] `make unit-test` still fails locally in unchanged `shortcuts/minutes` download tests with `blocked download URL: local/internal host is not allowed`

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * More consistent nested list rendering by normalizing space-pair indentation into tab-style nesting.

* **Bug Fixes**
  * Improved emphasis spacing: trims stray internal spaces in bold/italic while preserving ambiguous patterns, multi-asterisk cases, and inline code spans.
  * Emphasis normalization is more robust with Unicode-aware trimming and avoids altering code blocks.

* **Tests**
  * Added/expanded tests covering nested lists, varied emphasis spacing, and inline-code interactions.

### DIFF
--- a/shortcuts/doc/markdown_fix.go
+++ b/shortcuts/doc/markdown_fix.go
@@ -123,10 +123,10 @@ func fixBlockquoteHardBreaks(md string) string {
 //
 // Both fixes skip inline code spans to avoid modifying literal code content.
 var (
-	boldLeadingSpaceRe    = regexp.MustCompile(`(\*\*)\s+([^*\n].*?)(\*\*)`)
+	boldLeadingSpaceRe    = regexp.MustCompile(`(\*\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*\*)`)
 	boldTrailingSpaceRe   = regexp.MustCompile(`(\*\*\S[^*]*?)\s+(\*\*)`)
-	italicLeadingSpaceRe  = regexp.MustCompile(`(\*)\s+([^*\n].*?)(\*)`)
-	italicTrailingSpaceRe = regexp.MustCompile(`(\*\S[^*]*?)\s+(\*)`)
+	italicLeadingSpaceRe  = regexp.MustCompile(`(^|[^*])(\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*)([^*]|$)`)
+	italicTrailingSpaceRe = regexp.MustCompile(`(^|[^*])(\*\S[^*]*?)\s+(\*)([^*]|$)`)
 	// headingBoldRe uses [^*]+ (no asterisks) to avoid mismatching headings
 	// that contain multiple disjoint bold spans such as "# **foo** and **bar**".
 	headingBoldRe = regexp.MustCompile(`(?m)^(#{1,6})\s+\*\*([^*]+)\*\*\s*$`)
@@ -199,8 +199,8 @@ func fixBoldSpacingLine(line string) string {
 	if len(spans) == 0 {
 		line = boldLeadingSpaceRe.ReplaceAllString(line, "$1$2$3")
 		line = boldTrailingSpaceRe.ReplaceAllString(line, "$1$2")
-		line = italicLeadingSpaceRe.ReplaceAllString(line, "$1$2$3")
-		line = italicTrailingSpaceRe.ReplaceAllString(line, "$1$2")
+		line = italicLeadingSpaceRe.ReplaceAllString(line, "$1$2$3$4$5")
+		line = italicTrailingSpaceRe.ReplaceAllString(line, "$1$2$3$4")
 		return line
 	}
 	var sb strings.Builder
@@ -210,8 +210,8 @@ func fixBoldSpacingLine(line string) string {
 		seg := line[pos:loc[0]]
 		seg = boldLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3")
 		seg = boldTrailingSpaceRe.ReplaceAllString(seg, "$1$2")
-		seg = italicLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3")
-		seg = italicTrailingSpaceRe.ReplaceAllString(seg, "$1$2")
+		seg = italicLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3$4$5")
+		seg = italicTrailingSpaceRe.ReplaceAllString(seg, "$1$2$3$4")
 		sb.WriteString(seg)
 		// Preserve inline code span as-is.
 		sb.WriteString(line[loc[0]:loc[1]])
@@ -221,8 +221,8 @@ func fixBoldSpacingLine(line string) string {
 	seg := line[pos:]
 	seg = boldLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3")
 	seg = boldTrailingSpaceRe.ReplaceAllString(seg, "$1$2")
-	seg = italicLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3")
-	seg = italicTrailingSpaceRe.ReplaceAllString(seg, "$1$2")
+	seg = italicLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3$4$5")
+	seg = italicTrailingSpaceRe.ReplaceAllString(seg, "$1$2$3$4")
 	sb.WriteString(seg)
 	return sb.String()
 }
@@ -318,6 +318,9 @@ func normalizeNestedListIndentation(md string) string {
 		if len(matches) != 3 {
 			continue
 		}
+		if !hasPreviousNonBlankListItem(lines, i) {
+			continue
+		}
 		indent := matches[1]
 		if len(indent)%2 != 0 {
 			continue
@@ -326,6 +329,17 @@ func normalizeNestedListIndentation(md string) string {
 		lines[i] = tabs + line[len(indent):]
 	}
 	return strings.Join(lines, "\n")
+}
+
+func hasPreviousNonBlankListItem(lines []string, index int) bool {
+	for i := index - 1; i >= 0; i-- {
+		trimmed := strings.TrimSpace(lines[i])
+		if trimmed == "" {
+			continue
+		}
+		return listItemRe.MatchString(lines[i])
+	}
+	return false
 }
 
 // isListItemOrContinuation returns true for lines that are part of a list:
--- a/shortcuts/doc/markdown_fix.go
+++ b/shortcuts/doc/markdown_fix.go
@@ -15,24 +15,29 @@ import (
 //     and strips redundant ** from ATX headings. Applied only outside fenced
 //     code blocks, and skips inline code spans.
 //
-//  2. fixSetextAmbiguity: inserts a blank line before any "---" that immediately
+//  2. normalizeNestedListIndentation: rewrites space-pair-indented nested list
+//     markers to tab-indented markers. This avoids nested ordered list items
+//     being flattened or interpreted as plain text/code on re-import.
+//
+//  3. fixSetextAmbiguity: inserts a blank line before any "---" that immediately
 //     follows a non-empty line, preventing it from being parsed as a Setext H2.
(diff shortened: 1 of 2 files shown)


shortcuts/doc/markdown_fix.go
```diff
@@ -291,6 +304,44 @@ var contentContainers = [][2]string{
...
+var nestedListIndentRe = regexp.MustCompile(`^( {2,})([-*+]|\d+[.)]) `)
+
+func normalizeNestedListIndentation(md string) string {
+	lines := strings.Split(md, "\n")
+	for i, line := range lines {
+		matches := nestedListIndentRe.FindStringSubmatch(line)
+		if len(matches) != 3 {
+			continue
+		}
+		if !hasPreviousNonBlankListItem(lines, i) {
```
fangshuyu-768, line 394: This heuristic skips blank lines when searching upward, so a loose-list sibling like `1. a\n\n  1. b` would still see the earlier item and be rewritten as a nested list. That changes the Markdown structure. Could we stop at the first blank separator, or otherwise require the candidate line to still belong to the same list block before converting it?

shortcuts/doc/markdown_fix.go
```diff
@@ -291,6 +304,44 @@ var contentContainers = [][2]string{
...
+// spaces. We rewrite those space pairs to tabs because some downstream
+// round-trip paths treat multi-space indented ordered items as flat items or
+// literal text, while tab indentation remains nested and avoids 4-space code
+// block ambiguity.
+var nestedListIndentRe = regexp.MustCompile(`^( {2,})([-*+]|\d+[.)]) `)
+
+func normalizeNestedListIndentation(md string) string {
+	lines := strings.Split(md, "\n")
+	for i, line := range lines {
+		matches := nestedListIndentRe.FindStringSubmatch(line)
```
fangshuyu-768, line 390: A 4-space-indented marker inside a list item can also be an indented code block, not a nested list. Rewriting every `nestedListIndentRe` match to tabs would change the rendered semantics for cases like `- parent\n\n    1. code`. Can we guard against code-block contexts before normalizing?

shortcuts/doc/markdown_fix_test.go
```diff
@@ -222,6 +247,43 @@ func TestFixTopLevelSoftbreaks(t *testing.T) {
 	}
 }
 
+func TestNormalizeNestedListIndentation(t *testing.T) {
```
fangshuyu-768, line 285: Could we add regression coverage for the two edge cases above as part of this test table: (1) a loose-list sibling separated by a blank line, and (2) an indented code block inside a list item that starts with something resembling a list marker? Those would make the intended boundary of this normalization much clearer.

fangshuyu-768 (commented):
Thanks for the formatting fix. I left a few edge-case concerns around nested list normalization and the missing regression coverage.

shortcuts/doc/markdown_fix.go
GaoSSR: Good point. I changed the parent-list lookup to stop at the first blank line, so 1. a\n\n  1. b now stays as a loose-list sibling instead of being rewritten as nested content.

shortcuts/doc/markdown_fix.go
GaoSSR: Handled this through the same boundary check. If there is a blank separator before the indented marker, it is no longer treated as nested list context, so - parent\n\n    1. code stays as an indented code block.

shortcuts/doc/markdown_fix_test.go
GaoSSR: Added both regression cases to TestNormalizeNestedListIndentation: the blank-line-separated loose-list sibling and the indented code block case.

shortcuts/doc/markdown_fix.go
```diff
@@ -106,20 +111,25 @@ func fixBlockquoteHardBreaks(md string) string {
...
+//     "# **text**" → "# text"
 //
-// Both fixes skip inline code spans to avoid modifying literal code content.
+// The bold and italic spacing fixes only run on non-code segments so literal
+// code content is left unchanged.
 var (
+	boldLeadingSpaceRe    = regexp.MustCompile(`(\*\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*\*)`)
 	boldTrailingSpaceRe   = regexp.MustCompile(`(\*\*\S[^*]*?)\s+(\*\*)`)
-	italicTrailingSpaceRe = regexp.MustCompile(`(\*\S[^*]*?)\s+(\*)`)
+	italicLeadingSpaceRe  = regexp.MustCompile(`(^|[^*])(\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*)([^*]|$)`)
```
fangshuyu-768: **Issue 1 — `** hello **` (space on *both* sides) is a fixed point and stays unfixed.**

`boldLeadingSpaceRe` requires the content to end in `[^*\s\n]`, so any trailing space before the closing `**` makes it fail. Symmetrically, `boldTrailingSpaceRe` requires `\*\*\S` — non-space immediately after the opening `**` — so a leading space makes it fail. Neither regex fires, and the output is untouched. Same story for `* hello *` via `italicLeadingSpaceRe` / `italicTrailingSpaceRe`.

This is likely a common real-world export shape (e.g. `** Warning **`, `* Note *`). Trace:

- `** hello **`: leading regex content = `hello` (backtracked to drop trailing space), but the next char is ` `, not `*` → no match. Trailing regex: `\*\*\S` fails because the char after `**` is ` `. No change.

Suggested fix: add a combined regex that handles both ends at once, e.g. `(\*\*)\s+([^*\n]+?)\s+(\*\*)` (mirror for italic), applied before the one-sided ones. Alternatively run the bold pass twice — but the single-pass pair alone cannot converge on this input, so a loop without a new regex won't help either.

Also please add the symmetric test cases (`** hello **` → `**hello**`, `* hello *` → `*hello*`) so this gap can't silently re-appear.

shortcuts/doc/markdown_fix.go
```diff
@@ -190,26 +200,32 @@ func fixBoldSpacingLine(line string) string {
 	}
 	spans := scanInlineCodeSpans(line)
 	if len(spans) == 0 {
+		line = boldLeadingSpaceRe.ReplaceAllString(line, "$1$2$3")
 		line = boldTrailingSpaceRe.ReplaceAllString(line, "$1$2")
-		line = italicTrailingSpaceRe.ReplaceAllString(line, "$1$2")
+		line = italicLeadingSpaceRe.ReplaceAllString(line, "$1$2$3$4$5")
```
fangshuyu-768: **Issue 2 — consecutive italic spans on one line: only the first is normalized.**

`italicLeadingSpaceRe` uses `(^|[^*])(\*)…(\*)([^*]|$)` to emulate look-behind/look-ahead via captured boundary chars. Because Go's `ReplaceAllString` advances non-overlapping past the consumed match, the boundary char after the closing `*` is eaten and can't serve as the boundary before the *next* opening `*`. Concrete trace on `* a* * b*`:

1. Match at pos 0: consumes `* a* ` (including the trailing ` ` as `$5`). Replacement: `*a* `. Output so far: `*a* * b*`.
2. Engine resumes at pos 5 (`*`). `(^|[^*])` needs `^` (no, not at start anymore) or `[^*]` (no, pos 5 is `*`). Fails and advances.
3. At pos 6 (` `): matches `[^*]`, then `(\*)` wants pos 7 = `b` — fail.
4. No more matches. The second span `* b*` survives.

Same issue applies to `italicTrailingSpaceRe` for `*a *b *`-style inputs.

Suggested fix: loop until the pass is a fixed point, e.g.

```go
for {
    prev := seg
    seg = italicLeadingSpaceRe.ReplaceAllString(seg, "$1$2$3$4$5")
    seg = italicTrailingSpaceRe.ReplaceAllString(seg, "$1$2$3$4")
    if seg == prev { break }
}
```

And a test: `"* a* * b*"` → `"*a* *b*"`.

shortcuts/doc/markdown_fix.go
```diff
@@ -106,20 +111,25 @@ func fixBlockquoteHardBreaks(md string) string {
...
-//     Headings are already bold, so the inner ** is visually redundant and
-//     some renderers display the markers literally.
+//  3. Removes redundant bold around an entire ATX heading:
+//     "# **text**" → "# text"
 //
-// Both fixes skip inline code spans to avoid modifying literal code content.
+// The bold and italic spacing fixes only run on non-code segments so literal
+// code content is left unchanged.
 var (
+	boldLeadingSpaceRe    = regexp.MustCompile(`(\*\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*\*)`)
```
fangshuyu-768: **Issue 3 (nit) — `boldLeadingSpaceRe` lacks the `(^|[^*])` boundary guard that `italicLeadingSpaceRe` on line 131 has.**

For an input like `*** hello**` the bold regex matches starting at pos 1 (because `**` + space + content + `**` is satisfied there) and the output becomes `***hello**` — silently changing the emphasis nesting. The italic regex wouldn't do this because its `(^|[^*])` prefix forbids a leading `*`.

For symmetry and safety, consider changing to `(^|[^*])(\*\*)\s+([^*\n](?:[^*\n]*[^*\s\n])?)(\*\*)([^*]|$)` (and updating the replacement to include the boundary captures). Low-priority — stray `***` sequences in exports are rare — but the asymmetry stood out.

fangshuyu-768 (commented):
My earlier comments on the nested-list heuristic are all resolved — confirmed by code + new test cases. Taking a fresh look at the new emphasis (bold/italic) leading-space rules, I spotted three issues worth addressing before merge. All verified by tracing the regexes by hand against the code paths in `fixBoldSpacingLine`.

shortcuts/doc/markdown_fix.go
GaoSSR: Fixed. The old leading/trailing spacing regexes blocked each other when both inner sides had spaces, so `** hello **` and `* hello *` were left unchanged. I replaced that path with a narrower emphasis-spacing scan and added regression coverage for both cases.

shortcuts/doc/markdown_fix.go
GaoSSR: Fixed. The previous italic rewrite depended on captured boundary characters, so `ReplaceAllString` advanced past the boundary needed by the next span on the same line. The updated logic handles consecutive spans independently, and I added a regression test for `* a* * b*`.

shortcuts/doc/markdown_fix.go
GaoSSR: Confirmed and fixed. `*** hello**` could be rewritten incorrectly before. The new logic only normalizes simple `*...*` and `**...**` spans and ignores `***` runs, so it no longer changes that nesting-shaped near miss. Added a regression test for it as well.

fangshuyu-768 (approved):

GaoSSR:
@fangshuyu-768 Thanks for merge

PR #494, opened 2026-04-15
feat(sheets): add float image shortcuts

caojie0621:
## Summary                                                                                                                                                                                  
                                                                                                                                                                                              
  Add a complete set of shortcuts for floating images on spreadsheets, plus a                                                                                                                 
  dedicated media upload shortcut that correctly handles both the single-part                                                                                                                 
  and multipart upload paths for `parent_type=sheet_image`.                                                                                                                
                                                                                                                                                                                              
  ## New shortcuts                                                                                                                                                                            
                                                                                                                                                                                              
  | Shortcut | Purpose |                                                                                                                                                                      
  |---|---|                                                                                                                                                                                   
  | `sheets +media-upload` | Upload a local image as `parent_type=sheet_image`; auto-switches to multipart for `> 20 MB`. Returns `file_token` for use as `--float-image-token`. |            
  | `sheets +create-float-image` | Create a floating image anchored at a single cell. |
  | `sheets +update-float-image` | Update range / width / height / offset. At least one mutable field required. |
  | `sheets +get-float-image` | Fetch one float image's metadata. |
  | `sheets +list-float-images` | List all float images on a sheet. |                                                                                                                         
  | `sheets +delete-float-image` | Delete a float image by id. |                                                                                                                              
                                                                                                                                                                                              
  ## Client-side validation (create / update)                                                                                                                                                 
                                                                                                                                                                                              
  - `--range` must be a single anchor cell (`sheetId!A1:A1`); multi-cell spans rejected.                                                                                                      
  - `--range` prefix must equal `--sheet-id` (catches `--sheet-id sheet1 --range other!A1:A1` locally).                                                                                       
  - `--width` / `--height` must be `>= 20` pixels when set.                                                                                                                                   
  - `--offset-x` / `--offset-y` must be `>= 0` when set. Upper bound (< anchor cell's width/height) is enforced server-side and surfaced via a friendly hint (see below).                     
  - `+update-float-image` requires at least one of `--range` / `--width` / `--height` / `--offset-x` / `--offset-y`; ID-only invocations are rejected before a `PATCH {}` can reach the
  server.                                                                                                                                                                                     
                                                            
  ## Error classifier                                                                                                                                                                         
                                                                                                                                                                                              
  Add `LarkErrSheetsFloatImageInvalidDims = 1310246` to the centralized                                                                                                                       
  classifier. When the backend rejects width / height / offset values, the CLI                                                                                                                
  surfaces an `invalid_params` error with a hint pointing at the four flags.
                                                                                                                                                                                              
  ## Flag typing                          
                                                                                                                                                                                              
  Width / height / offset-x / offset-y flags are declared `Type: "int"` and                                                                                                                   
  `"width":"300"`. Aligns with other sheets shortcuts that take numeric flags.                                                                                                                
                                                                                                                                                                                              
  ## Scope least-privilege                                                                                                                                                                    
                                          
  Create / update / delete shortcuts declare only `sheets:spreadsheet:write_only`                                                                                                             
  (dropping the previously declared `sheets:spreadsheet:read`), matching the                                                                                                                  
  SKILL.md permission table.                                                                                                                                                                  
                                                                                                                                                                                              
  ## `cmd/api` form-field serialization bug fix                                                                                                                                               
                                               
  `internal/cmdutil/fileupload.go#BuildFormdata` previously rendered form field                                                                                                               
  values with `fmt.Sprintf("%v", v)`. JSON numbers unmarshal to `float64`, and                                                                                                                
  `%v` delegates to `%g` which switches to scientific notation at `~1e6`                                                                                                                      
  (e.g. `1185356` → `"1.185356e+06"`). Backends parsing the field as an integer                                                                                                               
  then reject it (seen as `1061002 params error` when uploading a 1.16 MB image                                                                                                               
  via `lark-cli api --file ... --data '{"size":1185356}'`).                    
                                                                                                                                                                                              
  Introduce `formatFormFieldValue` that uses                                                                                                                                                  
  `strconv.FormatFloat(n, 'f', -1, 64)` for `float64` and falls through to `%v`                                                                                                               
  for other types. Fixes all raw-API uploads carrying large numeric fields,                                                                                                                   
  not just the float-image case. Shortcut paths that use                                                                                                                                      
  `common.UploadDriveMedia{All,Multipart}` were already safe because those                                                                                                                    
  helpers format integers via `%d` explicitly.                                                                                                                                                
                                                                                                                                                                                              
  ## Test plan                                                                                                                                                                                
                                                                                                                                                                                              
  - [x] `go test ./shortcuts/sheets/ ./internal/cmdutil/ ./cmd/api/ ./cmd/service/ ./internal/output/`                                                                                        
  - [x] Create / update / delete / list / get on a live spreadsheet (21 scenarios incl. boundaries: `--width 20`, `--offset-x 0`, multi-cell range rejection, sheet-id prefix mismatch,       
  negative offsets, empty update payload, server-side `1310246` with hint, bot vs user read permissions, bot `docs +media-preview` access to the returned token).                             
  - [x] Unit: `TestFormatFormFieldValue` table-drives the scientific-notation boundary and pass-through types.                                                                                
  - [x] Unit: float-image validators (single-cell, sheet-id match, dim bounds, empty payload).                
  - [x] Unit: `normalizeBatchStyleRanges` was addressed on a sibling branch and is out of scope here.                                                                                         
                                                                                                     
  ## Skill docs                                                                                                                                                                               
                                                                                                                                                                                              
  - New `lark-sheets-media-upload.md`.                                                                                                                                                        
  - New `lark-sheets-{create,update,get,list,delete}-float-image.md`.                                                                                                                         
  - SKILL.md adds a Float Images table plus a pointer to `docs +media-preview`                                                                                                                
    for reading image bytes (the read APIs only return metadata + token).


shortcuts/sheets/sheet_float_image.go
```diff
@@ -0,0 +1,254 @@
...
+	if includeToken {
+		if s := runtime.Str("float-image-token"); s != "" {
+			body["float_image_token"] = s
+		}
+	}
+	if s := runtime.Str("range"); s != "" {
+		body["range"] = s
+	}
+	if runtime.Cmd.Flags().Changed("width") {
+		body["width"] = runtime.Str("width")
```
fangshuyu-768: `width`/`height`/`offset_x`/`offset_y` are documented as numeric pixel fields, but this helper serializes them from `runtime.Str(...)`, so the JSON body becomes strings like `"width":"200"`. Other sheets shortcuts that send numeric payloads declare `Type: "int"` and use `runtime.Int(...)` to avoid type mismatches. It would be safer to make these flags integer-typed and emit JSON numbers here.

shortcuts/sheets/sheet_float_image.go
```diff
@@ -0,0 +1,254 @@
...
+		{Name: "sheet-id", Desc: "sheet ID", Required: true},
+		{Name: "float-image-token", Desc: "image file token (from upload API)", Required: true},
+		{Name: "range", Desc: "anchor cell (e.g. sheetId!A1:A1)", Required: true},
+		{Name: "width", Desc: "width in pixels (>=20)"},
+		{Name: "height", Desc: "height in pixels (>=20)"},
+		{Name: "offset-x", Desc: "horizontal offset in pixels (>=0)"},
+		{Name: "offset-y", Desc: "vertical offset in pixels (>=0)"},
+		{Name: "float-image-id", Desc: "custom 10-char alphanumeric ID (auto-generated if omitted)"},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
```
fangshuyu-768, line 139: The command takes both `--sheet-id` (path param) and `--range`, but validation never checks that the range is a single anchor cell or that its sheet prefix matches `--sheet-id`. Right now `--sheet-id sheet1 --range other!A1:A1` passes locally and only fails downstream. Adding a client-side consistency check here would prevent malformed requests and give users a clearer error.

fangshuyu-768 (approved):

PR #516, opened 2026-04-16
feat(slides): add +replace-slide shortcut with block-level editing

ViperCai:
## Summary

Adds `lark-cli slides +replace-slide`, a shortcut for element-level editing of existing Lark Slides pages. Callers pass a JSON array of `block_replace` / `block_insert` parts and the CLI handles URL resolution, XML hygiene, client-side validation, and 3350001 hint enrichment over the native `xml_presentation.slide.replace` API.

Also refactors `skills/lark-slides/`: `SKILL.md` becomes a command index, with edit workflows, examples, and per-API references split into dedicated files, and a cheatsheet for supported block root elements.

## Why a dedicated shortcut

The native API has three sharp edges every caller hits:

1. **URL formats.** Users have `/slides/<token>` or `/wiki/<token>` URLs, not bare `xml_presentation_id`. The shortcut parses all three and resolves wiki URLs via `wiki.get_node`.
2. **Undocumented XML hygiene.** `block_replace` requires `id=<block_id>` on the replacement root; `<shape>` requires `<content/>`. Missing either returns a catch-all `3350001` with no guidance. The shortcut auto-injects both and surfaces the result in `--dry-run`.
3. **`3350001` is a catch-all.** The shortcut attaches a concrete common-causes checklist (missing `block_id` / invalid XML / coords out of 960×540) so agents and humans have a starting point.

## Highlights

- **Three input forms for `--presentation`**: bare token, `/slides/<token>` URL, or `/wiki/<token>` URL (auto-resolved via `wiki.get_node`).
- **Pre-API validation**: `--parts` length ∈ [1, 200]; action restricted to `block_replace` / `block_insert` (`str_replace` rejected); per-action required fields enforced.
- **Auto-injection**: `id` on replacement roots; `<content/>` on shapes (including self-closing `<shape/>`). `--dry-run` renders the exact injected body and path-encodes the presentation token to match what `Execute` sends.
- **Mixed actions supported** in a single batch (empirically verified on BOE; the shortcut does not split them).
- **Supported block root elements** documented with minimal XML snippets in `references/lark-slides-replace-slide.md`: `shape / line / polyline / img / icon / table / td / chart`. Explicit unsupported list: `video / audio / whiteboard` — these only appear as `<undefined>` export placeholders in SML 2.0, not as writable elements.

## Empirical corrections to existing docs (BOE-verified)

- `revision_id`: stale-but-existing values are accepted; only values greater than current return `3350002`.
- Wrong `block_id` → `3350001` (not a 200 with `failed_part_index`).
- Type-mismatched `block_replace` (e.g. using a shape's id with a `<td>` replacement) is silently accepted by the backend and may destroy content; `3350001` specifically signals a missing `block_id`.

## Test Plan

- [ ] `block_replace`: `id` auto-injected into replacement root; correct existing id preserved, wrong id overridden.
- [ ] `block_insert`: self-closing `<shape/>` auto-expanded with `<content/>`; omitting `insert_before_block_id` appends to page end.
- [ ] Mixed `block_replace` + `block_insert` in a single `--parts` executes successfully.
- [ ] `--presentation` with `/wiki/` URL: resolves via `get_node`; non-slides `obj_type` returns a clear error.
- [ ] `--presentation` with special-char tokens is path-encoded consistently in `--dry-run` preview and the real request URL.
- [ ] `--dry-run`: shows the fully injected request body without making an API call.
- [ ] `3350001` errors get the common-causes checklist hint; non-`3350001` errors get no slides-specific hint.
- [ ] `go test ./shortcuts/slides/...` passes.

## Summary by CodeRabbit

* **New Features**
  * Added `+replace-slide` shortcut for block-level slide replace/insert with client-side validation (1–200 parts), automatic root ID/<content/> injection, wiki-URL resolution, dry-run preview, and enriched error hints.

* **Documentation**
  * Added comprehensive guides and examples: replace-slide reference, edit workflows, examples (insert/replace), media-upload updates, XML get/replace docs, and SKILL.md refactor.

* **Tests**
  * Added extensive tests covering parsing, injection, validation, dry-run, wiki resolution, and error scenarios.


shortcuts/slides/slides_replace_slide.go
```diff
@@ -0,0 +1,351 @@
...
+		if ref.Kind == "wiki" {
+			presentationID = "<resolved_slides_token>"
+			dry.Desc("2-step orchestration: resolve wiki → replace slide parts").
+				GET("/open-apis/wiki/v2/spaces/get_node").
+				Desc("[1] Resolve wiki node to slides presentation").
+				Params(map[string]interface{}{"token": ref.Token})
+		} else {
+			dry.Desc(fmt.Sprintf("Replace %d part(s) on slide %s", len(parts), slideID))
+		}
+		dry.POST(fmt.Sprintf("/open-apis/slides_ai/v1/xml_presentations/%s/slide/replace", presentationID)).
```
fangshuyu-768: DryRun currently interpolates presentationID directly, but Execute path-encodes it with validate.EncodePathSegment. Can we mirror Execute here so the preview matches the actual request URL for tokens that contain reserved characters?

shortcuts/slides/slides_replace_slide_test.go
```diff
@@ -0,0 +1,571 @@
...
+		t.Fatal("expected error")
+	}
+	var exitErr *output.ExitError
+	if !errors.As(err, &exitErr) || exitErr.Detail == nil {
+		t.Fatalf("expected ExitError, got %v", err)
+	}
+	if exitErr.Detail.Code != 99991672 {
+		t.Fatalf("expected code 99991672, got %d", exitErr.Detail.Code)
+	}
+	if strings.Contains(exitErr.Detail.Hint, "missing <content/>") {
```
fangshuyu-768: This assertion looks too weak: buildSlides3350001Hint never emits "missing <content/>", so the test will pass even if a slides-specific 3350001 hint is added by mistake. Could we assert against the real enrichment strings, or simply require Hint to stay empty for non-3350001 errors?

fangshuyu-768 (commented):
(superseded — consolidated into a later review)

fangshuyu-768 (commented):
(superseded — consolidated into a later review)

fangshuyu-768 (commented):
(superseded — consolidated into a later review)

fangshuyu-768 (commented):
(superseded — consolidated into a later review)

fangshuyu-768 (commented):
(superseded — consolidated into a later review)

shortcuts/slides/helpers.go
```diff
@@ -148,6 +148,124 @@ func extractImagePlaceholderPaths(slideXMLs []string) []string {
...
+	afterOpen := xmlFragment[m[1]:]
+	if xmlContentTagRegex.MatchString(afterOpen) {
+		return xmlFragment
+	}
+
+	closeTag := "</" + tagName + ">"
+	if !strings.Contains(afterOpen, closeTag) {
+		return xmlFragment
+	}
+	return xmlFragment[:m[1]] + "<content/>" + afterOpen
```
fangshuyu-768, line 276: When the root `<shape>` already has non-`<content>` children (e.g. the user wrote `<shape type="text"><p>hi</p></shape>`), this injects `<content/>` as a **sibling** of `<p>`, producing `<shape><content/><p>hi</p></shape>`. Per SML 2.0, `<p>` is supposed to live inside `<content>`, so the resulting shape has an empty `<content>` and a stray `<p>` — the backend may accept it (passing the 3350001 gate) but the text won't render as the shape's body. If we're going to auto-fix missing `<content/>`, we should either (a) only inject when the shape has no children at all, or (b) wrap the existing children into `<content>…</content>`. The current behavior silently produces a schema-legal but semantically wrong layout. At minimum, the doc-comment above should call out that children are not rewrapped.

shortcuts/slides/helpers.go
```diff
@@ -148,6 +148,124 @@ func extractImagePlaceholderPaths(slideXMLs []string) []string {
...
+	if closer == "/>" {
+		prefix := xmlFragment[m[2]:m[3]]
+		attrs := xmlFragment[m[6]:m[7]]
+		trimmed := strings.TrimRight(attrs, " \t\n\r")
+		rest := xmlFragment[m[1]:]
+		return prefix + "<" + tagName + trimmed + "><content/></" + tagName + ">" + rest
+	}
+
+	afterOpen := xmlFragment[m[1]:]
+	if xmlContentTagRegex.MatchString(afterOpen) {
```
fangshuyu-768, line 258: `xmlContentTagRegex.MatchString(afterOpen)` matches any `<content` anywhere in the rest of the fragment — including deeply nested inside a child (e.g. `<shape><wrapper><content/></wrapper></shape>`) or even after the closing `</shape>` if the fragment isn't a single rooted element. The SML requirement is that `<content>` be a **direct child** of `<shape>`, not just present somewhere underneath it. Given the test matrix here is all flat structures this is low-risk today, but as soon as nested shapes/groups enter the picture the heuristic will give a false positive and skip injection. Consider scanning only the direct child span (up to the matching `</shape>` close) or walking tokens instead of substring-matching.

shortcuts/slides/slides_replace_slide.go
```diff
@@ -0,0 +1,337 @@
...
+const slides3350001Hint = "common causes: (1) block_id not found in current slide — re-run slide.get for latest XML; (2) invalid XML structure or unsupported element; (3) element coordinates exceed slide bounds (960×540)"
+
+// enrichSlidesReplaceError attaches slides3350001Hint when the API returns
+// 3350001 (invalid param). Other error codes pass through untouched.
+func enrichSlidesReplaceError(err error) error {
+	var exitErr *output.ExitError
+	if !errors.As(err, &exitErr) || exitErr.Detail == nil || exitErr.Detail.Code != larkCodeSlidesInvalidParam {
+		return err
+	}
+	exitErr.Detail.Hint = slides3350001Hint
```
fangshuyu-768: This unconditionally overwrites `exitErr.Detail.Hint`. If the backend ever attaches a more specific hint to a 3350001 response (or an upstream layer has already enriched the error), that context is clobbered and replaced with the generic checklist. Safer pattern is to only set the hint when it's empty, or to append with a separator: `if exitErr.Detail.Hint == "" { exitErr.Detail.Hint = slides3350001Hint }`. Same concern for future proofing if another wrapper runs before this one.

shortcuts/slides/slides_replace_slide.go
```diff
@@ -0,0 +1,337 @@
...
+		if err != nil {
+			return enrichSlidesReplaceError(err)
+		}
+
+		result := map[string]interface{}{
+			"xml_presentation_id": presentationID,
+			"slide_id":            slideID,
+			"parts_count":         len(injected),
+		}
+		if v := common.GetFloat(data, "revision_id"); v > 0 {
```
fangshuyu-768: `common.GetFloat(data, "revision_id"); v > 0` silently drops `revision_id == 0`. Revision ids in practice start at 1 so this is probably fine today, but the `> 0` doubles as a "did the server return this field" check, which is fragile — a future server that legitimately returns `0` would be omitted from the result. Prefer a presence check (`if raw, ok := data["revision_id"]; ok { … }`) like the block just below does for `failed_part_index` / `failed_reason`, so behavior is consistent across the three fields.

shortcuts/slides/slides_replace_slide_test.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+	if len(body.Parts) != 1 {
+		t.Fatalf("parts = %d, want 1", len(body.Parts))
+	}
+	got := body.Parts[0]
+	if got.Action != "block_replace" || got.BlockID != "bUn" {
+		t.Fatalf("part = %+v", got)
+	}
+	// The replacement must have id="bUn" injected into the <shape> root.
+	if !strings.Contains(got.Replacement, `id="bUn"`) {
```
fangshuyu-768, line 65: The input for this test is a self-closing `<shape type="rect" width="100"/>`, which should trigger **both** id-injection and the `<content/>` expansion (self-closing shape → open shape with `<content/>`). The test only asserts the `id="bUn"` injection; it would be worth also asserting that the self-closing form was expanded to `<shape …><content/></shape>` so a future change to `ensureShapeHasContent`'s call order can't silently regress this combined path. `TestReplaceSlideBlockReplacePreservesMatchingID` covers the content injection in isolation, but not alongside the id-injection branch that this test exercises.

skills/lark-slides/SKILL.md
```diff
@@ -12,357 +12,52 @@ metadata:
...
-└── Wiki Node (知识库节点, obj_type: slides)
-    └── obj_token → xml_presentation_id
-
-Slides (演示文稿)
-├── xml_presentation_id (演示文稿唯一标识)
-├── revision_id (版本号)
-└── Slide (幻灯片页面)
-    └── slide_id (页面唯一标识)
-```
+**编辑已有幻灯片页面**：优先用 [`+replace-slide`](references/lark-slides-replace-slide.md)（块级替换/插入，不动页序）；选择 action 和完整读-改-写流程见 [`lark-slides-edit-workflows.md`](references/lark-slides-edit-workflows.md)。
```
fangshuyu-768, line 17: This PR replaces ~240 lines of the original `SKILL.md` (the quick-start, workflow, style-selection table, page-layout suggestions, outline template, core rules, error cheat-sheet, and self-check checklist for **creating** PPTs) with a single one-liner pointing at `+replace-slide`. I checked the diff and the rest of the branch: that creation guidance does not appear to be migrated into `lark-slides-create.md`, `examples.md`, or any other reference. For users whose task is "create a new PPT" (the majority path the old skill served), the agent now loses the style/layout/outline scaffolding entirely.

This looks out of scope for a PR titled "add +replace-slide shortcut". Two suggestions: (1) restore the creation guidance in `SKILL.md` — or explicitly move it into `lark-slides-create.md` — so the skill stays usable for create flows; (2) split that doc restructuring into a separate PR so the scope of this one matches its title. If the deletion is intentional, it at least deserves a note in the PR description explaining where the create guidance went.

fangshuyu-768 (commented):
Nice shortcut — the 3350001 hint enrichment plus the `id` / `<content/>` auto-injection will save a lot of round-trips for callers. A few inline notes focused on edge cases in the XML helpers, the error-enrichment path, and a docs regression in `SKILL.md` that looks unrelated to the replace-slide feature.

shortcuts/slides/slides_replace_slide.go
ViperCai: Already addressed — `DryRun` on line 114 now uses `validate.EncodePathSegment(presentationID)` (same helper as `Execute`), so the preview URL matches the actual request path-encoding. Fixed in commit 762002d.

shortcuts/slides/slides_replace_slide_test.go
ViperCai: Already tightened — `TestReplaceSlideNon3350001ErrorNotEnriched` now asserts `!strings.Contains(hint, "common causes")` (the real checklist string), so adding any slides-specific hint to a non-3350001 path will fail the test. Fixed in commit 762002d.

shortcuts/slides/helpers.go
ViperCai: Acknowledged — the current regex scans the whole `afterOpen` span, which will false-positive on `<content>` buried inside a child element. Leaving as-is for now: the shortcut's supported root elements (shape/img/line/table/chart) are all flat in the SML 2.0 schema — none of them nest another `<shape>`-style element that could legitimately carry `<content>` inside it — so the current test matrix covers the real usage. Tracking a proper direct-child scan (or token walk) for the follow-up that adds nested group support.

shortcuts/slides/helpers.go
ViperCai: Fixed in 3dd0416 — `ensureShapeHasContent` now only injects `<content/>` when the shape body is empty. If the user already wrote non-content children (e.g. `<shape type="text"><p>hi</p></shape>`), the function returns the fragment untouched and lets the backend's 3350001 surface the schema issue, instead of producing the `<content/>` + sibling `<p>` layout. Updated `TestEnsureShapeHasContent` to assert the new behavior on both the `<p>` case and the `<contention/>` case.

shortcuts/slides/slides_replace_slide.go
ViperCai: Fixed in 3dd0416 — `enrichSlidesReplaceError` now only sets the generic checklist when `exitErr.Detail.Hint == ""`, so a more specific upstream hint (backend or earlier wrapper) is preserved.

shortcuts/slides/slides_replace_slide.go
ViperCai: Fixed in 3dd0416 — switched to a presence check: `if _, ok := data["revision_id"]; ok { result["revision_id"] = int(common.GetFloat(data, "revision_id")) }`. Behavior is now consistent with the `failed_part_index` / `failed_reason` branches below, and a legitimate `revision_id == 0` from a future server would be preserved.

shortcuts/slides/slides_replace_slide_test.go
ViperCai: Fixed in 3dd0416 — `TestReplaceSlideBlockReplaceInjectsID` now also asserts `<content/>` and `</shape>` are present in the outgoing replacement, so a future reorder between `ensureXMLRootID` and `ensureShapeHasContent` can't silently regress the combined self-closing path.

skills/lark-slides/SKILL.md
ViperCai: Checked against the rebased branch (3dd0416): `git diff origin/main..HEAD --stat skills/lark-slides/SKILL.md` → **60 lines changed (33+/27−), net +6**. Current file is 374 lines. There's no 240-line deletion.

The only fully removed section is the old "给已有 PPT 的已有页加图（整页替换）" block (~13 lines) that `+replace-slide` now obsoletes. Everything else is additive/refinement:

- top-of-file CRITICAL pointer to `+replace-slide`
- 必读 table: +1 row (edit-workflows)
- Workflow "给已有页加图" step: rewritten from 整页替换 to block_insert recipe
- 核心规则 7: flipped from "没有元素级编辑能力" to "优先块级替换"
- Shortcuts / API Resources / 权限表 / 错误码表 / 症状→修复表 / 参考文档表: new rows for `+replace-slide`, `xml_presentation.slide.get`, `xml_presentation.slide.replace`, 3350001, 3350002
- misc consistency fixes (XML casing, quote style)

All the create-flow scaffolding is intact and unmoved:
- 快速开始 (L37–49)
- 执行前必做 (L51–71)
- Workflow (L72–108) + jq 命令模板 (L110–133)
- 风格快速判断表 (L139–146)
- 页面布局建议 (L150–157)
- 大纲模板 (L161–174)
- 常用 Slide XML 模板 (L176–178)
- 核心概念 (L182–248)
- 核心规则 (L280–289)
- 创建前自查 (L320–330)
- 症状→修复表 (L332–352)

Nothing was migrated out to `lark-slides-create.md` / `examples.md` because nothing was removed from the create path.

PR #538, opened 2026-04-17
feat(mail): add email priority support for compose and read

chanthuang:
## Summary

Add email priority support to the mail module:

- **Write**: all compose shortcuts (`+send`, `+reply`, `+reply-all`, `+forward`, `+draft-create`) accept a new `--priority` flag (`high` / `normal` / `low`). When set, the draft's EML carries an `X-Cli-Priority` header that the backend uses to apply the priority. `+draft-edit` adds a `--set-priority` direct flag consistent with `--set-subject` / `--set-to` naming.
- **Read**: `+message` now surfaces priority in `priority_type` / `priority_type_text`, inferred from the `HIGH_PRIORITY` / `LOW_PRIORITY` labels in `label_ids` (with the existing `priority_type` field kept as a fallback).

`normal` and omitted `--priority` are equivalent — no header is set.

## Changes

- `shortcuts/mail/helpers.go`: add `priorityFlag`, `parsePriority`, `applyPriority`; extend `normalizeMessage` to derive priority from `label_ids`.
- `shortcuts/mail/mail_send.go` / `mail_reply.go` / `mail_reply_all.go` / `mail_forward.go` / `mail_draft_create.go`: register `priorityFlag`, apply the header when building the EML.
- `shortcuts/mail/mail_draft_edit.go`: register `--set-priority`; translate it to `set_header X-Cli-Priority` or `remove_header X-Cli-Priority`.
- Six skill references updated with the new flag.

## Test plan

- [x] `go build ./...`
- [x] `go test ./shortcuts/mail/...`
- [x] End-to-end verified: sending with `--priority high` / `--priority low` / omitted; reading back to see `priority_type_text` and `label_ids`; `+draft-edit --set-priority high` / `--set-priority normal` (clears priority).

## Backward compatibility

- Omitting `--priority` keeps behavior identical to before (no header).
- `--priority normal` is equivalent to omitting it.
- Read-side output is unchanged for emails without priority labels.

## Summary by CodeRabbit

* **New Features**
  * Added --priority to mail commands (send, draft-create, reply, reply-all, forward) and --set-priority to draft-edit to set email priority (high, normal, low). Omitting or using normal clears priority.
* **Behavior Change**
  * Priority is now derived from priority labels when present, overriding stored values; composed messages include the priority header when set.
* **Documentation**
  * Updated command docs to list new priority options.
* **Tests**
  * Added tests for parsing, header application, and draft-edit priority behavior.

### DIFF
--- a/shortcuts/mail/helpers.go
+++ b/shortcuts/mail/helpers.go
@@ -1163,6 +1163,7 @@ func buildMessageOutput(msg map[string]interface{}, html bool) map[string]interf
 	out["date_formatted"] = normalized.DateFormatted
 	out["message_state_text"] = normalized.MessageStateText
 	if normalized.PriorityType != "" {
+		out["priority_type"] = normalized.PriorityType
 		out["priority_type_text"] = normalized.PriorityTypeText
 	}
 	out["body_plain_text"] = normalized.BodyPlainText
--- a/shortcuts/mail/helpers.go
+++ b/shortcuts/mail/helpers.go
@@ -1743,6 +1743,17 @@ func parsePriority(value string) (string, error) {
 	}
 }
 
+// validatePriorityFlag validates the --priority flag value in Validate, so invalid
+// values are caught before Execute (and before dry-run prints an API plan).
+func validatePriorityFlag(runtime *common.RuntimeContext) error {
+	v := runtime.Str("priority")
+	if v == "" {
+		return nil
+	}
+	_, err := parsePriority(v)
+	return err
+}
+
 // applyPriority sets the X-Cli-Priority header on the EML builder if priority is non-empty.
 func applyPriority(bld emlbuilder.Builder, priority string) emlbuilder.Builder {
 	if priority == "" {
--- a/shortcuts/mail/helpers.go
+++ b/shortcuts/mail/helpers.go
@@ -1241,10 +1241,21 @@ func buildMessageForCompose(msg map[string]interface{}, urlMap map[string]string
 	out.MessageStateText = messageStateText(state)
 	out.FolderID = strVal(msg["folder_id"])
 	out.LabelIDs = toStringList(msg["label_ids"])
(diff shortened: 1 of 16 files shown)


shortcuts/mail/helpers.go
```diff
@@ -1708,6 +1720,37 @@ func priorityTypeText(priorityType string) string {
 	}
 }
 
+// priorityFlag is the common flag definition for --priority, shared by all compose shortcuts.
+var priorityFlag = common.Flag{
```
infeng, line 1724: **Issue:** `priorityFlag` is registered as an unrestricted string flag, and the compose shortcuts only call `parsePriority` from `Execute`. Because the shortcut runner executes `Validate` before `--dry-run`, invalid values are accepted by preflight: for example, `mail +send ... --priority urgent --dry-run` and `mail +draft-create ... --priority urgent --dry-run` both print a valid draft-create request, while the real execution later fails with `invalid --priority value`. That makes the new user-facing flag unreliable in dry-run/validation flows across the compose shortcuts.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Validate the priority value in the preflight path for every shortcut that registers `priorityFlag` (for example, a shared `validatePriority(runtime.Str("priority"), "--priority")` called from each `Validate`, or a declarative enum if case-insensitive values are not required). Add a regression test that an invalid priority fails validation/dry-run before the command advertises the API request.

infeng (dismissed):

shortcuts/mail/helpers.go
infeng: Fair enough. I agree this does not need to block the PR; treating it as an optional preflight consistency improvement rather than a required fix.

infeng (approved):

shortcuts/mail/helpers.go
infeng: Verified as fixed. The latest commit adds `validatePriorityFlag` to the compose shortcut Validate paths, and `--priority urgent --dry-run` now fails before printing the API request.

PR #561, opened 2026-04-20
fix(docs): trim trailing code block blanks

### LINKED ISSUE #53: docs +create: Code blocks get a spurious trailing empty line 飞书文档中的代码块总是跟一个空白行
## Summary

When creating a Feishu document from Markdown using `lark-cli docs +create`, every fenced code block in the resulting document gets an extra trailing empty line that does not exist in the source Markdown.

## Steps to Reproduce

1. Prepare a Markdown file with a fenced code block:

   ````markdown
   Some text before.

   ```
   line_one
   line_two
   ```

   Some text after.
   ````

2. Create a Feishu document:

   ```bash
   lark-cli docs +create --title "test" --content-file test.md
   ```

3. Open the created document in Feishu. The code block renders as:

   ```
   line_one
   line_two
   ⬇️ (unwanted blank line here)
   ```

## Expected Behavior

The code block should contain exactly the lines from the source Markdown, with no extra trailing blank line.

## Actual Behavior

Every code block has one extra empty line appended at the end. This is cosmetic but noticeable, especially when multiple code blocks are present — each one has the spurious blank line.

## Investigation

The Markdown content passed to the API was confirmed correct (no trailing newline in the code block). The extra line appears to be introduced by the Lark OpenAPI document creation endpoint or the CLI's Markdown-to-block conversion layer.

## Environment

- `@larksuite/cli` version: 1.0.0
- Platform: Windows 11
- Node.js: installed via npm global
- Brand: feishu

OwenYWT:
## Summary
- trim blank lines immediately before fenced code block closing markers in Markdown crossing the docs CLI boundary
- apply the normalization to docs create/update/fetch and markdown drive export
- add regression coverage for accumulated trailing blank lines while preserving internal code blank lines

Fixes #53.

## Test plan
- `go test ./shortcuts/common ./shortcuts/doc ./shortcuts/drive`
- `git diff --cached --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed trailing blank lines appearing in Markdown code blocks when creating, updating, fetching, and exporting documents, improving formatting consistency across all document operations.

### DIFF
--- a/shortcuts/common/markdown.go
+++ b/shortcuts/common/markdown.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package common
+
+import "strings"
+
+type markdownFence struct {
+	char byte
+	size int
+}
+
+func parseMarkdownFence(line string) (markdownFence, bool) {
+	trimmed := strings.TrimSpace(line)
+	if len(trimmed) < 3 {
+		return markdownFence{}, false
+	}
+	char := trimmed[0]
+	if char != '`' && char != '~' {
+		return markdownFence{}, false
+	}
+	size := 0
+	for size < len(trimmed) && trimmed[size] == char {
+		size++
+	}
+	if size < 3 {
+		return markdownFence{}, false
+	}
+	return markdownFence{char: char, size: size}, true
+}
+
+func isMarkdownFenceClose(line string, fence markdownFence) bool {
+	trimmed := strings.TrimSpace(line)
+	if len(trimmed) < fence.size {
+		return false
+	}
+	size := 0
+	for size < len(trimmed) && trimmed[size] == fence.char {
+		size++
+	}
+	return size >= fence.size && strings.TrimSpace(trimmed[size:]) == ""
+}
+
+func isMarkdownBlankLine(line string) bool {
+	return strings.TrimSpace(strings.TrimRight(line, "\r\n")) == ""
+}
+
+// TrimMarkdownCodeBlockTrailingBlanks removes blank lines immediately before
+// fenced code block closing markers. Lark's document Markdown round-trip can
+// append one such blank line per fetch/update cycle, causing code blocks to grow.
+func TrimMarkdownCodeBlockTrailingBlanks(markdown string) string {
+	if markdown == "" {
+		return markdown
+	}
+
+	lines := strings.SplitAfter(markdown, "\n")
+	out := make([]string, 0, len(lines))
+	pendingBlanks := make([]string, 0, 2)
+	var fence markdownFence
+	inCodeBlock := false
+
+	for _, line := range lines {
+		if !inCodeBlock {
+			out = append(out, line)
+			if parsed, ok := parseMarkdownFence(line); ok {
+				fence = parsed
+				inCodeBlock = true
+			}
+			continue
+		}
+
+		if isMarkdownFenceClose(line, fence) {
+			pendingBlanks = pendingBlanks[:0]
+			out = append(out, line)
+			inCodeBlock = false
+			continue
+		}
+
+		if isMarkdownBlankLine(line) {
+			pendingBlanks = append(pendingBlanks, line)
+			continue
+		}
+
+		if len(pendingBlanks) > 0 {
+			out = append(out, pendingBlanks...)
+			pendingBlanks = pendingBlanks[:0]
+		}
+		out = append(out, line)
+	}
+
+	if len(pendingBlanks) > 0 {
+		out = append(out, pendingBlanks...)
+	}
+	return strings.Join(out, "")
+}
--- a/shortcuts/common/markdown_test.go
+++ b/shortcuts/common/markdown_test.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package common
+
+import "testing"
+
+func TestTrimMarkdownCodeBlockTrailingBlanks(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "trims one trailing blank line before closing fence",
+			in:   "before\n```bash\necho hello\n\n```\nafter\n",
+			want: "before\n```bash\necho hello\n```\nafter\n",
+		},
+		{
+			name: "trims accumulated trailing blank lines",
+			in:   "```go\nfmt.Println(1)\n\n\n```\n",
+			want: "```go\nfmt.Println(1)\n```\n",
+		},
+		{
+			name: "keeps intentional blank lines inside code content",
+			in:   "```\nline one\n\nline two\n\n```\n",
+			want: "```\nline one\n\nline two\n```\n",
+		},
+		{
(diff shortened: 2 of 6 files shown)


shortcuts/doc/docs_create.go
```diff
@@ -68,7 +68,7 @@ var DocsCreate = common.Shortcut{
 
 func buildDocsCreateArgs(runtime *common.RuntimeContext) map[string]interface{} {
 	args := map[string]interface{}{
-		"markdown": runtime.Str("markdown"),
+		"markdown": common.TrimMarkdownCodeBlockTrailingBlanks(runtime.Str("markdown")),
```
fangshuyu-768, line 71: **Does this actually close issue #53?**

Issue #53 reports that a freshly-created Feishu document, opened **in the Feishu UI**, renders an extra blank line at the end of every code block — the user explicitly confirms the markdown *sent* to the API was already clean. That suggests the blank line is introduced by the MCP `create-doc` tool or the downstream docx block conversion, not by the CLI-side markdown.

This PR normalises markdown on the CLI side of the boundary (trim on create input, trim on fetch/export output). That will:
- ✅ Prevent accumulation across fetch → edit → update round-trips (good and worth having).
- ✅ Make `docs +fetch` / `drive +export` output look tidy.
- ❌ Not change what renders inside Feishu after `docs +create` — the docx block created by the server still contains the server-added trailing blank.

If the intent is to close #53 as reported, the trim-on-input is a no-op here (the user's source markdown already didn't have the blank). Could we either (a) verify the fix in Feishu UI after this PR and update the issue, or (b) reword the PR / issue link from "Fixes #53" to "Mitigates round-trip growth related to #53" so it's clear the rendering issue remains open?

shortcuts/common/markdown.go
```diff
@@ -0,0 +1,95 @@
...
+	size := 0
+	for size < len(trimmed) && trimmed[size] == fence.char {
+		size++
+	}
+	return size >= fence.size && strings.TrimSpace(trimmed[size:]) == ""
+}
+
+func isMarkdownBlankLine(line string) bool {
+	return strings.TrimSpace(strings.TrimRight(line, "\r\n")) == ""
+}
```
fangshuyu-768, line 46: **`strings.TrimRight(line, "\r\n")` is redundant here.**

`strings.TrimSpace` already strips `\r`, `\n`, `\t`, and space, so the chained `TrimRight(line, "\r\n")` has no effect on the result. Can simplify to:

```go
func isMarkdownBlankLine(line string) bool {
    return strings.TrimSpace(line) == ""
}
```

shortcuts/common/markdown.go
```diff
@@ -0,0 +1,95 @@
...
+
+import "strings"
+
+type markdownFence struct {
+	char byte
+	size int
+}
+
+func parseMarkdownFence(line string) (markdownFence, bool) {
+	trimmed := strings.TrimSpace(line)
```
fangshuyu-768, line 14: **`TrimSpace` over-tolerates indentation — 4-space-indented "fences" are actually indented code block content.**

CommonMark allows 0–3 spaces of indentation before a fence marker; 4+ spaces means the line is an indented code block *inside* the surrounding context and the backticks are literal. Because `parseMarkdownFence` does `strings.TrimSpace(line)`, a line like `"    ```"` (four leading spaces) is treated as an opening fence, and subsequent lines are then ignored by `TrimMarkdownCodeBlockTrailingBlanks` until it sees a matching close.

In practice this is mostly academic for markdown coming out of Lark (server output isn't deeply indented), but it can trigger false positives if a user's source passed through `docs +create`/`docs +update` nests a literal fence example inside an indented block. Suggest matching CommonMark: count leading spaces, bail out if ≥ 4.

Same applies to `isMarkdownFenceClose` at L33.

shortcuts/common/markdown.go
```diff
@@ -0,0 +1,95 @@
...
+			pendingBlanks = pendingBlanks[:0]
+		}
+		out = append(out, line)
+	}
+
+	if len(pendingBlanks) > 0 {
+		out = append(out, pendingBlanks...)
+	}
+	return strings.Join(out, "")
+}
```
fangshuyu-768, line 95: **Two parallel fence parsers now live in the repo — worth a DRY note for a follow-up.**

`shortcuts/doc/markdown_fix.go` has its own `applyOutsideCodeFences` that does fence tracking with a simpler, less-robust rule (it only recognises triple backticks, not `~~~`, not variable fence length). This new `shortcuts/common/markdown.go` has more correct fence handling (either char, `>= 3` length, matching close length).

Not a blocker for this PR, but once this lands it'd be worth switching `applyOutsideCodeFences` to drive off `parseMarkdownFence` / `isMarkdownFenceClose` so the two implementations can't drift (e.g. the old one doesn't currently handle `~~~` fences at all — see `fixExportedMarkdown` pipeline). Feel free to file a follow-up and leave this PR as-is.

shortcuts/common/markdown_test.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			if got := TrimMarkdownCodeBlockTrailingBlanks(tt.in); got != tt.want {
+				t.Fatalf("TrimMarkdownCodeBlockTrailingBlanks() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
```
fangshuyu-768, line 52: **A few test-coverage gaps worth plugging while the test file is fresh:**

- CRLF line endings (Windows-authored markdown — the issue reporter is on Windows 11).
- Fences longer than 3 chars: `` `````go\n…\n\n````` `` (5-backtick open, 5-backtick close) must be preserved; a 3-backtick line in the middle is *not* a close.
- Input that does not end with a trailing newline (`SplitAfter` retains the final token without `\n`; worth confirming round-trip is byte-exact).
- Unclosed fenced block (fence opened, EOF reached before a close) — current code keeps `pendingBlanks` at the end; the assertion just protects against regression.
- Tab-only "blank" lines inside code (`\t\n`): currently dropped as blank by `TrimSpace`; confirm that matches intent for code blocks where tab-only lines can have semantic meaning (e.g. Makefiles, whitespace-sensitive DSLs).
- Tilde fence with info string: `~~~python\nx\n\n~~~` (already have one tilde case, but without info string).

None of these are likely to be broken in the current implementation, but the test file doubles as a spec for future changes.

fangshuyu-768 (commented):
A few notes on top of what CodeRabbit already flagged (the `docs_fetch.go` double type-assert and the `normalizeDocsUpdateResult` consistency in `docs_create.go`). Biggest concern is scope vs. issue #53 — the rest are minor correctness / test-gap / DRY notes.

SunPeiYang996:
Thanks for putting this fix together, and for adding the related regression coverage.

This issue has already been resolved in the V2 document flow, and V2 together with the latest document skill is the path we plan to maintain going forward. For that reason, we will not merge this CLI-side compatibility trimming logic for now, to avoid introducing duplicate handling or subtle behavior differences across the V1/V2 paths.

Users are encouraged to upgrade to the latest version and use the latest document skill, which uses the V2 flow to address this issue:

```bash
lark-cli update
```

The investigation and the coverage across `docs create/update/fetch` and `drive export` are still appreciated. We will close this PR for now; if a similar issue shows up again in the latest V2 flow, we can follow up directly there.

PR #569, opened 2026-04-20
feat(doc): add pre-write semantic warnings to docs +update

herbertliu:
## Summary

Add two static semantic checks to `docs +update` that run before the MCP
`update-doc` call. They inform the user about CLI/MCP contract edges that
commonly cause confusing round-trip results, without blocking execution.

## Changes

- **`shortcuts/doc/docs_update_check.go`** (new): `docsUpdateWarnings(mode, markdown)` returns human-readable warnings based on:
  1. **`replace_*` + blank-line markdown** — `replace_range` / `replace_all` only swap text inside an existing block, so `\n\n` renders as literal text, not a paragraph break. Hint: use `delete_range + insert_before`.
  2. **Combined bold+italic emphases** (`***text***`, `**_text_**`, `_**text**_`) — Lark stores only one of the two emphases; these patterns silently downgrade. Hint: split into two separate emphases.
- **`shortcuts/doc/docs_update.go`**: wire the checks into `Execute` before `CallMCPTool`, writing each warning to stderr prefixed with `warning:`.
- **`shortcuts/doc/docs_update_check_test.go`** (new): table-driven tests for both checks plus aggregation and empty-input cases.

## Behavior

- Warnings only — the update still proceeds after printing them.
- Not emitted in dry-run mode (kept quiet during planning).
- `replace_*`-only: the blank-line check does NOT fire on `insert_*` / `append` / `overwrite` where multi-paragraph markdown is valid.

## Test Plan

- [x] `go test ./shortcuts/doc/...` passes
- [x] `go vet ./shortcuts/...` clean
- [x] `gofmt -l .` clean
- [x] `golangci-lint run --new-from-rev=origin/main` shows `0 issues`
- [x] Manual: sample invocations with replace_range + `\n\n`, with `***text***`, and both together each emit the expected warning lines

## Related

First batch of the AI-Agent pitfall review (Cases 1 and 5). Case 2
(heading-type warning) is intentionally deferred — it requires an extra
block fetch and should be a separate PR with a `--strict` / verbose
opt-in, so it doesn't add latency to every `docs +update`.

## Summary by CodeRabbit

* **New Features**
  * Pre-submit validation for documentation updates now warns about potential formatting issues (multiline paragraph breaks in replace modes and combined bold+italic patterns).
  * Warnings are printed to the console prefixed with "warning:" before running the update, and ignore fenced code and inline code to reduce false positives.

* **Tests**
  * Added unit and end-to-end tests covering the new semantic warnings and verifying that dry-run executions suppress those warnings.

### DIFF
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package doc
+
+import (
+	"regexp"
+	"strings"
+)
+
+// docsUpdateWarnings returns a list of human-readable warnings for a
+// `docs +update` invocation based on static analysis of the mode and
+// Markdown payload. The warnings describe CLI/MCP contract edges that
+// commonly surprise users; the update is still executed — callers
+// decide whether to stop at a warning.
+//
+// Warnings emitted (current):
+//
+//  1. replace_* modes do not split blocks. A Markdown payload containing
+//     a blank line (\n\n) implies the caller expects multiple paragraphs,
+//     but replace_range / replace_all only swap in-block text. The
+//     resulting block will contain the blank line as literal text and
+//     appear as a single paragraph in the UI.
+//
+//  2. Lark does not round-trip bold+italic. Markdown like ***text*** or
+//     **_text_** / _**text**_ is stored as only one of the two emphases
+//     (usually italic), silently dropping the other. The user wanted
+//     both; they will get one.
+func docsUpdateWarnings(mode, markdown string) []string {
+	var warnings []string
+	if w := checkDocsUpdateReplaceMultilineMarkdown(mode, markdown); w != "" {
+		warnings = append(warnings, w)
+	}
+	if w := checkDocsUpdateBoldItalic(markdown); w != "" {
+		warnings = append(warnings, w)
+	}
+	return warnings
+}
+
+// checkDocsUpdateReplaceMultilineMarkdown flags markdown that contains a
+// blank-line paragraph break under a replace_* mode. Returns an empty
+// string when the combination is fine.
+func checkDocsUpdateReplaceMultilineMarkdown(mode, markdown string) string {
+	if mode != "replace_range" && mode != "replace_all" {
+		return ""
+	}
+	// A CR/LF-robust check: both "\n\n" and "\r\n\r\n" count as paragraph
+	// separators. We normalize line endings once before the substring match.
+	normalized := strings.ReplaceAll(markdown, "\r\n", "\n")
+	if !strings.Contains(normalized, "\n\n") {
+		return ""
+	}
+	return "--mode=" + mode + " does not split a block into multiple paragraphs; " +
+		"the blank line in --markdown will render as literal text. " +
+		"For multiple paragraphs, use --mode=delete_range followed by --mode=insert_before."
+}
+
+// reBoldItalicTriple matches ***text*** with non-whitespace text between.
+var reBoldItalicTriple = regexp.MustCompile(`\*\*\*\S[^*]*?\S\*\*\*|\*\*\*\S\*\*\*`)
+
+// reBoldItalicUnderscoreInside matches **_text_** — bold wrapping an
+// underscore italic. Same downgrade issue in Lark.
+var reBoldItalicUnderscoreInside = regexp.MustCompile(`\*\*_\S[^_*]*?\S_\*\*|\*\*_\S_\*\*`)
+
+// reBoldItalicUnderscoreOutside matches _**text**_ — underscore italic
+// wrapping a bold.
+var reBoldItalicUnderscoreOutside = regexp.MustCompile(`_\*\*\S[^_*]*?\S\*\*_|_\*\*\S\*\*_`)
+
+// checkDocsUpdateBoldItalic flags Markdown emphases that attempt to
+// combine bold and italic in a way Lark cannot represent.
+func checkDocsUpdateBoldItalic(markdown string) string {
+	if markdown == "" {
+		return ""
+	}
+	if reBoldItalicTriple.MatchString(markdown) ||
+		reBoldItalicUnderscoreInside.MatchString(markdown) ||
+		reBoldItalicUnderscoreOutside.MatchString(markdown) {
+		return "Lark does not support combined bold+italic markers (***text***, **_text_**, _**text**_); " +
+			"the emphasis will be downgraded to either bold or italic. " +
+			"Split into two separate emphases or drop one of them."
+	}
+	return ""
+}
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -14,18 +14,26 @@ import (
 // commonly surprise users; the update is still executed — callers
 // decide whether to stop at a warning.
 //
+// Both checks ignore fenced code blocks (```…``` and ~~~…~~~, with up
+// to 3 leading spaces per CommonMark §4.5), inline code spans, and
+// backslash-escaped emphasis markers so that literal Markdown content
+// embedded in code samples or escaped prose does not produce false
+// positives.
+//
 // Warnings emitted (current):
 //
 //  1. replace_* modes do not split blocks. A Markdown payload containing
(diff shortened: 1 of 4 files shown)


shortcuts/doc/docs_update_check.go
```diff
@@ -0,0 +1,83 @@
...
+// blank-line paragraph break under a replace_* mode. Returns an empty
+// string when the combination is fine.
+func checkDocsUpdateReplaceMultilineMarkdown(mode, markdown string) string {
+	if mode != "replace_range" && mode != "replace_all" {
+		return ""
+	}
+	// A CR/LF-robust check: both "\n\n" and "\r\n\r\n" count as paragraph
+	// separators. We normalize line endings once before the substring match.
+	normalized := strings.ReplaceAll(markdown, "\r\n", "\n")
+	if !strings.Contains(normalized, "\n\n") {
```
fangshuyu-768: This blank-line check will also fire for fenced code blocks, inline code, or literal Markdown examples that contain `\n\n` but are not meant to become multiple paragraphs. Could we strip or skip code fences / code spans before the paragraph-break check so the warning only applies to real prose?

shortcuts/doc/docs_update_check.go
```diff
@@ -0,0 +1,83 @@
...
+// wrapping a bold.
+var reBoldItalicUnderscoreOutside = regexp.MustCompile(`_\*\*\S[^_*]*?\S\*\*_|_\*\*\S\*\*_`)
+
+// checkDocsUpdateBoldItalic flags Markdown emphases that attempt to
+// combine bold and italic in a way Lark cannot represent.
+func checkDocsUpdateBoldItalic(markdown string) string {
+	if markdown == "" {
+		return ""
+	}
+	if reBoldItalicTriple.MatchString(markdown) ||
```
fangshuyu-768: These regexes seem likely to false-positive inside inline code, fenced code, or escaped markers such as `\\***escaped***`. It may be safer to ignore code spans / code fences and escaped emphasis markers before running the bold+italic detection.

shortcuts/doc/docs_update_check_test.go
```diff
@@ -0,0 +1,169 @@
...
+				t.Errorf("hint should suggest delete_range/insert_before remediation, got: %s", got)
+			}
+		})
+	}
+}
+
+func TestCheckDocsUpdateBoldItalic(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
fangshuyu-768, line 160: It may be worth adding a few negative cases here for fenced code, inline code, and escaped markers, so future changes do not reintroduce false positives for literal Markdown content.

shortcuts/doc/docs_update_check.go
fangshuyu-768: Resolved in 199aadb32 (follow-up #578). `checkDocsUpdateReplaceMultilineMarkdown` now delegates to `proseHasBlankLine`, which tracks ```/~~~ fence state line-by-line and only considers blank lines outside fence regions. New negative tests: `blank_line_inside_backtick_fenced_code_is_not_flagged`, `blank_line_inside_tilde_fenced_code_is_not_flagged`, plus mixed-case positives confirming prose blanks still fire when the fence also contains blanks.

shortcuts/doc/docs_update_check.go
fangshuyu-768: Resolved in 199aadb32 (follow-up #578). `checkDocsUpdateBoldItalic` now runs `stripEscapedEmphasisMarkers(stripMarkdownCodeRegions(markdown))` before the regex sweep — fenced code lines are blanked, inline code spans are masked with equal-length whitespace (via `scanInlineCodeSpans`), and `\*` / `\_` escapes are stripped. The regex set is also table-driven across all six combined-emphasis shapes (`***…***`, `___…___`, `**_…_**`, `__*…*__`, `_**…**_`, `*__…__*`). One documented limitation remains for the rare `\\***text***` (literal double-backslash + real emphasis) shape — the simple string replace can't reason about double escapes, and a full CommonMark escape parser felt like overkill for this heuristic.

shortcuts/doc/docs_update_check_test.go
fangshuyu-768: Resolved in 199aadb32 (follow-up #578). Added 14+ negative cases across fenced code (backtick and tilde), inline code spans, and backslash-escaped markers (`\*` and `\_`), covering all six combined-emphasis shapes. Also added two positive tests (`real_triple_asterisks_outside_inline_code_still_flags`, `real_emphasis_after_a_fence_closed_by_longer_marker_still_flags`) to guard against over-sanitization by the strip step.

shortcuts/doc/docs_update_check_test.go
```diff
@@ -0,0 +1,375 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got := checkDocsUpdateReplaceMultilineMarkdown(tt.mode, tt.markdown)
+			hasHint := got != ""
+			if hasHint != tt.wantHint {
+				t.Fatalf("checkDocsUpdateReplaceMultilineMarkdown(%q, %q) = %q, wantHint=%v",
+					tt.mode, tt.markdown, got, tt.wantHint)
+			}
+			if tt.wantHint && !strings.Contains(got, "delete_range") {
+				t.Errorf("hint should suggest delete_range/insert_before remediation, got: %s", got)
```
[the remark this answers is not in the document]
fangshuyu-768: Addressed in #581. The assertion now requires both `delete_range` and `insert_before` in the warning string — dropping either half of the remediation will fail `TestCheckDocsUpdateReplaceMultilineMarkdown`.

shortcuts/doc/docs_update_check_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+func TestDocsUpdateWarningsEmpty(t *testing.T) {
+	t.Parallel()
+
+	// Clean markdown in a non-replace mode produces zero warnings.
+	warnings := docsUpdateWarnings("insert_before", "plain paragraph text")
+	if len(warnings) != 0 {
+		t.Fatalf("expected no warnings, got: %v", warnings)
+	}
+}
```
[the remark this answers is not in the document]
fangshuyu-768: Addressed in #581. Added `tests/cli_e2e/docs/docs_update_dryrun_test.go` → `TestDocs_UpdateDryRunSuppressesSemanticWarnings`, which invokes `docs +update --dry-run --mode=replace_range --markdown "***combined***\n\nsecond"` (crafted to trip both pre-write warnings) and asserts neither the `warning:` prefix nor either warning body appears in stdout/stderr.

Sanity-checked the test's regression-detection power locally: temporarily wired `docsUpdateWarnings` into `DryRun`, rebuilt, and the E2E failed with the injected warnings visible on stderr. Reverted and re-ran — passes. The test has real teeth, not just a rubber stamp.

fangshuyu-768 (approved):

shortcuts/doc/docs_update_check.go
herbertliu: Thanks — this was addressed in the final merged version. `checkDocsUpdateReplaceMultilineMarkdown` now runs through `proseHasBlankLine`, which tracks fenced code block state (both ``` and ~~~, CommonMark §4.5 fence tolerance) and only counts a blank line when it sits between real prose lines. Blank lines inside fences are no longer flagged.

shortcuts/doc/docs_update_check.go
herbertliu: Thanks — addressed in the final version. The bold+italic check now sanitizes the markdown first via `stripMarkdownCodeRegions` (blanks out fenced code blocks AND inline code spans) and `stripEscapedEmphasisMarkers` (neutralizes `\*`/`\_`/`\``). The detection was also expanded to six documented shapes (added `___text___`, `__*text*__`, `*__text__*`). Escaped markers and literal samples inside code no longer trigger it.

shortcuts/doc/docs_update_check_test.go
herbertliu: Done — the merged test file now covers all the false-positive angles you called out: blank lines inside backtick and tilde fences, triple-asterisks/underscores inside fenced code and inline code spans, and escaped emphasis markers. Also added CommonMark §4.5 edge cases (3-space indented fences, longer close markers, 4-space-indent disqualifying a fence) to guard against regressions there.

PR #595, opened 2026-04-21
feat: add read receipt support (--request-receipt + +send-receipt)

xzcong0820:
# PR Summary: Read Receipt Support for Lark CLI Mail                                                                                                                                                                  
                                                                                                                                                                                                                        
  ## Overview
  This pull request adds end-to-end RFC 3798 Message Disposition Notification support to the `lark-cli mail` module, enabling both requesting and responding to read receipts.                                          
                                                                  
  ## Key Features

  **Request Side (Compose Commands)**
  - New `--request-receipt` flag for `+send`, `+reply`, `+reply-all`, `+forward`, `+draft-create`, and `+draft-edit`
  - Adds `Disposition-Notification-To` header addressed to the sender                                                                                                                                                   
  - Validates sender address resolution before proceeding
  - SKILL.md workflow explicitly gates the flag: agents must NOT infer intent from subject / body — `--request-receipt` is added only on explicit user request                                                          
                                                                                                                                                                                                                        
  **Response Side**
  - New `+send-receipt` shortcut: confirms "I have read this" by sending a system-templated receipt reply                                                                                                               
    - Non-customizable body (subject / recipient / sent / read time), mirroring Outlook / Thunderbird / Lark client behavior                                                                                            
    - Gated by `high-risk-write` framework requiring `--yes` flag                                                                                                                                                       
    - Targets messages carrying the `READ_RECEIPT_REQUEST` label                                                                                                                                                        
  - New `+decline-receipt` shortcut: dismisses the read-receipt banner WITHOUT sending a receipt                                                                                                                        
    - Mirrors the Lark client's "不发送 / Don't send" button                                                                                                                                                            
    - Clears only the `READ_RECEIPT_REQUEST` label via `batch_modify_message`; no outgoing mail                                                                                                                         
    - `Risk: "write"` (not `high-risk-write`) — no `--yes` needed; idempotent on re-run                                                                                                                                 
                                                                                                                                                                                                                        
  **Read-Path Hints**                                                                                                                                                                                                   
  - `+message`, `+messages`, and `+thread` commands emit stderr tips when encountering receipt requests                                                                                                                 
  - Hint exposes BOTH response paths so agents present a real choice:
    - `+send-receipt --yes …` — user agrees to confirm read                                                                                                                                                             
    - `+decline-receipt …` — user silences the banner without confirming                                                                                                                                                
                                                                                                                                                                                                                        
  ## Design Highlights                                                                                                                                                                                                  
                                                                  
  The receipt subject prefix detects language from original content: CJK text triggers "已读回执：" while other content uses "Read receipt: ". Receipt body generation uses system templates rather than free-form text,
   mirroring client behavior.
                                                                                                                                                                                                                        
  Backend coupling involves the private header `X-Lark-Read-Receipt-Mail: 1`, extracted into `BodyExtra.IsReadReceiptMail` for label management.                                                                        
  
  Security measures include validation against CR/LF injection and dangerous Unicode (Bidi overrides, zero-width characters), with shell-safe quoting for hint commands and control-character stripping from terminal   
  output.                                                         
                                                                                                                                                                                                                        
  ## Testing & Documentation                                      

  - Unit tests cover header validation, language detection, body generation, hint rendering (including line-injection defense), subject idempotence, and receipt-vs-decline shortcut metadata                           
  - Integration tests (httpmock) cover: `+send` / `+reply` / `+reply-all` / `+forward` attaching the DNT header; `+send` rejecting `--request-receipt` when no sender is resolvable; `+decline-receipt` label removal
  happy path and already-cleared idempotence                                                                                                                                                                            
  - New reference docs `lark-mail-send-receipt.md` and `lark-mail-decline-receipt.md` explain constraints, consent requirements, and response fields
  - Skill workflow (section 9) describes the full privacy-safe decision tree on both sides 

## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added `--request-receipt` flag to mail composition commands (`+draft-create`, `+draft-edit`, `+reply`, `+reply-all`, `+forward`, `+send`) to request read receipts from recipients.
  * Added `+send-receipt` shortcut to send read-receipt responses.
  * Added `+decline-receipt` shortcut to dismiss read-receipt prompts.

* **Documentation**
  * Updated reference documentation for all new shortcuts and flags.


shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,349 @@
...
+	Description: "Send a read-receipt reply for an incoming message that requested one (i.e. carries the READ_RECEIPT_REQUEST label). Body is auto-generated (subject / recipient / send time / read time) to match the Lark client's receipt format — callers cannot customize it, matching the industry norm that read-receipt bodies are system-generated templates, not free-form replies. Intended for agent use after the user confirms.",
+	Risk:        "high-risk-write",
+	Scopes: []string{
+		"mail:user_mailbox.message:send",
+		"mail:user_mailbox.message:modify",
+		"mail:user_mailbox.message:readonly",
+		"mail:user_mailbox:readonly",
+		"mail:user_mailbox.message.address:read",
+		"mail:user_mailbox.message.subject:read",
+	},
```
infeng, line 84: **Issue:** `+send-receipt` now reuses `fetchFullMessage(..., false)`, which issues `message.get(format=plain_text_full)` at runtime, but this shortcut's `Scopes` block does not declare `mail:user_mailbox.message.body:read`. The existing read shortcuts that use the same fetch path (`+message`, `+messages`, `+thread`) all declare that scope. As written, callers can satisfy the static scope set here and still fail only after starting the command, even though the receipt flow only needs labels / subject / sender / references / internal_date.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Either add `mail:user_mailbox.message.body:read` everywhere this command advertises/auth-checks scopes, or switch the receipt flow to a metadata-only message fetch so it does not depend on body-read at all.

shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,349 @@
...
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		messageId := runtime.Str("message-id")
+		mailboxID := resolveComposeMailboxID(runtime)
+		return common.NewDryRunAPI().
+			Desc("Send read receipt: fetch the original message → verify the READ_RECEIPT_REQUEST label is present → build a reply with subject \"已读回执：<original>\" (zh) or \"Read Receipt: <original>\" (en) picked by CJK detection on the original subject, In-Reply-To / References threading, and X-Lark-Read-Receipt-Mail: 1 → create draft and send. The backend extracts the private header, sets BodyExtra.IsReadReceiptMail, and DraftSend applies the READ_RECEIPT_SENT label to the outgoing message.").
+			GET(mailboxPath(mailboxID, "messages", messageId)).
+			GET(mailboxPath(mailboxID, "profile")).
+			POST(mailboxPath(mailboxID, "drafts")).
+			Body(map[string]interface{}{"raw": "<base64url-EML>"}).
+			POST(mailboxPath(mailboxID, "drafts", "<draft_id>", "send"))
```
infeng, line 101: **Issue:** DryRun no longer matches Execute. The real path calls `fetchFullMessage(..., false)`, which sends `message.get` with `format=plain_text_full`, but the advertised DryRun sequence only shows a bare `GET /messages/{id}`. That hides the new preflight dependency and makes it harder to catch auth or response-shape mismatches before the user runs the command for real.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Keep DryRun in lockstep with Execute by adding the same `format=plain_text_full` param (or whatever metadata-only format the final implementation settles on), and add a regression test for the receipt path so future preflight drift is caught.

skills/lark-mail/SKILL.md
```diff
@@ -59,6 +59,7 @@ metadata:
 6. **新邮件** — `+send` 存草稿（默认），加 `--confirm-send` 发送
 7. **确认投递** — 立即发送后用 `send_status` 查询投递状态，定时发送后在预定时间后再查询；取消定时发送用 `cancel_scheduled_send`
 8. **编辑草稿** — `+draft-edit` 修改已有草稿。正文编辑通过 `--patch-file`：回复/转发草稿用 `set_reply_body` op 保留引用区，普通草稿用 `set_body` op
+9. **已读回执** — 写信时加 `--request-receipt` 请求对方回执；拉信看到 `label_ids` 含 `READ_RECEIPT_REQUEST`（或 `-607`）时，**必须先问用户**是否回执，用户同意后用 `+send-receipt` 响应（不要自动回执，涉及隐私）
```
infeng: 在skill-template/domains/mail.md 也要有这个改动，这个文件应该是自动生成的，在 registry 仓库执行 make all project=mail

xzcong0820:
@coderabbitai resume

xzcong0820:
@coderabbitai review

xzcong0820:
@coderabbitai review

shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,360 @@
...
+		// +message / +thread which all list this scope.
+		"mail:user_mailbox.message.body:read",
+	},
+	AuthTypes: []string{"user"},
+	Flags: []common.Flag{
+		{Name: "message-id", Desc: "Required. Message ID of the incoming mail that requested a read receipt.", Required: true},
+		{Name: "mailbox", Desc: "Mailbox email address that owns the receipt reply (default: me)."},
+		{Name: "from", Desc: "Sender email address for the From header. Defaults to the mailbox's primary address."},
+	},
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
```
chanthuang, line 91: **Issue:** `messageId` should be `messageID` per Go naming conventions ([Go wiki: Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)). The rest of this codebase consistently uses `messageID` — e.g. `maybeHintReadReceiptRequest(... messageID string, ...)`, `normalizeMessageID()`, `fetchFullMessage(... messageID ...)`. This local is used throughout both `DryRun` and `Execute` in this file.

**Severity:** 💡 Suggestion

**Suggested Fix:** Rename `messageId` → `messageID` in both `DryRun` (line 91) and `Execute` (line 108) closures, plus all downstream references.

shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,360 @@
...
+		"mail:user_mailbox.message:readonly",
+		"mail:user_mailbox:readonly",
+		"mail:user_mailbox.message.address:read",
+		"mail:user_mailbox.message.subject:read",
+		// +send-receipt doesn't read the body content itself, but
+		// fetchFullMessage(..., false) uses format=plain_text_full which
+		// the backend scope-checks against body:read. Declared explicitly
+		// to keep the static Scopes truthful and aligned with +triage /
+		// +message / +thread which all list this scope.
+		"mail:user_mailbox.message.body:read",
```
chanthuang, line 83: Note to other reviewers: an earlier comment on this block flagged `mail:user_mailbox.message.body:read` as missing from `Scopes` (🚫 Must Fix). **That is a false positive** — the scope IS declared on the line immediately below (with a 4-line comment explaining why it is needed despite `+send-receipt` not reading body content). The fetch path uses `format=plain_text_full` which the backend scope-checks against `body:read`.

shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,360 @@
...
+		{Name: "message-id", Desc: "Required. Message ID of the incoming mail that requested a read receipt.", Required: true},
+		{Name: "mailbox", Desc: "Mailbox email address that owns the receipt reply (default: me)."},
+		{Name: "from", Desc: "Sender email address for the From header. Defaults to the mailbox's primary address."},
+	},
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		messageId := runtime.Str("message-id")
+		mailboxID := resolveComposeMailboxID(runtime)
+		return common.NewDryRunAPI().
+			Desc("Send read receipt: fetch the original message → verify the READ_RECEIPT_REQUEST label is present → build a reply with subject \"已读回执：<original>\" (zh) or \"Read Receipt: <original>\" (en) picked by CJK detection on the original subject, In-Reply-To / References threading, and X-Lark-Read-Receipt-Mail: 1 → create draft and send. The backend extracts the private header, sets BodyExtra.IsReadReceiptMail, and DraftSend applies the READ_RECEIPT_SENT label to the outgoing message.").
+			GET(mailboxPath(mailboxID, "messages", messageId)).
```
chanthuang: Note to other reviewers: an earlier comment flagged DryRun as "showing a bare GET without format param" (🚨 Should Fix). **That is a false positive** — this line already includes `.Params(map[string]interface{}{"format": messageGetFormat(false)})`, which correctly reflects the `fetchFullMessage(..., false)` call in Execute.

xzcong0820:
@coderabbitai review

xzcong0820:
@coderabbitai full review

xzcong0820:
@coderabbitai full review

shortcuts/mail/mail_forward.go
```diff
@@ -125,6 +129,12 @@ var MailForward = common.Shortcut{
 		if senderEmail != "" {
 			bld = bld.From("", senderEmail)
 		}
+		if err := requireSenderForRequestReceipt(runtime, senderEmail); err != nil {
```
infeng: 🚨 Should fix: this has the same fallback problem that `+reply-all` already corrected above. When `resolveComposeSenderEmail(runtime)` is empty, `senderEmail` falls back to `orig.headTo`, which is just the first original `To` recipient, not a verified sender identity. With `--request-receipt`, that can silently write `Disposition-Notification-To` for somebody else in CC/shared-mailbox flows. Please validate `--request-receipt` against the pre-fallback resolved sender here as well, and add a forward regression test that keeps the current mailbox out of the original `To` list.

shortcuts/mail/mail_reply.go
```diff
@@ -136,6 +140,12 @@ var MailReply = common.Shortcut{
 		if senderEmail != "" {
 			bld = bld.From("", senderEmail)
 		}
+		if err := requireSenderForRequestReceipt(runtime, senderEmail); err != nil {
```
infeng: 🚨 Should fix: `requireSenderForRequestReceipt` now runs **after** `senderEmail` falls back to `orig.headTo`, but `orig.headTo` is only `toList[0]` from the original message (`helpers.go`), not the current caller's identity. In a CC/shared-mailbox case, `--request-receipt` will therefore emit `Disposition-Notification-To` for another recipient instead of forcing the caller to resolve/provide their own sender address. `+reply-all` moved this guard before the fallback in the same PR for exactly this reason; `+reply` needs the same treatment, plus a regression test where the mailbox is only in CC.

shortcuts/mail/mail_send_receipt.go
```diff
@@ -0,0 +1,360 @@
...
+			Desc("Send read receipt: fetch the original message → verify the READ_RECEIPT_REQUEST label is present → build a reply with subject \"已读回执：<original>\" (zh) or \"Read receipt: <original>\" (en) picked by CJK detection on the original subject, In-Reply-To / References threading, and X-Lark-Read-Receipt-Mail: 1 → create draft and send. The backend extracts the private header, sets BodyExtra.IsReadReceiptMail, and DraftSend applies the READ_RECEIPT_SENT label to the outgoing message.").
+			GET(mailboxPath(mailboxID, "messages", messageID)).
+			Params(map[string]interface{}{"format": messageGetFormat(false)}).
+			GET(mailboxPath(mailboxID, "profile")).
+			POST(mailboxPath(mailboxID, "drafts")).
+			Body(map[string]interface{}{"raw": "<base64url-EML>"}).
+			POST(mailboxPath(mailboxID, "drafts", "<draft_id>", "send"))
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		return validateConfirmSendScope(runtime)
```
chanthuang: **Issue:** `validateConfirmSendScope` is a no-op here. That function gates on `runtime.Bool("confirm-send")` — returning nil immediately when the flag is false. Since `+send-receipt` has no `--confirm-send` flag (it uses `Risk: "high-risk-write"` + `--yes` instead), the scope check for `mail:user_mailbox.message:send` never actually runs.

Consequence: a user missing the send scope won't get a clear `missing_scope` error at validation time; instead, the command proceeds through message fetch → EML build → draft create, only to fail at `drafts.send` with a less informative backend error.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Replace with a direct scope check that doesn't gate on `--confirm-send`:
```go
Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
    return validateSendScope(runtime) // check send scope unconditionally
},
```
Or inline the scope check if no such helper exists.

xzcong0820: Partial true, but conclusion doesn't hold.                                                                                                                                                                          
  ▎                                                               
  ▎ You're right that validateConfirmSendScope is effectively a no-op here — it short-circuits on runtime.Bool("confirm-send"), and +send-receipt doesn't expose that flag. But the consequence ("error only surfaces at
  ▎  drafts.send") doesn't follow: mail:user_mailbox.message:send is already declared in this shortcut's static Scopes (lines 71–84), and runner.checkShortcutScopes (shortcuts/common/runner.go:625) pre-checks the 
  ▎ static scope list against the stored token before Validate runs. A user with a stale token gets the exact same ExitAuth / missing_scope error (with the auth login --scope recovery hint) one phase earlier than    
  ▎ Validate, never reaching draft create or send.                
  ▎
  ▎ So validateSendScope-style dynamic check would be redundant. The right cleanup is to align with +send, which also keeps :send in static Scopes and skips validateConfirmSendScope for the same reason. Pushed       
  ▎ https://github.com/larksuite/cli/pull/595/commits/c1814cb removing the no-op call and adding a comment explaining why this Validate is intentionally absent. Thanks for catching the dead code path even though the 
  ▎ impact analysis didn't pan out — it pushed the cleanup that aligns with +send.

chanthuang (approved):
LGTM. All issues from my previous reviews have been addressed.

PR #605, opened 2026-04-22
feat(drive): pre-flight 10000-rune total cap for +add-comment reply_elements

herbertliu:
## Summary

The open-platform comment endpoint `/open-apis/drive/v1/files/{token}/new_comments` rejects oversized payloads with an opaque `[1069302] Invalid or missing parameters` — no field name, no length information, no indication that length is even the cause. Callers (especially agents) end up binary-searching the content to figure out what hit the cap.

This PR adds a pre-flight check inside `parseCommentReplyElements` so `drive +add-comment` rejects over-cap input client-side with a structured, actionable error before any network round-trip.

> The first version of this PR set a 300-byte per-text-element limit, justified by the empirical pattern "~80 Chinese chars OK, ~130 Chinese chars fail." A subsequent round of probing the live API showed that pattern doesn't reproduce and the actual contract is materially different (cap is 100× higher, on the total across elements, and rune-based not byte-based). The current implementation reflects the corrected findings; see "Empirical basis" below.

## Behavior

- **Cap**: combined character (rune) count across **all** `type=text` elements in `--content` must be ≤ **10000**.
- **Counting**: server counts raw runes; byte width and HTML escape state are irrelevant. 10000 ASCII = 10000 Chinese = 10000 `<` chars all sit at the limit, even though they encode to 10000 / 30000 / 40000 bytes respectively.
- **Multi-element**: the cap is on the *sum*, not per element. Splitting one long element into multiple smaller ones does **not** bypass the cap; the hint explicitly says so.
- **Non-text elements**: `mention_user` (open_id) and `link` (URL) don't count toward the budget.
- **Failure shape**: structured `text_too_long` ExitError naming the offending element index and the running total at that point. Hint walks the caller through the fix.

```
$ lark-cli drive +add-comment --doc "$DOCX" --type docx --full-comment --content '[{"type":"text","text":"…10001 chars…"}]'

[validation] --content reply_elements text totals 10001 characters at element #1 (this element: 10001); the server caps the combined length at 10000 characters across ALL reply_elements

hint: shorten the comment so the combined text across all reply_elements fits within 10000 characters. The server enforces this cap on the TOTAL — splitting one long element into multiple smaller text elements does NOT help (they all add up against the same 10000-rune budget). Server returns an opaque [1069302] on overflow, so this check is pre-flight; no escape transform changes the count (server reads raw runes).
```

## Empirical basis

Probed via `drive file.comments create_v2` against a fresh docx, bypassing the shortcut so the server's actual responses are visible:

| Input | Result |
|---|---|
| 10000 ASCII chars (10000 B) | ✅ OK |
| 10001 ASCII chars (10001 B) | ❌ `[1069302]` |
| 10000 Chinese chars (30000 B) | ✅ OK |
| 10001 Chinese chars (30003 B) | ❌ `[1069302]` |
| 10000 `<` chars (would-be 40000 B if escaped) | ✅ OK |
| 10000 `&` chars | ✅ OK |
| 5000 a + 5000 b across two elements (total 10000) | ✅ OK |
| 5000 a + 5001 b across two elements (total 10001) | ❌ `[1069302]` |
| 9999 a + 1 b across two elements (total 10000) | ✅ OK |
| 4000 + 4000 + 4000 across three elements (total 12000) | ❌ `[1069302]` |

Two takeaways: (1) the cap is on **total runes**, not per-element bytes; (2) splitting doesn't bypass the cap, so the previous "split into multiple elements" advice was actively misleading and is no longer in the code, the tests, or the skill md.

The schema doc currently advertises "1-1000 characters" for `text` elements. The live API accepts 10× that. The discrepancy is documented in a comment on `maxCommentTotalRunes`; if the schema is later corrected to 10000 (or any other value), re-probe with `drive file.comments create_v2` and update the constant.

## Changes

- **`shortcuts/drive/drive_add_comment.go`** — `parseCommentReplyElements` now tracks a running rune total across reply_elements and rejects at the first element pushing the cumulative count past 10000. Counts the raw input via `utf8.RuneCountInString`; escape (`<` → `&lt;`, `>` → `&gt;`) still happens for rendering but no longer participates in the length check. The `maxCommentTotalRunes = 10000` constant carries an inline doc comment with the probe matrix.
- **`shortcuts/drive/drive_add_comment_test.go`** — `TestParseCommentReplyElementsTextLength` rewritten with 11 cases: single-element boundaries (ASCII / Chinese / angle brackets at exactly 10000 and 10001), multi-element total-cap (5000+5000 OK, 5000+5001 rejected with the rejection index pointing at element #2), early-element overshoot (first element at 10001 reports index #1, not the trailing element), and `mention_user` not double-counting. New `TestParseCommentReplyElementsHintForbidsSplitAdvice` watchdog fails CI if anyone re-introduces the discredited "split into multiple text elements" advice into the hint.
- **`skills/lark-drive/references/lark-drive-add-comment.md`** — replaces the old "300-byte per-element / split as workaround" bullets with two short ones describing the 10000-rune total cap and the `mention_user` / `link` exemption.

## Test Plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./shortcuts/drive/... -count=1` passes; race-mode pass clean
- [x] `golangci-lint run --new-from-rev=origin/main` — 0 issues
- [x] `node scripts/skill-format-check/index.js skills` passes
- [x] Full unit suite (`go test $(go list ./... | grep -v cli_e2e) -count=1`) — all packages green
- [x] Live API smoke against a real docx with the new binary: 100 / 5000 / 10000-char single elements accepted, 10001 / 10100 single rejected pre-flight, 5000+5000 split accepted, 5000+5001 split rejected pre-flight (matches the empirical matrix above)

### DIFF
--- a/shortcuts/drive/drive_add_comment.go
+++ b/shortcuts/drive/drive_add_comment.go
@@ -18,6 +18,23 @@ import (
 
 const defaultLocateDocLimit = 10
 
+// maxCommentTextElementBytes is the per-text-element byte budget the server
+// accepts in `drive +add-comment` reply_elements.
+//
+// The open-platform comment API returns an opaque `[1069302] Invalid or
+// missing parameters` when a single `{"type":"text","text":"…"}` exceeds
+// this limit, with no indication that length is the cause or which element
+// is at fault. Empirically, Chinese text up to ~80 characters (≈240 bytes
+// UTF-8) lands reliably and ~130 characters (≈390 bytes) fails. 300 bytes
+// is inside the known-safe zone: it catches the 130-char failure case
+// pre-flight while leaving headroom for typical ~100-char comments.
+//
+// ASCII callers can fit ~300 chars; Chinese callers ~100 chars per element.
+// Agents that have a longer message should split it across multiple text
+// elements — the comment UI still renders them as a single contiguous
+// comment.
+const maxCommentTextElementBytes = 300
+
 type commentDocRef struct {
 	Kind  string
 	Token string
@@ -612,12 +629,25 @@ func parseCommentReplyElements(raw string) ([]map[string]interface{}, error) {
 			if strings.TrimSpace(input.Text) == "" {
 				return nil, output.ErrValidation("--content element #%d type=text requires non-empty text", index)
 			}
-			if utf8.RuneCountInString(input.Text) > 1000 {
-				return nil, output.ErrValidation("--content element #%d text exceeds 1000 characters", index)
+			// Measure the escaped form, not the raw input, because the server
+			// receives the escaped text and any '<' / '>' / '&' expand to 4-5
+			// bytes (`&lt;`, `&gt;`, `&amp;`). Without this, an input that
+			// hits the limit on one of those characters would still surface
+			// as the opaque [1069302] from the server.
+			escaped := escapeCommentText(input.Text)
+			if byteLen := len(escaped); byteLen > maxCommentTextElementBytes {
+				runeCount := utf8.RuneCountInString(input.Text)
+				return nil, output.ErrWithHint(
+					output.ExitValidation,
+					"text_too_long",
+					fmt.Sprintf("--content element #%d text is %d characters (%d bytes); per-element limit is ~%d bytes (≈100 Chinese characters / 300 ASCII characters)",
+						index, runeCount, byteLen, maxCommentTextElementBytes),
+					"split the content across multiple {\"type\":\"text\",\"text\":\"...\"} elements — the comment UI still renders them as one contiguous comment. The server returns an opaque [1069302] on overflow, so this check is pre-flight. Note: '<' / '>' / '&' are HTML-escaped and counted in their escaped form (4-5 bytes each).",
+				)
 			}
 			replyElements = append(replyElements, map[string]interface{}{
 				"type": "text",
-				"text": escapeCommentText(input.Text),
+				"text": escaped,
 			})
 		case "mention_user":
 			mentionUser := firstNonEmptyString(input.MentionUser, input.Text)
--- a/shortcuts/drive/drive_add_comment.go
+++ b/shortcuts/drive/drive_add_comment.go
@@ -18,22 +18,34 @@ import (
 
 const defaultLocateDocLimit = 10
 
-// maxCommentTextElementBytes is the per-text-element byte budget the server
-// accepts in `drive +add-comment` reply_elements.
+// maxCommentTotalRunes is the cap on the combined character (rune) count
+// across all `reply_elements[].text` fields in a single
+// `drive +add-comment` request.
 //
-// The open-platform comment API returns an opaque `[1069302] Invalid or
-// missing parameters` when a single `{"type":"text","text":"…"}` exceeds
-// this limit, with no indication that length is the cause or which element
-// is at fault. Empirically, Chinese text up to ~80 characters (≈240 bytes
-// UTF-8) lands reliably and ~130 characters (≈390 bytes) fails. 300 bytes
-// is inside the known-safe zone: it catches the 130-char failure case
-// pre-flight while leaving headroom for typical ~100-char comments.
+// The open-platform `/open-apis/drive/v1/files/{token}/new_comments`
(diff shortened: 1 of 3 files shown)


fangshuyu-768 (commented):
Two concerns on the byte-limit pre-flight.

**1. Byte check happens before HTML escaping (`shortcuts/drive/drive_add_comment.go`, lines ~579 and ~590).** `len(input.Text)` is measured before `escapeCommentText` runs, but each `<` / `>` becomes 4 bytes (`&lt;` / `&gt;`) after escaping. A payload with even one `<` at exactly 300 bytes raw becomes 303 bytes after escaping, which the server may still reject with the same opaque [1069302]. Suggest measuring the escaped output:
\`\`\`go
escaped := escapeCommentText(input.Text)
if byteLen := len(escaped); byteLen > maxCommentTextElementBytes {
    // ...
}
\`\`\`
and storing the already-escaped value to avoid double work later.

**2. `mention_user` and `link` elements bypass the byte limit (lines ~569–586).** If the server enforces a per-element limit regardless of type, those should be validated too. Worth confirming whether the limit is text-only, and either adding the same check or documenting the exemption.

shortcuts/drive/drive_add_comment.go
```diff
@@ -559,8 +576,15 @@ func parseCommentReplyElements(raw string) ([]map[string]interface{}, error) {
 			if strings.TrimSpace(input.Text) == "" {
 				return nil, output.ErrValidation("--content element #%d type=text requires non-empty text", index)
 			}
-			if utf8.RuneCountInString(input.Text) > 1000 {
-				return nil, output.ErrValidation("--content element #%d text exceeds 1000 characters", index)
+			if byteLen := len(input.Text); byteLen > maxCommentTextElementBytes {
```
fangshuyu-768: Two concerns on this pre-flight:

**1. Byte check happens before HTML escaping.** `len(input.Text)` is measured here, but `escapeCommentText(input.Text)` runs at line ~586 and turns each `<` / `>` into 4 bytes (`&lt;` / `&gt;`). A payload with even one `<` at exactly 300 bytes raw becomes 303 bytes after escaping, which the server may still reject with the same opaque [1069302]. Suggest measuring the escaped output and storing the already-escaped value:
```go
escaped := escapeCommentText(input.Text)
if byteLen := len(escaped); byteLen > maxCommentTextElementBytes {
    // ...
}
replyElements = append(replyElements, map[string]interface{}{
    "type": "text",
    "text": escaped,
})
```
The existing `TestParseCommentReplyElementsEscapesAngleBrackets` does not cover this byte-limit interaction.

**2. Only the `text` branch is checked.** The `mention_user` and `link` cases below bypass the byte limit entirely. If the server enforces a per-element limit regardless of type, those should be validated too. Worth confirming whether the limit is text-only and either adding the same check in those branches or documenting the exemption in the skill md.

fangshuyu-768 (commented):
Inline note; expands on the earlier summary review.

shortcuts/drive/drive_add_comment.go
herbertliu: Fixed in 8aa340e — addressing both concerns:

**1. Byte check now happens after escape.** Refactored `parseCommentReplyElements` to compute `escaped := escapeCommentText(input.Text)` first, then measure `len(escaped)` against the 300-byte budget, and store the already-escaped value (no double escape). The error message now ends with a note that '<' / '>' / '&' are HTML-escaped and counted in their escaped form (4-5 bytes each). Two new regression tests pin the boundary:
- '99 ascii + 75 <' (raw 174B, escaped 399B) is now rejected with '(399 bytes)' in the message
- '60 <' alone (raw 60B, escaped 240B) is accepted

**2. mention_user / link branches.** Documented in the skill md that the limit only applies to `type=text`. Both `mention_user` (`ou_xxxx` open_id, ~30 chars) and `link` (URL) are inherently far below 300 bytes by construction; adding the same byte check there would mostly be defensive code without a real failure mode. The skill md notes that if anyone hits a server-side 1069302 on those types, please file an issue and we can extend the check.

fangshuyu-768 (approved):

PR #620, opened 2026-04-23
docs(lark-doc): document read-only block types that create/update silently skip

herbertliu:
## Summary

Addresses Case 6 in the lark-cli pitfall list: `docs +create` / `docs +update` returns success but silently drops blocks the underlying `create-doc` MCP tool cannot write (sync blocks, some AddOns, anything that round-trips as `<!-- Unsupported block type: N -->` from `fetch-doc`). Agents hit this repeatedly when scaffolding weekly reports or other templates and get no signal the block is gone.

Rather than adding a runtime warning (which would require MCP-side block-type inventory that CLI doesn't own), this PR documents the known read-only block set and the signals an Agent can use to detect a silent drop, so the skill consumer knows to avoid these tags in generated markdown.

## Changes

- **`skills/lark-doc/references/lark-doc-block-type-limits.md`** (new): read-only block inventory (`<reference-synced>`, `<source-synced>`), the `<!-- Unsupported block type: N -->` marker origin, detection signals, recommended workarounds
- **`skills/lark-doc/SKILL.md`**: new "Block type 写入限制" section linking the new reference right after the shortcuts table

## Test Plan

- [x] Docs-only change; no code modified
- [x] Links resolve within the skill tree

## Summary by CodeRabbit

* **Documentation**
  * Clarified that document create/update operations do not write every block type and some blocks are effectively read-only.
  * Added a reference guide describing which block types are skipped or emitted as placeholders on fetch, troubleshooting signals for missing content, and guidance to avoid generating unsupported sync tags when authoring or binding UI.


skills/lark-doc/references/lark-doc-block-type-limits.md
```diff
@@ -0,0 +1,42 @@
+# Block type limits (read-only blocks)
+
+> **前置条件：** 先阅读 [`../SKILL.md`](../SKILL.md) 了解 `docs +create` / `docs +update` 的调用方式。
+
+`docs +create` / `docs +update` 底层的 `create-doc` MCP 工具**不支持部分 block 类型**。当 markdown 中出现这些块时，服务端会静默跳过或以 HTML 注释兜底，**API 返回成功但 block 不会真的写入**，是 AI Agent 经常踩坑的盲区。
```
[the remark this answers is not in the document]
herbertliu: Fixed in 7449025 — split the opening paragraph so write-path (silent skip) and fetch-path ('' placeholder) are explicitly two separate behaviors, avoiding the implication that create-doc itself emits the comment.

fangshuyu-768 (commented):
Reviewed the changes; a few concerns about consistency with the rest of the lark-doc skill before this lands.

**1. Tag name mismatch (`skills/lark-doc/references/lark-doc-block-type-limits.md`, lines ~19–20).** The new table uses `<reference-synced>` / `<source-synced>` (kebab-case, reversed word order), but the existing skill is consistent with `<synced_reference>` / `<synced_source>` (snake_case): see `skills/lark-doc/SKILL.md:52` and `skills/lark-doc/references/lark-doc-xml.md:48`. Agents copying from this new doc will produce tags the create/update path doesn't recognize. Please align with the existing form.

**2. Read-only inventory is incomplete (the "已知只读 block" section).** `skills/lark-doc/references/lark-doc-xml.md:48` already names five non-creatable block types: `synced_reference`, `synced_source`, `bitable`, `base_ref`, `okr`. This table only documents the first two; consider adding the other three so the doc is the single source of truth.

**3. "Silently skipped" is narrower than reality.** The phrasing "API returns success but the block does not appear" reads as if these blocks never exist in documents. Per `lark-doc-xml.md:48`, they can be moved/fetched, just not created. Worth distinguishing "block can exist via move/fetch" from "block cannot be created/updated" so agents don't mis-handle fetched output that contains them.

skills/lark-doc/references/lark-doc-block-type-limits.md
```diff
@@ -0,0 +1,47 @@
...
+
+两条路径都不会抛错，是 AI Agent 经常踩坑的盲区。
+
+本文档列出已知的只读 / 受限 block 类型，供 Agent 在撰写 markdown 前自检。
+
+## 已知只读 block（API 只支持 fetch，不支持 create / update）
+
+| 块类型 | Markdown / HTML 形式 | 现象 | 推荐做法 |
+|--------|---------------------|------|---------|
+| 引用同步块 | `<reference-synced source-block-id="..." source-document-id="...">...</reference-synced>` | 静默跳过；API 返回成功但文档中不会出现此块 | 通过 UI 手动绑定；或在 skill 中把"同步块占位"作为单独的手工步骤记录 |
```
fangshuyu-768: Tag name is inconsistent with the rest of the lark-doc skill. `skills/lark-doc/SKILL.md:52` and `skills/lark-doc/references/lark-doc-xml.md:48` both use `<synced_reference>` / `<synced_source>` (snake_case). This new doc uses `<reference-synced>` / `<source-synced>` (kebab-case, reversed word order). Agents copying from here will produce tags the create/update path doesn't recognize. Please align with the existing form.

fangshuyu-768 (commented):
Inline notes below — these expand on the earlier summary review and anchor each finding to a specific line.

skills/lark-doc/references/lark-doc-block-type-limits.md
```diff
@@ -0,0 +1,47 @@
...
+`docs +create` / `docs +update` 底层的 `create-doc` MCP 工具**不支持部分 block 类型**。具体行为分为两条路径：
+
+- **写入路径（create / update）**：markdown 中出现这些 block 时，写入阶段会被静默跳过，API 返回 `code=0 / success=true`，但 block 根本没进文档。
+- **读取路径（`docs +fetch` 导出）**：同类无法稳定序列化成 markdown 的原生 block，会以占位注释 `<!-- Unsupported block type: N -->` 呈现。
+
+两条路径都不会抛错，是 AI Agent 经常踩坑的盲区。
+
+本文档列出已知的只读 / 受限 block 类型，供 Agent 在撰写 markdown 前自检。
+
+## 已知只读 block（API 只支持 fetch，不支持 create / update）
```
fangshuyu-768: Read-only inventory is incomplete. `skills/lark-doc/references/lark-doc-xml.md:48` already names five non-creatable block types: `synced_reference`, `synced_source`, `bitable`, `base_ref`, `okr`. This table only documents the first two — consider adding `bitable`, `base_ref`, and `okr` so this doc is the single source of truth.

skills/lark-doc/references/lark-doc-block-type-limits.md
```diff
@@ -0,0 +1,47 @@
+# Block type limits (read-only blocks)
+
+> **前置条件：** 先阅读 [`../SKILL.md`](../SKILL.md) 了解 `docs +create` / `docs +update` 的调用方式。
+
+`docs +create` / `docs +update` 底层的 `create-doc` MCP 工具**不支持部分 block 类型**。具体行为分为两条路径：
+
+- **写入路径（create / update）**：markdown 中出现这些 block 时，写入阶段会被静默跳过，API 返回 `code=0 / success=true`，但 block 根本没进文档。
```
fangshuyu-768: Phrasing is narrower than reality. `lark-doc-xml.md:48` describes these blocks as `不可创建，仅支持移动` — they can exist in documents (via fetch / move), just can't be created. Saying `静默跳过；API 返回成功但文档中不会出现此块` reads as if they never appear, which can mislead agents handling fetched output that contains them. Worth distinguishing 'block can exist via move/fetch' from 'block cannot be created/updated'.

skills/lark-doc/references/lark-doc-block-type-limits.md
herbertliu: Fixed in 245a0a5 — switched both the reference doc and the SKILL.md callout to the snake_case forms (`<synced_reference>` / `<synced_source>`), matching what `lark-doc-xml.md:48` and `SKILL.md:52` already use. Added a 'XML 标签 (snake_case)' column header so future contributors don't reintroduce the kebab-case form.

skills/lark-doc/references/lark-doc-block-type-limits.md
herbertliu: Fixed in 245a0a5 — extended the inventory to all five non-creatable block types from `lark-doc-xml.md:48` (`synced_reference`, `synced_source`, `bitable`, `base_ref`, `okr`). Each row gets its own typical XML form and migration guidance. Added a note about the `block_copy_insert_after` exclusion for the same set + `task`.

skills/lark-doc/references/lark-doc-block-type-limits.md
herbertliu: Fixed in 245a0a5 — rewrote the opening to explicitly separate three paths: create/update (silently skipped), fetch (serialized normally for blocks already in the doc), and block_move_after (works across documents). Replaced the misleading '不会真的出现' phrasing with '不能创建/更新，但可以 fetch / move'. The new tagline is: 'these blocks aren't fully read-only, just _can't be newly created_'.

fangshuyu-768:
Pushed three follow-up commits — flagging them here so the trail is visible alongside the resolved review threads.

**1. Dead link to `../../lark-cli-dev/SKILL.md`** (commit `12f5399`)

The trailing "相关 MCP 层记录" section pointed at a skill that doesn't exist in the repo: `find skills -type d -name "lark-cli*"` returns nothing, git log shows no history, and the topics named in the link description (有序列表编号重置 / callout emoji 误配 / 代码块嵌套围栏) don't appear anywhere else in `skills/`. Dropped the section entirely. If a real target shows up later, link it from a fresh edit rather than carry the phantom reference forward.

**2. `§三` / `§四` references didn't deep-link** (commit `12f5399`)

Two cross-references to `lark-doc-xml.md` named the section number in the link text but the URL was just the filename, so readers landed at the top of the file. Converted both to GitHub anchor form (`lark-doc-xml.md#三资源块`, `lark-doc-xml.md#四块级复制与移动`), matching the CJK-anchor convention already used in `skills/lark-mail/references` and `skills/lark-base/references`.

**3. `<!-- Unsupported block type: N -->` section wasn't qualified as v1-specific** (commit `d1845a5`)

The 5-block read-only inventory (synced_reference, synced_source, bitable, base_ref, okr) is rooted in Lark's data model and applies to both v1 MCP and v2 OpenAPI write paths — that part of the doc was correctly version-agnostic. The placeholder comment, however, is a serialization fallback specific to v1 MCP `fetch-doc` rendering to markdown. v2 fetch (`POST /open-apis/docs_ai/v1/documents/<token>/fetch`) defaults to XML and represents blocks structurally, so this comment form doesn't appear there. Without the qualifier, an agent reading this doc on v2 might either (a) chase a placeholder that never materializes, or (b) treat its absence as evidence the block was successfully written — both lead to the same silent-drop pitfall the PR is trying to surface.

Edits: scoped the placeholder mention in the "概念区分" bullet, added a `(v1 fetch / markdown 输出)` qualifier to the section heading + a one-line v2-fetch contrast at the top of the section, broadened the "create-doc 不会把注释解析" closing line to "v1 / v2 写入路径都不会" (so the don't-feed-comments-back advice survives v2 migration), and tagged signal #2 in the detection list as v1-specific.

Net diff for all three: `+10 / -12` lines, no new sections.

SunPeiYang996:
Thanks for documenting this issue. The pitfall is real, but I do not think we should encode this limitation in the skill documentation, so I would like to reject this PR for now.

Main reasons:

1. AI Agents should not need to memorize which block types cannot be created and can only be moved. What the Agent actually needs is accurate runtime feedback from the API response when blocks are filtered out: which blocks were skipped, why they were skipped, and what the suggested next step is. We plan to add this kind of diagnostic hint to the API response later.

2. Guiding the Agent through skill documentation turns a temporary implementation limitation into a long-term context burden. It consumes the Agent's attention and context window, and the information can easily become stale as the API evolves.

3. If create/update supports more block types in the future, users with an older cached skill would still be guided by outdated documentation and would not naturally benefit from the upgraded API behavior. For this reason, capability boundaries like this are better reported by the runtime interface instead of being hardcoded in the skill.

So I would like to reject this PR for now. The preferred direction is to improve the create/update response diagnostics, for example by returning skipped / unsupported blocks and their reasons.

PR #623, opened 2026-04-23
feat(doc): add docs +batch-update for multi-operation sequential updates

herbertliu:
## Summary

Addresses Case 10 in the lark-cli pitfall list: Agents applying several related edits to one document currently pay N MCP round-trips and get N separate \`{success:true}\` responses, with no unified error/recovery story. This PR adds a \`docs +batch-update\` shortcut that accepts a JSON array of operations and runs them sequentially against a single document.

## Changes

- **\`shortcuts/doc/docs_batch_update.go\`** (new): \`DocsBatchUpdate\` shortcut; \`batchUpdateOp\` / \`batchUpdateResult\` types; \`parseBatchUpdateOps\` / \`validateBatchUpdateOp\` / \`buildBatchUpdateArgs\` helpers.
- **\`shortcuts/doc/shortcuts.go\`**: register the new shortcut.
- **\`shortcuts/doc/docs_batch_update_test.go\`** (new): parse (8 cases) + validate (9 cases) + build-args (3 subtests) coverage.

## Explicit non-goal: atomicity

There is no server-side transaction for \`update-doc\`. A mid-batch failure leaves the document in a partial-apply state. The shortcut description states this up front. Callers choose how to handle it:

- \`--on-error=stop\` (default): halt at the first failure; the response still reports how many ops landed.
- \`--on-error=continue\`: best-effort; every op runs, each entry in \`results[]\` has its own \`success\`/\`error\`.

Pair with \`docs +fetch\` after a partial run to reconcile state manually.

## Shape

\`\`\`bash
lark-cli docs +batch-update --doc <URL> --operations '[
  {\"mode\":\"replace_range\",\"markdown\":\"New intro\",\"selection_by_title\":\"## Intro\"},
  {\"mode\":\"insert_before\",\"markdown\":\"> Note: revised\",\"selection_with_ellipsis\":\"## Intro\"},
  {\"mode\":\"delete_range\",\"selection_with_ellipsis\":\"stale section...end\"}
]'
\`\`\`

Response:

\`\`\`json
{
  \"doc\": \"<URL>\",
  \"total\": 3,
  \"applied\": 3,
  \"stopped_early\": false,
  \"results\": [
    {\"index\":0, \"mode\":\"replace_range\", \"success\":true, \"result\":{...}},
    ...
  ]
}
\`\`\`

## Validation semantics

Single-op validation runs against every op **before** any MCP call, so a malformed entry fails the whole invocation up front rather than after N successful ops. This matches what batch-users expect from \`kubectl apply\`-style tooling.

## Test Plan

- [x] \`go test ./shortcuts/doc/...\` passes
- [x] \`go vet ./...\` clean
- [x] \`gofmt -l\` clean
- [x] Coverage: parse/validate/build 90-100%
- [ ] Manual smoke: run a 3-op batch against a test doc and verify \`results[]\` shape + partial-failure path

## Summary by CodeRabbit

* **New Features**
  * Added a batch document update command: accepts a JSON array of operations, supports dry-run (planned per-op sequence), executes ops sequentially, and returns aggregated metadata plus per-op results.

* **Bug Fixes / Behavior**
  * Improved validation (JSON shape, UTF‑8 BOM trimming, mutually exclusive selection rules, mode-specific requirements), advisory warnings per op, configurable stop-on-error vs continue behavior, and proper handling of cancellation with partial results.

* **Tests**
  * Extensive tests covering parsing, validation, dry-run output, execution scenarios (full success, stop-on-failure, continue-on-error) and cancellation.

### DIFF
--- a/shortcuts/doc/docs_batch_update.go
+++ b/shortcuts/doc/docs_batch_update.go
@@ -49,6 +49,23 @@ var validBatchOnError = map[string]bool{
 // left in a partial-apply state; pair with --on-error=stop + `docs +fetch`
 // to recover manually. This tradeoff is explicit in the shortcut's
 // description.
+//
+// Two design notes worth calling out for callers:
+//
+//  1. Validate runs once up front against the static op list, before any
+//     MCP write. It cannot foresee in-batch staleness — for example, if
+//     op[0] deletes a section that op[1]'s --selection-by-title resolves
+//     into, op[1] passes Validate but fails at execute time. This is by
+//     design (sequential semantics; each op sees the doc state produced
+//     by the previous op). If you need stricter atomicity, split the
+//     batch or fetch the doc between groups of related ops.
+//
+//  2. On --on-error=stop the shortcut emits TWO failure signals: the
+//     stdout JSON envelope (with stopped_early=true and the partial
+//     result list), AND a non-zero exit via the returned error. They are
+//     complementary: scripts that key on exit code see "the batch did
+//     not complete cleanly"; callers parsing JSON see "exactly which
+//     ops succeeded and how far we got". Don't treat them as redundant.
 var DocsBatchUpdate = common.Shortcut{
 	Service:     "docs",
 	Command:     "+batch-update",
@@ -138,7 +155,7 @@ var DocsBatchUpdate = common.Shortcut{
 				}
 				continue
 			}
-			normalizeDocsUpdateResult(out, op.Markdown)
+			normalizeWhiteboardResult(out, op.Markdown)
 			results = append(results, batchUpdateResult{
 				Index: i, Mode: op.Mode, Success: true, Result: out,
 			})
@@ -206,7 +223,7 @@ func parseBatchUpdateOps(raw string) ([]batchUpdateOp, error) {
 // own Validate phase, before any MCP work, so a single malformed op fails
 // the whole invocation up front instead of after N successful ops.
 func validateBatchUpdateOp(op batchUpdateOp) error {
-	if !validModes[op.Mode] {
+	if !validModesV1[op.Mode] {
 		return fmt.Errorf("invalid mode %q, valid: append | overwrite | replace_range | replace_all | insert_before | insert_after | delete_range", op.Mode)
 	}
 	if op.Mode != "delete_range" && op.Markdown == "" {
@@ -215,11 +232,11 @@ func validateBatchUpdateOp(op batchUpdateOp) error {
 	if op.SelectionWithEllipsis != "" && op.SelectionByTitle != "" {
 		return fmt.Errorf("selection_with_ellipsis and selection_by_title are mutually exclusive")
 	}
-	if needsSelection[op.Mode] && op.SelectionWithEllipsis == "" && op.SelectionByTitle == "" {
+	if needsSelectionV1[op.Mode] && op.SelectionWithEllipsis == "" && op.SelectionByTitle == "" {
 		return fmt.Errorf("mode=%s requires selection_with_ellipsis or selection_by_title", op.Mode)
 	}
 	if op.SelectionByTitle != "" {
-		if err := validateSelectionByTitle(op.SelectionByTitle); err != nil {
+		if err := validateSelectionByTitleV1(op.SelectionByTitle); err != nil {
 			return err
 		}
 	}
--- a/shortcuts/doc/docs_batch_update.go
+++ b/shortcuts/doc/docs_batch_update.go
@@ -128,6 +128,21 @@ var DocsBatchUpdate = common.Shortcut{
 		successCount := 0
 
 		for i, op := range ops {
+			// Honor cancellation between ops so a user interrupt or parent
+			// timeout doesn't have to wait for the rest of the batch. Emit
+			// the partial envelope first so callers see exactly how far we
+			// got, mirroring the --on-error=stop shape.
+			if err := ctx.Err(); err != nil {
+				runtime.Out(map[string]interface{}{
+					"doc":           runtime.Str("doc"),
+					"total":         len(ops),
+					"applied":       successCount,
+					"results":       results,
+					"stopped_early": true,
+				}, nil)
+				return err
+			}
+
 			// Re-run the same static warnings the single-op shortcut emits so
 			// batch users get the same advisory signal per-op.
 			for _, w := range docsUpdateWarnings(op.Mode, op.Markdown) {
@@ -200,7 +215,12 @@ func buildBatchUpdateArgs(docID string, op batchUpdateOp) map[string]interface{}
 // with a clearer error on the two mistakes users make most often: passing a
 // single object instead of an array, or passing an empty array.
 func parseBatchUpdateOps(raw string) ([]batchUpdateOp, error) {
-	trimmed := strings.TrimSpace(raw)
+	// Strip a leading UTF-8 BOM (U+FEFF) before the prefix check.
+	// strings.TrimSpace doesn't treat BOM as whitespace, so payloads written
+	// by editors / shells that prepend a BOM (PowerShell redirection, some
+	// Windows editors) would otherwise fail the "[" prefix probe with a
+	// confusing "must be a JSON array" error.
+	trimmed := strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(raw), "\ufeff"))
 	if trimmed == "" {
 		return nil, common.FlagErrorf("--operations is required")
 	}
--- a/shortcuts/doc/docs_batch_update.go
+++ b/shortcuts/doc/docs_batch_update.go
@@ -1,0 +1,227 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package doc
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+// batchUpdateOp is one entry in the --operations JSON array. The field set
+// mirrors the flags of `docs +update` so an operations file is effectively a
+// serialized batch of +update invocations against a single document.
+//
+// Omit-empty is deliberate: a caller may submit an append / overwrite op
+// without any selection, and the MCP call itself ignores unset fields.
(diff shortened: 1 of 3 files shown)


fangshuyu-768 (commented):
Some design questions about partial-failure recovery for sequential batches; nothing blocking, but worth thinking through before this is the recommended path.

**1. Two error signals on `--on-error=stop`.** When op N fails, `Execute` calls `runtime.Out()` to emit the partial result, then returns `callErr`. Callers see both a JSON result (with `stopped_early=true`) and a non-zero exit code. Single-update doesn't have this ambiguity because it fails fast. Worth deciding whether one of the two signals is sufficient.

**2. Result envelope lacks a per-op block-id map.** The PR says "pair with `docs +fetch` after a partial run to recover manually", but reconciling a 10-op batch mid-failure without knowing which blocks each op touched is hard. Including affected block ids per op would make recovery tractable.

**3. Validation can't catch in-batch staleness.** The `Validate` hook runs upfront, but it can't guard against later ops referencing block-ids that earlier ops invalidated (e.g., op[0] deletes a section that op[1]'s `selection-by-title` resolves into). For sequential semantics this may be acceptable, but it's worth calling out in the docs alongside the partial-failure note.

shortcuts/doc/docs_batch_update.go
```diff
@@ -0,0 +1,227 @@
...
+			out, callErr := common.CallMCPTool(runtime, "update-doc", args)
+			if callErr != nil {
+				results = append(results, batchUpdateResult{
+					Index: i, Mode: op.Mode, Success: false, Error: callErr.Error(),
+				})
+				if stopOnError {
+					fmt.Fprintf(runtime.IO().ErrOut,
+						"error: --operations[%d] failed; stopping (--on-error=stop); %d/%d applied before the failure\n",
+						i, successCount, len(ops))
+					runtime.Out(map[string]interface{}{
```
fangshuyu-768, line 162: Two error signals on `--on-error=stop`: `runtime.Out(...)` emits the partial result with `stopped_early=true` AND the function returns `callErr`, so callers see both a JSON success-shaped response and a non-zero exit code. Single-update doesn't have this ambiguity because it fails fast. Worth deciding which signal is canonical — scripts that key on exit code will see the JSON as residual stderr/stdout, and agents that key on JSON shape may not check exit code.

shortcuts/doc/docs_batch_update.go
```diff
@@ -0,0 +1,227 @@
...
+				continue
+			}
+			normalizeDocsUpdateResult(out, op.Markdown)
+			results = append(results, batchUpdateResult{
+				Index: i, Mode: op.Mode, Success: true, Result: out,
+			})
+			successCount++
+		}
+
+		runtime.Out(map[string]interface{}{
```
fangshuyu-768, line 180: The result envelope (`total / applied / stopped_early`) doesn't include a per-op block-id map. The PR description says 'pair with `docs +fetch` after a partial run to recover manually,' but reconciling a 10-op batch mid-failure without knowing which blocks each op touched is hard. Including affected block ids per op (e.g., new `applied: [{op_index, block_ids: [...]}, ...]`) would make recovery tractable for agents.

fangshuyu-768 (commented):
Three design notes inline; expands on the earlier summary review.

shortcuts/doc/docs_batch_update.go
```diff
@@ -0,0 +1,227 @@
...
+	Description: "Apply a sequence of update-doc operations to a single document. Sequential execution, not atomic — partial failure leaves the document in a partial-apply state. Pair with --on-error=stop + docs +fetch to recover.",
+	Risk:        "write",
+	Scopes:      []string{"docx:document:write_only", "docx:document:readonly"},
+	AuthTypes:   []string{"user", "bot"},
+	Flags: []common.Flag{
+		{Name: "doc", Desc: "document URL or token", Required: true},
+		{Name: "operations", Desc: "JSON array of operations (each entry: mode, markdown, selection_with_ellipsis, selection_by_title, new_title)", Required: true, Input: []string{common.File, common.Stdin}},
+		{Name: "on-error", Default: "stop", Desc: "behavior when a single operation fails: stop | continue"},
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
```
fangshuyu-768, line 81: The upfront Validate hook can't catch in-batch staleness: if op[0] deletes a section that op[1]'s `selection-by-title` resolves into, op[1] passes Validate but fails at MCP time. For sequential semantics this may be by design, but it's worth calling out alongside the partial-failure note in the docstring/skill md so agents don't assume Validate guards against everything pre-execution.

shortcuts/doc/docs_batch_update_test.go
```diff
@@ -0,0 +1,441 @@
...
+	if envelope.Data.Total != 2 {
+		t.Errorf("expected total=2, got %d", envelope.Data.Total)
+	}
+	if envelope.Data.Applied != 1 {
+		t.Errorf("expected applied=1 before stop, got %d", envelope.Data.Applied)
+	}
+	if !envelope.Data.StoppedEarly {
+		t.Errorf("expected stopped_early=true, got false")
+	}
+}
```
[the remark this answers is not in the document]
herbertliu: Fixed in 3d0302e — added `TestDocsBatchUpdateContinuesOnError` covering the success/fail/success-after-fail scenario. Uses an httpmock stub with HTTP 500 status for op[1] (CallMCPTool surfaces it as an error) sandwiched between two success stubs, then asserts: total=3, applied=2, stopped_early=false, results[0].Success=true, results[1] has non-empty Error, and results[2].Success=true (proving the loop continued past op[1]'s failure).

shortcuts/doc/docs_batch_update.go
herbertliu: Documented as a deliberate design decision in 3d0302e — added a 'Two design notes' block to the Go doc on `DocsBatchUpdate`:

> On --on-error=stop the shortcut emits TWO failure signals: the stdout JSON envelope (with stopped_early=true and the partial result list), AND a non-zero exit via the returned error. They are complementary: scripts that key on exit code see 'the batch did not complete cleanly'; callers parsing JSON see 'exactly which ops succeeded and how far we got'. Don't treat them as redundant.

Rationale: collapsing to one signal would cost callers in the other camp — exit-code-only loses the per-op detail; JSON-only would force every shell script to JQ-parse to know whether to abort. Keeping both feels right for a CLI surface.

shortcuts/doc/docs_batch_update.go
herbertliu: Acknowledged but deferring to a follow-up. The MCP `update-doc` v1 response doesn't carry block-id information for affected blocks (just `{success: true}` typically), so producing a per-op block-id map would require an extra fetch-doc per op (or a diff against a pre-batch snapshot) — that's a non-trivial scope.

The cleaner path is to layer this on top of v2 `docs +update`, where `--block-id` is already the input/output contract: a v2 batch wrapper can echo the block IDs each op touched without any extra API calls. Since v1 is hidden/deprecated, investing here would be writing throwaway code.

For now the recovery story is what's stated in the description: pair with `docs +fetch` after a partial run. Will revisit once v2 batch surface is decided.

shortcuts/doc/docs_batch_update.go
herbertliu: Documented as deliberate sequential semantics in 3d0302e — same Go doc block on `DocsBatchUpdate`:

> Validate runs once up front against the static op list, before any MCP write. It cannot foresee in-batch staleness — for example, if op[0] deletes a section that op[1]'s --selection-by-title resolves into, op[1] passes Validate but fails at execute time. This is by design (sequential semantics; each op sees the doc state produced by the previous op). If you need stricter atomicity, split the batch or fetch the doc between groups of related ops.

Adding mid-batch re-validation would couple Validate to MCP state and lose the up-front 'fail before any MCP call' guarantee, which is the primary value of the static check. Documenting the limit explicitly seems like the better trade.

fangshuyu-768:
Stepping back from the line-by-line review to raise a structural concern: **should `+batch-update` exist on the v1 surface at all?**

Codebase context:
- `versioned_help.go:46` prints `[deprecated] docs +update with v1 API is deprecated and will be removed in a future release` on every v1 invocation.
- `docs +create`, `+fetch`, and `+update` all have `*_v2.go` variants. This PR is the only doc shortcut that lands purely on v1 with no v2 path.

What pushes me to raise this now (rather than as a v2 follow-up) is that the rationale used in the per-op `block_id` thread above applies to the whole PR by identical logic:

> "v1 is hidden/deprecated, investing here would be writing throwaway code."

That argument is correct, and it's self-defeating. Adding one field to v1 was rejected as not worth it, but landing a whole new shortcut on v1 is several times the surface area — tests, docstring, skill teaching, plus the migration cost when v1 is removed. If "one field of v1 investment is throwaway," "one shortcut of v1 investment" is the same conclusion at higher cost. Either both are worth doing or neither is.

The two shapes I think justify shipping:

1. **v2-only.** Drop the MCP `update-doc` call. Each op becomes `{command, block_id, content, pattern, src_block_ids, revision_id}` against `PUT /open-apis/docs_ai/v1/documents/<token>`. Side benefit: the per-op `block_id` map from the earlier thread falls out of the v2 response contract for free, closing the partial-recovery gap with zero extra round-trips.
2. **Dual-track from day one,** mirroring `docs +update`'s `--api-version v1|v2`. v1 path keeps current behavior with the same deprecation warning; v2 path uses the v2 op schema. More work than (1), but matches the existing pattern in this package.

What I'd push back on is the current shape: v1-only with v2 deferred. That commits the team to maintaining a v1-pinned shortcut on a sunset timeline, asks agents to adopt a surface that will need to be rewritten or removed, and — by the same logic used to defer block-id reporting — burns review/test budget on code that's expected to be thrown away.

**The one thing that flips this:** an internal v1 sunset timeline substantially longer than the rest of the v2 migration. If v1 has guaranteed remaining life that's long enough to amortize this work, v1-only is defensible as a near-term ergonomic win. Without that, the default "deprecated = no new investment" convention should apply.

Happy to be talked out of this if I'm misreading the timeline or the migration path. Otherwise I'd rather we land (1) or (2) than ship v1-only and chase a v2 follow-up.

PR #625, opened 2026-04-23
feat(doc): add --omit-title flag to docs +fetch for round-trip safety

herbertliu:
## Summary

Addresses Case 13 in the lark-cli pitfall list: \`docs +fetch --format=pretty\` prepends \`# <title>\` for human readability, but Lark stores the document title as an independent field and \`docs +update --mode=overwrite\` only rewrites the body. Piping pretty output through overwrite therefore accumulates a duplicate H1 block on every round-trip — observed in practice as 6+ duplicate H1s on the PR-tracker doc before anyone noticed.

Add \`--omit-title\` (off by default) so round-trip callers get a clean body.

## Changes

- **\`shortcuts/doc/docs_fetch.go\`**: new \`--omit-title\` bool flag; extract \`renderFetchPretty\` from the Execute closure for testability.
- **\`shortcuts/doc/docs_fetch_test.go\`** (new): 6-case table-driven coverage + a dedicated round-trip-safety assertion.
- **\`skills/lark-doc/references/lark-doc-fetch.md\`**: document the flag and the round-trip use case.

## Scope notes

- Default behavior unchanged — human readers still see the H1 title up top.
- Only \`--format=pretty\` is affected; \`json\` / \`ndjson\` / \`csv\` / \`table\` never carried the synthetic prefix.
- This is the Option 2 resolution from the Case 13 brief (additive flag). Option 1 (change default) was rejected on breaking-change risk; Option 3 (auto-strip on overwrite) needs either an extra API call to learn the title or a fragile heuristic.

## Test Plan

- [x] \`go test ./shortcuts/doc/...\` passes
- [x] \`go vet ./...\` clean
- [x] \`gofmt -l\` clean
- [ ] Manual smoke: run a fetch → overwrite cycle with and without the flag and confirm no H1 accumulates in the latter

## Summary by CodeRabbit

* **New Features**
  * Added `--omit-title` flag to `docs +fetch` command for `--format=pretty` output, allowing users to exclude the initial title line from rendered markdown.

* **Documentation**
  * Updated documentation with new `--omit-title` flag examples and parameter descriptions.

### DIFF
--- a/shortcuts/doc/docs_fetch.go
+++ b/shortcuts/doc/docs_fetch.go
@@ -24,6 +24,7 @@ var DocsFetch = common.Shortcut{
 		{Name: "doc", Desc: "document URL or token", Required: true},
 		{Name: "offset", Desc: "pagination offset"},
 		{Name: "limit", Desc: "pagination limit"},
+		{Name: "omit-title", Type: "bool", Desc: "in --format=pretty output, skip the leading '# <title>' line. Lark stores document title as an independent field, so including it when piping fetch output into `docs +update --mode=overwrite` accumulates duplicate H1 blocks on every round-trip."},
 	},
 	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
 		args := map[string]interface{}{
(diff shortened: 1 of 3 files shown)


fangshuyu-768 (commented):
Functional concern on flag scoping.

**`--omit-title` is exposed for both v1 and v2 paths but only consumed by v1 (`shortcuts/doc/docs_fetch.go`, lines ~27–28).** Calling `--api-version v2 --format pretty --omit-title` silently drops the flag because `executeFetchV2` never references it. Suggest either moving the flag into a v1-scoped helper alongside `v1FetchFlags()` (with `Hidden: true`) or rejecting the combination in the `Validate` hook when `useV2Fetch(runtime)` returns true. As-is, agents that compose `--omit-title` with v2 get unexpected behavior with no error.

shortcuts/doc/docs_fetch.go
```diff
@@ -24,6 +24,7 @@ var DocsFetch = common.Shortcut{
 		{Name: "doc", Desc: "document URL or token", Required: true},
 		{Name: "offset", Desc: "pagination offset"},
 		{Name: "limit", Desc: "pagination limit"},
+		{Name: "omit-title", Type: "bool", Desc: "in --format=pretty output, skip the leading '# <title>' line. Lark stores document title as an independent field, so including it when piping fetch output into `docs +update --mode=overwrite` accumulates duplicate H1 blocks on every round-trip."},
```
fangshuyu-768, line 27: `--omit-title` is exposed at the top-level Flags array (shared by v1 and v2 paths) but only consumed inside `executeFetchV1` via `renderFetchPretty` at line 75. Calling `--api-version v2 --format pretty --omit-title` silently drops the flag, which contradicts the user's expectation that documented flags work on every supported path. Suggest either moving this flag into a v1-scoped helper (with `Hidden: true` on v2) or rejecting the combination in the `Validate` hook when `useV2Fetch(runtime)` returns true. As-is, agents that compose `--omit-title` with v2 get unexpected output with no error.

fangshuyu-768 (commented):
One inline note; expands on the earlier summary review.

herbertliu:
Closing this PR — superseded by the docs v2 API refactor (#638).

The Case 13 root cause was that `docs +fetch --format pretty` prepended a `# <title>` line, and feeding that back into `docs +update --mode overwrite` accumulated duplicate H1 blocks each round-trip.

v2's `executeFetchV2` writes only the document content (XML or markdown body) without synthesizing a title prefix, so the round-trip naturally round-trips. The `--omit-title` flag this PR adds is no longer needed; v1 will be removed before any new round-trip workflows would be designed against it.

PR #646, opened 2026-04-23
feat(mail): support calendar events in emails

chanthuang:
## Summary

Add support for embedding calendar invitations in outgoing emails and parsing calendar event data from received emails.

### Sending calendar invitations

Use `--event-*` flags with any compose command to attach an iCalendar invitation (`text/calendar`). Recipients receive a standard calendar invite and can accept or decline the event.

```bash
lark-cli mail +send \
  --to alice@example.com --cc bob@example.com \
  --subject "Team sync" \
  --body "<p>Please join our weekly sync.</p>" \
  --event-summary "Team sync" \
  --event-start "2026-05-10T14:00+08:00" \
  --event-end "2026-05-10T15:00+08:00" \
  --event-location "Meeting room 5F" \
  --confirm-send
```

**New flags** (available on `+send`, `+reply`, `+reply-all`, `+forward`, `+draft-create`):

| Flag | Required | Description |
|------|----------|-------------|
| `--event-summary <text>` | — | Event title. Enables calendar invitation mode when set. Requires `--event-start` and `--event-end`. |
| `--event-start <time>` | When `--event-summary` is set | Event start time (ISO 8601, e.g. `2026-05-10T14:00+08:00`) |
| `--event-end <time>` | When `--event-summary` is set | Event end time (ISO 8601) |
| `--event-location <text>` | No | Event location |

**Constraints:**
- `--event-summary`, `--event-start`, and `--event-end` must all be provided together
- `--event-*` flags are mutually exclusive with `--send-time` (scheduled delivery)
- `To` and `Cc` recipients are automatically added as attendees; `Bcc` recipients are not

### Editing calendar events in drafts

`+draft-edit` gains two new patch operations via `--patch-file`:

- `set_calendar` — attach or replace the calendar invitation in a draft
- `remove_calendar` — remove the calendar invitation from a draft

### Reading calendar invitations

When a received email contains a calendar invitation, `+message` output now includes a `calendar_event` field:

```json
{
  "calendar_event": {
    "method": "REQUEST",
    "uid": "abc123",
    "summary": "Team sync",
    "start": "2026-05-10T06:00:00Z",
    "end": "2026-05-10T07:00:00Z",
    "location": "Meeting room 5F",
    "organizer": "sender@example.com",
    "attendees": ["alice@example.com", "bob@example.com"]
  }
}
```

## Test plan

- [x] `+send` / `+draft-create` / `+reply` / `+reply-all` / `+forward` with `--event-*` flags produce valid RFC 5545 iCalendar output
- [x] `To` and `Cc` recipients appear as `ATTENDEE` entries; `Bcc` recipients are excluded
- [x] `--event-*` and `--send-time` cannot be used together (validated at input)
- [x] `set_calendar` and `remove_calendar` patch ops work correctly in `+draft-edit`
- [x] `+message` correctly parses and surfaces `calendar_event` from received invitation emails
- [x] CN parameter quoting handles special characters per RFC 5545
- [x] Line folding handles both CRLF and LF line endings

## Summary by CodeRabbit

* **New Features**
  * Calendar invitation support in mail shortcuts (send, reply, reply-all, forward, draft-create, draft-edit) via --event-summary/--event-start/--event-end/--event-location (ISO‑8601). Produces a text/calendar part (placed under multipart/alternative), supports set/remove event edits, enforces start+end and end > start, and rejects use with scheduled send or Bcc.

* **New Features**
  * ICS generation and parsing with calendar_event exposure in message output.

* **Documentation**
  * CLI and mail docs updated with flags, examples, constraints, and output fields.

* **Tests**
  * Extensive unit and integration coverage for validation, MIME handling, ICS build/parse, and end-to-end flows.

### DIFF
--- a/shortcuts/mail/draft/patch_calendar.go
+++ b/shortcuts/mail/draft/patch_calendar.go
@@ -122,8 +122,7 @@ func FindPartByMediaType(root *Part, mediaType string) *Part {
 }
 
 // findAllPartsByMediaType walks the MIME tree and returns every part with
-// the given media type. Used by setCalendarPart to update all copies of the
-// ICS (e.g. one inside multipart/alternative and one as an inline attachment).
+// the given media type. Used in tests to assert tree contents.
 func findAllPartsByMediaType(root *Part, mediaType string) []*Part {
 	if root == nil {
 		return nil
--- a/shortcuts/mail/draft/patch_calendar.go
+++ b/shortcuts/mail/draft/patch_calendar.go
@@ -1,0 +1,179 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package draft
+
+import (
+	"fmt"
+	"strings"
+)
+
+const calendarMediaType = "text/calendar"
+
+// applyCalendarSet installs or replaces the text/calendar MIME part in the
+// snapshot. The caller is expected to have pre-built icsData using the
+// snapshot's From/To/Cc addresses.
+func applyCalendarSet(snapshot *DraftSnapshot, icsData []byte) error {
+	if len(icsData) == 0 {
+		return fmt.Errorf("set_calendar: ICS data is empty (shortcut layer must pre-build it)")
+	}
+	setCalendarPart(snapshot, icsData)
+	return nil
+}
+
+// applyCalendarRemove strips the text/calendar part from the snapshot.
+// No-op if no calendar part exists.
+func applyCalendarRemove(snapshot *DraftSnapshot) error {
+	removeCalendarPart(snapshot)
+	return nil
+}
+
+// setCalendarPart places exactly one text/calendar part inside
+// multipart/alternative, matching the Feishu client behavior. Any existing
+// text/calendar parts elsewhere in the tree are removed first.
+func setCalendarPart(snapshot *DraftSnapshot, icsData []byte) {
+	newPart := &Part{
+		MediaType:   calendarMediaType,
+		MediaParams: map[string]string{"charset": "UTF-8"},
+		Body:        icsData,
+		Dirty:       true,
+	}
+
+	if snapshot.Body == nil {
+		snapshot.Body = newPart
+		return
+	}
+
+	// Remove all existing text/calendar parts from everywhere in the tree.
+	if strings.EqualFold(snapshot.Body.MediaType, calendarMediaType) {
+		snapshot.Body = newPart
+		return
+	}
+	removeAllPartsByMediaType(snapshot.Body, calendarMediaType)
+
+	// Place inside the existing multipart/alternative.
+	if alt := FindPartByMediaType(snapshot.Body, "multipart/alternative"); alt != nil {
+		alt.Children = append(alt.Children, newPart)
+		alt.Dirty = true
+		return
+	}
+
+	// No multipart/alternative exists. If the body is a single leaf,
+	// wrap it in multipart/alternative together with the calendar.
+	if !snapshot.Body.IsMultipart() {
+		original := *snapshot.Body
+		snapshot.Body.MediaType = "multipart/alternative"
+		snapshot.Body.MediaParams = nil
+		snapshot.Body.Body = nil
+		snapshot.Body.TransferEncoding = ""
+		snapshot.Body.RawEntity = nil
+		snapshot.Body.Children = []*Part{&original, newPart}
+		snapshot.Body.Dirty = true
+		return
+	}
+
+	// Multipart body without an alternative sub-part (e.g. multipart/mixed
+	// with a text/html child). Find the first text/* child and wrap it in
+	// a new multipart/alternative that also contains the calendar.
+	for i, child := range snapshot.Body.Children {
+		if child != nil && strings.HasPrefix(strings.ToLower(child.MediaType), "text/") {
+			alt := &Part{
+				MediaType: "multipart/alternative",
+				Children:  []*Part{child, newPart},
+				Dirty:     true,
+			}
+			snapshot.Body.Children[i] = alt
+			snapshot.Body.Dirty = true
+			return
+		}
+	}
+
+	// Fallback: append to the root multipart container.
+	snapshot.Body.Children = append(snapshot.Body.Children, newPart)
+	snapshot.Body.Dirty = true
+}
+
+func removeCalendarPart(snapshot *DraftSnapshot) {
+	if snapshot.Body == nil {
+		return
+	}
+	if strings.EqualFold(snapshot.Body.MediaType, calendarMediaType) {
+		snapshot.Body = nil
+		return
+	}
+	removeAllPartsByMediaType(snapshot.Body, calendarMediaType)
+}
+
+// FindPartByMediaType walks the MIME tree and returns the first part with
+// the given media type, or nil when not found.
+func FindPartByMediaType(root *Part, mediaType string) *Part {
+	if root == nil {
+		return nil
+	}
+	if strings.EqualFold(root.MediaType, mediaType) {
+		return root
+	}
+	for _, child := range root.Children {
+		if found := FindPartByMediaType(child, mediaType); found != nil {
+			return found
+		}
+	}
+	return nil
+}
+
+// findAllPartsByMediaType walks the MIME tree and returns every part with
+// the given media type. Used by setCalendarPart to update all copies of the
+// ICS (e.g. one inside multipart/alternative and one as an inline attachment).
+func findAllPartsByMediaType(root *Part, mediaType string) []*Part {
+	if root == nil {
+		return nil
+	}
+	var result []*Part
+	if strings.EqualFold(root.MediaType, mediaType) {
+		result = append(result, root)
+	}
+	for _, child := range root.Children {
+		result = append(result, findAllPartsByMediaType(child, mediaType)...)
+	}
+	return result
+}
+
+// removePartByMediaType removes the first part with the given media type from
+// the MIME tree. The parent is marked dirty when a removal happens.
+func removePartByMediaType(root *Part, mediaType string) {
+	if root == nil {
+		return
+	}
+	for i, child := range root.Children {
+		if child != nil && strings.EqualFold(child.MediaType, mediaType) {
+			root.Children = append(root.Children[:i], root.Children[i+1:]...)
+			root.Dirty = true
+			return
+		}
+		removePartByMediaType(child, mediaType)
+	}
+}
+
+// removeAllPartsByMediaType removes every part with the given media type from
+// the MIME tree, at all nesting levels.
+func removeAllPartsByMediaType(root *Part, mediaType string) {
+	if root == nil {
+		return
+	}
+	var kept []*Part
+	removed := false
+	for _, child := range root.Children {
+		if child != nil && strings.EqualFold(child.MediaType, mediaType) {
+			removed = true
+			continue
+		}
+		kept = append(kept, child)
+	}
+	if removed {
+		root.Children = kept
+		root.Dirty = true
+	}
+	for _, child := range root.Children {
+		removeAllPartsByMediaType(child, mediaType)
+	}
+}
--- a/shortcuts/mail/draft/patch_calendar.go
+++ b/shortcuts/mail/draft/patch_calendar.go
@@ -62,11 +62,21 @@ func setCalendarPart(snapshot *DraftSnapshot, icsData []byte) {
 	// wrap it in multipart/alternative together with the calendar.
 	if !snapshot.Body.IsMultipart() {
 		original := *snapshot.Body
+		// Reset all header-carrying fields so the serializer constructs a fresh
+		// Content-Type from MediaType instead of reusing the stale leaf headers.
+		snapshot.Body.Headers = nil
 		snapshot.Body.MediaType = "multipart/alternative"
 		snapshot.Body.MediaParams = nil
+		snapshot.Body.ContentDisposition = ""
+		snapshot.Body.ContentDispositionArg = nil
+		snapshot.Body.ContentID = ""
+		snapshot.Body.PartID = ""
 		snapshot.Body.Body = nil
 		snapshot.Body.TransferEncoding = ""
 		snapshot.Body.RawEntity = nil
+		snapshot.Body.Preamble = nil
+		snapshot.Body.Epilogue = nil
+		snapshot.Body.EncodingProblem = false
 		snapshot.Body.Children = []*Part{&original, newPart}
 		snapshot.Body.Dirty = true
 		return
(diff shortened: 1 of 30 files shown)


shortcuts/mail/draft/patch_calendar.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+	snapshot.Body.Dirty = true
+}
+
+func removeCalendarPart(snapshot *DraftSnapshot) {
+	if snapshot.Body == nil {
+		return
+	}
+	removePartByMediaType(snapshot.Body, calendarMediaType)
```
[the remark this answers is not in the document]
chanthuang: Fixed: 未找到命令：removeCalendarPart already handles the standalone case — lines 100–102 set 未找到命令：snapshot.Body when the root body itself is . The concern was valid for a previous revision but the current code is correct.

shortcuts/mail/ics/parser.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// splitProperty splits "NAME;PARAMS:VALUE" into (name-with-params, value).
+// The first unquoted colon is the separator.
+func splitProperty(line string) (string, string) {
+	idx := strings.Index(line, ":")
+	if idx < 0 {
+		return line, ""
+	}
+	return line[:idx], line[idx+1:]
```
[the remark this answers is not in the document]
chanthuang: Fixed: the current `splitProperty` implementation (lines 123–136) already handles this correctly — it scans character-by-character tracking a `inQuote` flag and only splits on colons outside double-quoted strings. The concern described an older `strings.Index` version; the code was already updated before this review.

shortcuts/mail/ics/ics_test.go
```diff
@@ -0,0 +1,499 @@
...
+			break
+		}
+	}
+	if strings.Contains(uid, "@") {
+		t.Errorf("auto-generated UID should be pure UUID (no @host suffix), got %q", uid)
+	}
+	// UUID v4 has 36 chars (8-4-4-4-12 plus 4 dashes).
+	if len(uid) != 36 {
+		t.Errorf("auto-generated UID should be 36-char UUID, got %d chars: %q", len(uid), uid)
+	}
```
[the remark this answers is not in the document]
chanthuang: Acknowledged. The current test rejects `@` and enforces 36-char length which is sufficient to catch our UUID generator being broken. A full v4 regex check is a nice-to-have; will track for a follow-up test hardening PR to keep this PR focused.

shortcuts/mail/draft/patch_calendar.go
```diff
@@ -0,0 +1,117 @@
...
+		&original,
+		{
+			MediaType:   calendarMediaType,
+			MediaParams: map[string]string{"charset": "UTF-8"},
+			Body:        icsData,
+			Dirty:       true,
+		},
+	}
+	snapshot.Body.Dirty = true
+}
```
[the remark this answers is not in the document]
chanthuang: Fixed in commit 350a8ba. The shallow-copy path now clears `Headers`, `ContentDisposition`, `ContentDispositionArg`, `ContentID`, `PartID`, `Preamble`, `Epilogue`, and `EncodingProblem` before repurposing the part as the `multipart/alternative` wrapper, so `orderedPartHeaders` will construct a fresh Content-Type from `MediaType` instead of reusing the stale leaf headers. Tests pass.

shortcuts/mail/draft/patch_calendar.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	for i, child := range root.Children {
+		if child != nil && strings.EqualFold(child.MediaType, mediaType) {
+			root.Children = append(root.Children[:i], root.Children[i+1:]...)
+			root.Dirty = true
+			return
+		}
+		removePartByMediaType(child, mediaType)
+	}
+}
```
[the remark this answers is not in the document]
chanthuang: Fixed in commit 13cc6d0: updated the doc comment on `findAllPartsByMediaType` to clearly state it is used in tests rather than claiming production usage by `setCalendarPart`. `removePartByMediaType` likewise is test-only; both functions remain in the package file (accessible to package-level tests without export) rather than moved to `_test.go`, which is consistent with the other unexported helpers in this file.

shortcuts/mail/mail_draft_edit.go
```diff
@@ -132,6 +140,29 @@ var MailDraftEdit = common.Shortcut{
...
+				if _, _, err := parseEventTimeRange(patch.Ops[i].EventStart, patch.Ops[i].EventEnd); err != nil {
+					return output.ErrValidation("set_calendar: %v", err)
+				}
+				calData := buildCalendarBodyFromArgs(
+					patch.Ops[i].EventSummary,
+					patch.Ops[i].EventStart,
+					patch.Ops[i].EventEnd,
+					patch.Ops[i].EventLocation,
+					draftFromEmail,
+					joinAddresses(snapshot.To),
```
infeng: **Issue:** `set_calendar` builds the ICS before the patch ops are applied, so it always reads attendees from `snapshot.To/Cc` instead of the post-edit recipient lists. A single `+draft-edit --set-to/--set-cc ... --set-event-* ...` invocation can therefore save headers addressed to the new people while the embedded VCALENDAR still lists the old attendees.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Derive attendees from the pending patch (for example by applying recipient ops to a copy of the snapshot first, or by computing the effective To/Cc after merging `set_recipients` ops) before calling `buildCalendarBodyFromArgs`. A regression test should cover editing recipients and calendar in the same command.

shortcuts/mail/draft/patch_calendar.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+// setCalendarPart places exactly one text/calendar part inside
+// multipart/alternative, matching the Feishu client behavior. Any existing
+// text/calendar parts elsewhere in the tree are removed first.
+func setCalendarPart(snapshot *DraftSnapshot, icsData []byte) {
+	newPart := &Part{
+		MediaType:   calendarMediaType,
+		MediaParams: map[string]string{"charset": "UTF-8"},
```
infeng: **Issue:** `+draft-edit --set-event-*` inserts a new `text/calendar` part with only `charset=UTF-8`. Unlike the compose path, the serialized MIME here never adds `method=REQUEST`, so draft-edit-generated invites are not RFC 6047-compliant and may not render RSVP UI in Outlook/Gmail.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Include the same calendar Content-Type params used by `emlbuilder` (`method=REQUEST`, and any required filename/disposition metadata) when creating or replacing the calendar part, and add a serialization-level test that asserts the updated draft contains `Content-Type: text/calendar; method=REQUEST; ...`.

shortcuts/mail/mail_forward.go
```diff
@@ -214,6 +221,11 @@ var MailForward = common.Shortcut{
 			bld = bld.TextBody([]byte(composedTextBody))
 		}
 		bld = applyPriority(bld, priority)
+		if calData := buildCalendarBody(runtime, senderEmail, to, ccFlag); calData != nil {
+			bld = bld.CalendarBody(calData)
+		} else if len(sourceMsg.OriginalCalendarICS) > 0 {
```
infeng, line 307: **Issue:** This fallback forwards `body_calendar` verbatim when no `--event-*` flags are provided. That raw ICS still carries the original organizer and attendee list, so forwarding a meeting invite to a different audience can emit an interactive calendar part whose RSVP metadata no longer matches the new `To`/`Cc` headers.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Either drop `OriginalCalendarICS` on plain forwards, or regenerate the VCALENDAR from the forwarded recipient set before attaching it. A regression test should cover forwarding an invite without `--event-*` to different recipients.

shortcuts/mail/mail_draft_edit.go
chanthuang: Fixed in d7a82b3. When `set_calendar` is processed, we now scan `patch.Ops` for any preceding `set_recipients` ops and use their addresses as `toAddrs`/`ccAddrs` instead of `snapshot.To/Cc`. This ensures the ICS ATTENDEE list reflects the final post-edit recipients.

shortcuts/mail/draft/patch_calendar.go
chanthuang: Fixed in 83b621a. `setCalendarPart` now sets `MediaParams: map[string]string{"charset": "UTF-8", "method": "REQUEST"}`, matching the `emlbuilder` path. Added a test assertion to cover this.

shortcuts/mail/mail_forward.go
chanthuang: This is intentional — Strategy A in the design spec (§11.2 of `mail-calendar-feature.md`). We pass through the original ICS bytes verbatim without modifying ATTENDEE or adding `X-LARK-MAIL-DRAFT`, which matches the Rust SDK client behavior. Because the forwarded ICS lacks the `X-LARK-MAIL-DRAFT` marker, `+draft-edit` treats it as read-only — users cannot accidentally edit the attendee list to mismatch the MIME headers. The forwarded invite is intended as informational context for the new recipients, not as a new RSVP event targeting them.

chanthuang:
@coderabbitai review

shortcuts/mail/mail_forward.go
infeng: Re-checked the current head `d7a82b3`. Given the explicit passthrough strategy described by the author and the fact that forwarded ICS stays read-only (no `X-LARK-MAIL-DRAFT`), the current behavior is intentional rather than a missed verification. I am treating this as a product/design choice, not a remaining implementation bug.

shortcuts/mail/draft/patch_calendar.go
infeng: Verified as fixed on the current head `d7a82b3`: the draft patch path now carries `method=REQUEST` on the generated `text/calendar` part, so this issue is addressed.

shortcuts/mail/mail_draft_edit.go
infeng: Re-checked the current head `d7a82b3`. The direct `--set-to/--set-cc` path is fixed, but verification still fails for the patch-file recipient ops: `set_calendar` now folds in `set_recipients`, yet it still ignores `add_recipient` / `remove_recipient`, which remain documented and validated patch operations. A patch file that increments recipients and adds `set_calendar` in the same edit can still serialize stale `ATTENDEE` metadata.

chanthuang: Fixed in ed0cbff. Extracted `effectiveRecipients()` that replays all three recipient op types — `set_recipients`, `add_recipient`, `remove_recipient` — against a copy of `snapshot.To/Cc`, so both direct flags and patch-file recipient ops are reflected in the ICS ATTENDEE list before `Apply()` runs.

infeng: Verified as fixed on the current head `ed0cbff`: `effectiveRecipients()` now replays `set_recipients`, `add_recipient`, and `remove_recipient`, so the draft-edit calendar ATTENDEE list follows both direct flags and patch-file recipient edits.

infeng (dismissed):
LGTM. All prior review threads resolved. Baseline: head `ed0cbff478f03738e61e61523640eb36ecee7fd8`.

shortcuts/mail/emlbuilder/builder.go
```diff
@@ -992,6 +996,17 @@ func writeSingleBodyPartHeaders(buf *bytes.Buffer, ct string, body []byte) {
...
+// writeCalendarPart writes the text/calendar MIME part with the RFC 6047
+// method=REQUEST parameter. The part lives inside multipart/alternative
+// as an alternative representation of the message body.
+func writeCalendarPart(buf *bytes.Buffer, body []byte) {
+	cte := selectCTE(body)
+	fmt.Fprintf(buf, "Content-Type: text/calendar; method=REQUEST; charset=UTF-8\n")
+	fmt.Fprintf(buf, "Content-Transfer-Encoding: %s\n\n", cte)
+	writeFoldedBody(buf, encodeBodyContent(body, cte), lineWidthForCTE(cte))
+	buf.WriteByte('\n')
+}
```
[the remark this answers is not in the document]
chanthuang: Fixed in 4433baf. Added `extractICSMethod()` that scans the ICS body for `METHOD:` and falls back to `REQUEST` when absent. `writeCalendarPart` now uses the extracted method in the Content-Type header, so passthrough ICS with `METHOD:CANCEL` or `METHOD:REPLY` produce correct MIME metadata. Added regression tests for REQUEST/CANCEL/REPLY and no-method cases.

shortcuts/mail/helpers.go
```diff
@@ -1458,6 +1483,29 @@ func buildMessageForCompose(msg map[string]interface{}, urlMap map[string]string
...
+					Location:  parsed.Location,
+					Organizer: parsed.Organizer,
+					Attendees: parsed.Attendees,
+				}
+				if !parsed.Start.IsZero() {
+					ce.Start = parsed.Start.Format(time.RFC3339)
+				}
+				if !parsed.End.IsZero() {
+					ce.End = parsed.End.Format(time.RFC3339)
+				}
```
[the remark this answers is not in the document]
chanthuang: Fixed in c444c00. Added `.UTC()` before `.Format(time.RFC3339)` for both `Start` and `End`, so `TZID=Asia/Shanghai` offsets are normalized to `Z` in the output.

shortcuts/mail/ics/parser.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+		case name == "ORGANIZER" || strings.HasPrefix(name, "ORGANIZER;"):
+			if email := extractMailto(value); email != "" {
+				event.Organizer = email
+			}
+		case name == "ATTENDEE" || strings.HasPrefix(name, "ATTENDEE;"):
+			if email := extractMailto(value); email != "" {
+				event.Attendees = append(event.Attendees, email)
+			}
+		}
```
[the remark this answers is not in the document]
chanthuang: Fixed in 323df2e. Introduced `propUpper := strings.ToUpper(name)` before the switch and updated all property comparisons — UID, SUMMARY, LOCATION, DTSTART, DTEND, RECURRENCE-ID, ORGANIZER, ATTENDEE — to use `propUpper == "X" || strings.HasPrefix(propUpper, "X;")`, matching RFC 5545 §3.1 case-insensitivity and handling parameterized forms consistently across all properties.

shortcuts/mail/ics/ics_test.go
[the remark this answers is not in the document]
infeng: Re-checked the current head `323df2e`. This one is test hardening rather than a product bug: the implementation already sources auto-generated UIDs from `github.com/google/uuid` via `uuid.New()`, so the runtime behavior is already constrained to standard UUID generation. I am treating the current test as sufficient for this change and not requiring an extra parser/version assertion here.

infeng (dismissed):

haidaodashushu (approved):

PR #651, opened 2026-04-24
feat(auth): mint tenant access token from env app credentials

fengliu222:
## Summary

Enable headless callers — remote AI agents, CI jobs, and ephemeral containers — to authenticate with bot identity using only environment variables, eliminating the need for `config init`, `auth login`, or an OS keychain.

When `LARKSUITE_CLI_APP_ID` and `LARKSUITE_CLI_APP_SECRET` are set, the env credential provider now exchanges them for a tenant access token on demand and caches the result in-process until it expires. Existing env-var combinations (pre-issued UAT / TAT, strict-mode overrides, etc.) continue to work unchanged.

Motivation: AGENTS.md notes the CLI's primary consumers are AI agents, and many of those agents run in environments without a browser or a writable keychain. Previously, setting only `APP_ID + APP_SECRET` left the env provider unable to produce a TAT, surfacing `TokenUnavailableError` instead of falling through to the app-secret exchange flow that the default provider already implements.

## Changes

- **auth**: new shared `auth.FetchTenantAccessToken` helper that performs the `/open-apis/auth/v3/tenant_access_token/internal` exchange. Both the default and env credential providers reuse it.
- **auth**: env provider mints and caches a TAT from `APP_ID + APP_SECRET` when `LARKSUITE_CLI_TENANT_ACCESS_TOKEN` is not set. Cache honours the server-reported expiry with a 30-second safety margin; app_id and brand are keyed into the cache; explicit TAT env vars still take priority.
- **auth**: `SupportedIdentities` and `DefaultAs` now include bot when `APP_SECRET` is available, so commands work without an explicit `--as bot` flag when only app credentials are provided. Strict-mode overrides continue to win.
- **docs**: new "Environment-based credentials (agents & CI)" section in `README.md` / `README.zh.md`; headless pointer added to the Quick Start (AI Agent) section; `lark-shared` skill documents the env path and updated identity table; `[Unreleased]` entry in `CHANGELOG.md`.

## Test Plan

- [x] Unit tests pass — `go test -race -count=1 ./extension/credential/env/... ./internal/auth/... ./internal/credential/...`
  - New cases cover mint, cache hit, env-TAT precedence, post-expiry refetch, app_id mismatch block, and missing-secret fallthrough.
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go mod tidy` produces no drift
- [x] `golangci-lint run --new-from-rev=origin/main` reports 0 issues
- [x] Manual local verification against a live Feishu app:
  - `lark-cli auth status` reports `identity: bot`, `defaultAs: bot` with only env vars set.
  - `lark-cli api GET /open-apis/bot/v3/info` returns `code: 0` end-to-end (TAT exchange + SDK transport + API round-trip).
  - `lark-cli docs +create --as bot --title ... --markdown ...` successfully creates a doc.
  - Setting a bogus `LARKSUITE_CLI_TENANT_ACCESS_TOKEN` forces the server to reject the request, confirming explicit env TAT takes precedence over minting.

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Env-var authentication for headless/CI use
  * Automatic bot identity detection from app credentials
  * Lazy tenant access token minting with in-process caching

* **Documentation**
  * Added headless/CI guides to README and README.zh
  * Documented env var names, precedence, and identity selection
  * Updated skill docs for non-interactive bot auth workflow

* **Refactor**
  * Shared tenant-access-token exchange logic extracted

* **Tests**
  * Added tests for token exchange, caching and identity inference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Features
+
+- **auth**: env credential provider now exchanges `LARKSUITE_CLI_APP_ID` + `LARKSUITE_CLI_APP_SECRET` for a tenant access token on demand, enabling zero-config bot auth for remote AI agents, CI jobs, and ephemeral containers with no keychain or `auth login` step. The token is cached per-process until it expires.
+- **auth**: infer `SupportsBot` and `DefaultAs=bot` from `LARKSUITE_CLI_APP_SECRET` so commands work without an explicit `--as bot` flag when only app credentials are provided.
+
+### Refactor
+
+- **auth**: extract the tenant_access_token exchange into a shared `auth.FetchTenantAccessToken` helper, reused by the default and env credential providers.
+
+### Documentation
+
+- README / README.zh: document environment-based credential resolution under the Authentication section, with a pointer from the Quick Start (AI Agent) for headless setups.
+- `lark-shared` skill: note env-based bot auth as an alternative to `config init` + `auth login` for headless agents.
+
 ## [v1.0.17] - 2026-04-22
 
 ### Features
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ export LARKSUITE_CLI_APP_SECRET=yyy
 lark-cli docs +create --as bot --title "Report" --markdown "# Hello"
 ```
 
-Environment credentials take priority over any `config init` profile on the same machine, so the same binary works interactively on a workstation and headlessly in CI without extra flags. Because user identity requires an interactive authorization step, env-only setups can call bot-capable commands only; mix in `LARKSUITE_CLI_USER_ACCESS_TOKEN` to enable `--as user`.
+Environment credentials take priority over any `config init` profile on the same machine, so the same binary works interactively on a workstation and headlessly in CI without extra flags. User identity normally requires an interactive `auth login`, so `APP_ID + APP_SECRET` alone grants bot identity. To run `--as user` purely from environment, issue a user access token ahead of time and export it as `LARKSUITE_CLI_USER_ACCESS_TOKEN` alongside `LARKSUITE_CLI_APP_ID`.
 
 ## Three-Layer Command System
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ lark-cli calendar +agenda
 ## Quick Start (AI Agent)
 
 > The following steps are for AI Agents. Some steps require the user to complete actions in a browser.
+>
+> **Headless environments (remote agents, CI jobs, containers without a browser):** skip Steps 2–3 and inject `LARKSUITE_CLI_APP_ID` + `LARKSUITE_CLI_APP_SECRET` as environment variables instead. See [Environment-based credentials](#environment-based-credentials-agents--ci) for the full variable list.
 
 **Step 1 — Install**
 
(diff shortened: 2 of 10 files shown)


fengliu222:
Pushed `d50b6fe` addressing the review feedback:

**Applied (5/6):**
- **env lock + hung exchange** (Copilot @ env.go:205, CodeRabbit @ env.go:189-217) — the lock is still held across the exchange to serialise concurrent callers (prevents thundering herd), but the default `http.Client` fallback now carries a 30s timeout so a hung TCP connection can't wedge the mutex indefinitely. Callers needing a different budget can still pass `Provider.HTTPClient` or a context with a deadline.
- **Package-level test seam** (CodeRabbit @ env.go:30-39) — moved to an unexported `Provider.fetchTAT` field. Tests now stub per-instance, so parallel tests can't interfere with each other. The package-level `tatFetcher` var is gone.
- **`openBaseURL` validation** (Copilot @ tat_fetch.go:38) — explicit empty + scheme check returns a clear error instead of the opaque `http.NewRequest` failure.
- **Non-200 error body** (CodeRabbit actionable @ tat_fetch.go:61) — non-200 responses now include a bounded 1 KiB body snippet so `code`/`msg` from the open platform surface in CLI/CI logs.
- **SKILL.md / README wording** (Copilot @ SKILL.md:32) — rewrote the contradictory paragraph into a table covering both `--as bot` (app_secret → minted TAT, or pre-issued TAT) and `--as user` (pre-issued UAT) paths. Mirror fix in README.md / README.zh.md.

Also added `internal/auth/tat_fetch_test.go` covering success, empty / malformed URL, non-200 with body, non-zero API code, and trailing-slash normalisation.

**Deferred (1/6, optional):**
- **default provider caches TAT via `sync.Once`** (CodeRabbit optional @ default_provider.go:120-142) — agreed the default provider should move to expiry-aware caching for symmetry with the env provider, but this is a behaviour change to existing code outside the scope of this PR (the original `sync.Once` caching is preserved exactly as-is by the refactor). Happy to tackle it as a follow-up.

Local gates green: `go vet`, `gofmt -l`, `go mod tidy`, `go test -race` on the touched packages, and `golangci-lint run --new-from-rev=origin/main` report 0 issues.

fengliu222:
Pushed `924a195` addressing the second round of feedback:

**Applied (6/6 actionable + nits):**
- **TTL clamp safety-margin invariant** (CodeRabbit @ env.go:245) — always subtract `tatSafetyMargin`; when the resulting TTL is non-positive we surface the token once without caching so the next call re-mints instead of serving an entry that already violates the margin. Added `TestResolveToken_TATShortLifetimeNotCached` to lock this behaviour in.
- **`http.DefaultClient` has no timeout** (CodeRabbit @ tat_fetch.go:31) — nil `hc` now falls back to `&http.Client{Timeout: 30s}` so a hung TCP cannot block the caller, and the exported doc-comment documents the contract. This matches the guard the env provider already applies.
- **Inner `trimmed` shadowed outer** (CodeRabbit @ tat_fetch.go:67-73) — renamed the body-snippet local to `body`.
- **Empty `appSecret` coverage** (CodeRabbit @ tat_fetch_test.go) — added `TestFetchTenantAccessToken_RejectsEmptyAppSecret` symmetric to the app_id case.
- **Stub call counter not goroutine-safe** (CodeRabbit @ env_test.go:18-28) — `withStubFetcher` now returns `*atomic.Int32`; call-sites use `calls.Load()`. Keeps `-race` clean if a future test drives the stub concurrently.
- **No-cache assertion** (CodeRabbit @ env_test.go:416-433) — `TestResolveToken_TATReturnsNilWhenSecretMissing` now also asserts `p.tatCache` stays at its zero value.

**Deferred (1):**
- **Validate URL before credentials for fail-fast ordering** (CodeRabbit nit @ tat_fetch.go:32-37) — CodeRabbit explicitly called this "purely stylistic; safe to ignore." The env provider already validates the brand independently, so in practice the credential checks never fire on a misconfigured URL. Skipped to keep this round tight; happy to apply if maintainers feel strongly.

Local gates still clean: `go vet`, `gofmt -l`, `go test -race` on the touched packages, and `golangci-lint run --new-from-rev=origin/main` all report 0 issues.

JackZhao10086:
Hi, fengliu222,
Thanks a lot for taking the time to share this suggestion! 🙏
We've received your feedback and really appreciate it. But, through the injection method of environment variables, all credentials should be explicitly injected through environment variables, and no other bypass methods should be provided. The tat can be injected by configuring the environment variable "LARKSUITE_CLI_TENANT_ACCESS_TOKEN".

PR #186, opened 2026-04-01
feat(im): add +bot-receive-diagnose for IM bot receive diagnostics (read-only)

xi-ByteD:
## What
- Add lark-cli im +bot-receive-diagnose (read-only) with --offline/--timeout/--event-type (default im.message.receive_v1 )
- Structured output: ok/summary/checks/next_steps to diagnose credentials, OpenAPI reachability, and event WebSocket startup
- Fix: WebSocket probe now returns deterministically within --timeout (no hanging)
## Why
- Addresses frequent “bot not responding / not receiving messages” troubleshooting need; aligns with #33 (bot doctor/probe request)
## Tests
- go test -v -race -count=1 -timeout=5m ./cmd/... ./internal/... ./shortcuts/... (pass)
## Docs
- Update skills/lark-im/SKILL.md
- Add skills/lark-im/references/lark-im-bot-receive-diagnose.md
## Limitations
- Cannot read console subscription/scope/availability directly; reported as actionable hints

## Summary by CodeRabbit

* **New Features**
  * Added im +bot-receive-diagnose CLI shortcut for bot receive diagnostics with timeout, offline, and event-type options; it's exposed in the IM shortcuts list.

* **Documentation**
  * Added command reference and skill docs describing usage, flags, structured output, and troubleshooting hints.

* **Tests**
  * Added tests covering command registration, validation, offline/online probe behaviors, and corrupted key-file handling.

* **Chores**
  * Improved keychain handling with optional file-based master-key selection via env var and made token acquisition context-aware.

PR #666, opened 2026-04-26
feat(event): add +bot-receive-diagnose shortcut for bot event reception readiness

### LINKED ISSUE #33: Feature Request: First-class bot doctor / probe for bot credentials
## Summary

`lark-cli` currently has `doctor`, but from an integration perspective it is still much more oriented toward config / login / user-auth flows than toward **bot credential health checks**.

For systems that integrate `lark-cli` as the official CLI execution surface, there is a strong need for a first-class **bot probe / bot doctor** capability.

## What we need

We need a way to:

- check availability using **bot credentials**
- directly retrieve **bot info**
- clearly determine whether the configured `app_id` / `app_secret` are valid

In other words, we need a bot-oriented health check, not just a user-auth/config-oriented doctor flow.

## Current workaround

Today this can be approximated by calling:

```bash
lark-cli api GET /open-apis/bot/v3/info --as bot
```

This works as a low-level probe, but it is not ideal because:

- the intent is not explicit
- output is not normalized as a health check / diagnosis result
- callers need to interpret API-level errors themselves
- it does not provide a stable first-class UX for bot health verification

## Proposed UX

Either of these would solve the problem well:

### Option A

```bash
lark-cli doctor --as bot
```

### Option B

```bash
lark-cli bot probe
```

A good result should ideally include fields such as:

- `ok`
- `app_id`
- resolved `bot_name`
- `bot_open_id`
- endpoint / brand info if relevant
- normalized error message if credentials are invalid or the bot cannot be resolved

## Why this matters

This is especially important for:

- AI agents
- orchestration systems
- wrapper CLIs
- server-side integrations that route across multiple bots

In these environments, bot credentials are often managed externally, and the first thing the integrator needs is a reliable answer to:

- does this bot credential pair actually work?
- which bot identity does it resolve to?
- is the failure caused by invalid credentials, endpoint mismatch, or something else?

Without a first-class bot doctor/probe, integrators usually end up re-implementing this logic outside the CLI.

## Suggested behavior

- Use bot credentials directly
- Return machine-readable structured output
- Normalize common failures, for example:
  - invalid `app_id` / `app_secret`
  - bot info endpoint unavailable
  - brand / endpoint mismatch
  - bot identity cannot be resolved
- Keep this stable enough for automation and CI-style health checks

## Related context

This request is related to broader identity / integration needs such as #18, but this issue is specifically about **first-class bot credential diagnosis and probing**.

## Expected impact

This would make `lark-cli` significantly easier to use as a reliable integration surface, not only as an interactive terminal tool.

xi-ByteD:
…on readiness

Adds a new read-only shortcut `event +bot-receive-diagnose` that helps users diagnose why a bot does not receive events (default: im.message.receive_v1). It performs structured checks on local bot configuration, tenant access token acquisition, open-platform endpoint reachability (inferred from the token call), and WebSocket long-connection subscription. Output is a machine- readable `{summary, checks, next_steps}` payload with pass/warn/fail/skip counts, and all standard `--format` options are supported.

The shortcut lives under shortcuts/event/ rather than shortcuts/im/ because it reuses the event-dispatcher/WebSocket stack shared with `event +subscribe` and the flag `--event-type` keeps the door open for diagnosing non-IM events.

Addresses review feedback on #186 by splitting the diagnose change from the keychain improvements that were originally bundled in the same PR. The keychain file-fallback has since been implemented independently in #285, so only the diagnose half is delivered here.

Closes #33

## Summary

Adds a new read-only shortcut `lark-cli event +bot-receive-diagnose` that helps bot owners
diagnose why they are not receiving events (especially `im.message.receive_v1`). Closes #33.

This PR is **split out from #186** per reviewer feedback. #186 bundled two unrelated changes
(a keychain fallback + the bot-receive diagnostic); this PR ships only the diagnostic, and
places it under the `event` module instead of `im` since event subscription readiness is the
primary concern.

## Motivation

Issue #33 reports a common class of support questions: "I configured my bot, but no events
are delivered." The root cause is usually one of:

- bot tenant access token cannot be resolved (wrong app id / secret, suspended app)
- event subscription mode is misconfigured (long-connection vs. callback URL)
- required scope `im:message:receive_as_bot` is missing
- a local WebSocket long-connection cannot be established

Today users have to open a ticket and walk the support team through the full setup. This
shortcut consolidates the checks into a single command and produces structured next-steps.

## What the shortcut does

Runs, in order:

1. `token_bot` — resolves the bot tenant access token via the configured credential provider
2. `endpoint_open` — inferred from (1): if the token call reached the open-platform, pass; else fail
3. `ws_dial` — opens a short-lived WebSocket long-connection and subscribes to `event-type`
4. Produces a `summary` table + a deduplicated, priority-sorted `next_steps` list

`--offline` skips network probes (useful for CI / air-gapped environments); in offline mode
only local config is validated and the summary is marked accordingly.

## Design notes

- **No raw HTTP.** Per the `shortcuts-no-raw-http` depguard rule, this shortcut does not
  import `net/http`. Endpoint reachability is derived from the SDK-backed token call instead
  of a separate probe, which also avoids SSRF static-analysis findings.
- **Testability seam.** `var botReceiveProbeWebsocket = diagnoseBotReceiveWebsocket` is a
  package-level hook so tests can stub WebSocket dialing deterministically.
- **Auth.** `AuthTypes: []string{"bot"}` — the shortcut only runs under `--as bot`.
- **Risk.** `Risk: "read"` — no writes, no side effects on the open-platform side.
- **Output.** Uses the standard envelope (`summary` / `checks` / `next_steps`) so
  `--format json` is stable for scripting.

## Files

- `shortcuts/event/bot_receive_diagnose.go` (new, 394 lines)
- `shortcuts/event/bot_receive_diagnose_test.go` (new, 304 lines, 7 tests)
- `shortcuts/event/shortcuts.go` (+1 line — register)
- `skills/lark-event/SKILL.md` (+1 line — table entry)
- `skills/lark-event/references/lark-event-bot-receive-diagnose.md` (new, 56 lines)
- `CHANGELOG.md` (+6 lines — `[Unreleased]` entry)

## Test plan

- `go vet ./...` — clean
- `make unit-test` — 30 packages, all green (new tests cover: registration, flag validation,
  offline short-circuit, online success, token-failure surfaces `endpoint_open:fail`,
  next-steps uniqueness & priority ordering)
- `golangci-lint run` — 0 issues (depguard rule verified)

## Relationship to #186

- The keychain file-fallback portion of #186 is **dropped from this PR**; upstream #285
  has since landed a more complete implementation (uses `internal/vfs` with
  `O_CREATE|O_EXCL` + retry for race safety), so a separate PR from me would duplicate
  that work.
- The bot-receive diagnostic in #186 lived under `shortcuts/im/`; this PR moves it to
  `shortcuts/event/` per review comment.
- #186 will be closed in favor of this PR.

Closes #33.

## Summary by CodeRabbit

* **New Features**
  * Added `+bot-receive-diagnose` read-only diagnostic to validate bot event receipt (default: im.message.receive_v1). Supports `--timeout` and `--offline`, validates args, skips network checks when offline, and outputs structured `{summary, checks, next_steps}`.

* **Documentation**
  * Added user-facing reference page and skill entry describing command usage, flags, checks, and output shape.

* **Tests**
  * Added comprehensive tests covering validation, offline/online behaviors, check outcomes, next-steps sorting/uniqueness, and scope derivation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Features
+
+- **event**: Add `event +bot-receive-diagnose` — a read-only diagnostic for bots that do not receive events (WebSocket / endpoint / credential checks with structured `{summary, checks, next_steps}` output). Closes #33.
+
 ## [v1.0.19] - 2026-04-24
 
 ### Features
--- a/shortcuts/event/bot_receive_diagnose.go
+++ b/shortcuts/event/bot_receive_diagnose.go
@@ -1,0 +1,467 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package event
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/larksuite/cli/internal/core"
+	"github.com/larksuite/cli/internal/credential"
+	"github.com/larksuite/cli/internal/output"
+	"github.com/larksuite/cli/shortcuts/common"
+
+	lark "github.com/larksuite/oapi-sdk-go/v3"
+	larkcore "github.com/larksuite/oapi-sdk-go/v3/core"
+	larkevent "github.com/larksuite/oapi-sdk-go/v3/event"
+	"github.com/larksuite/oapi-sdk-go/v3/event/dispatcher"
+	larkws "github.com/larksuite/oapi-sdk-go/v3/ws"
+)
+
+const defaultDiagnoseEventType = "im.message.receive_v1"
+
+// Seams for testing — keep package-level vars so tests can stub out the
+// real network probes without touching the shortcut wiring.
+var (
+	botReceiveProbeWebsocket = diagnoseBotReceiveWebsocket
+)
+
+type botReceiveCheck struct {
+	Name    string `json:"name"`
+	Status  string `json:"status"`
+	Message string `json:"message"`
+	Hint    string `json:"hint,omitempty"`
+}
+
+type botReceiveSummary struct {
+	OK    bool `json:"ok"`
+	Pass  int  `json:"pass"`
+	Warn  int  `json:"warn"`
+	Fail  int  `json:"fail"`
+	Skip  int  `json:"skip"`
+	Total int  `json:"total"`
+}
+
+type botReceivePayload struct {
+	EventType string            `json:"event_type"`
+	Summary   botReceiveSummary `json:"summary"`
+	Checks    []botReceiveCheck `json:"checks"`
+	NextSteps []string          `json:"next_steps,omitempty"`
+}
+
+type silentSDKLogger struct{}
+
+func (l *silentSDKLogger) Debug(_ context.Context, _ ...interface{}) {}
+func (l *silentSDKLogger) Info(_ context.Context, _ ...interface{})  {}
+func (l *silentSDKLogger) Warn(_ context.Context, _ ...interface{})  {}
+func (l *silentSDKLogger) Error(_ context.Context, _ ...interface{}) {}
+
+var _ larkcore.Logger = (*silentSDKLogger)(nil)
+
+// BotReceiveDiagnose is a read-only diagnostic shortcut that inspects whether
+// a bot is correctly set up to receive events (default: im.message.receive_v1).
+// It reuses the event/WebSocket stack shared with `event +subscribe`.
+//
+// Scopes is intentionally the empty slice rather than a hard-coded scope list:
+// bot (tenant access token) credentials do not carry an OAuth-style granted
+// scope list, so the shared scope pre-check cannot verify a specific event
+// permission for bot identity. Declaring a static scope here would both be a
+// no-op today and risk blocking the diagnostic in the future if the shared
+// pre-check ever learns to fetch bot scopes — bots *missing* the scope are
+// exactly the audience this diagnostic is designed to help. The CLI surfaces
+// the required scope for the specific --event-type as an advisory check
+// inside Execute instead.
+var BotReceiveDiagnose = common.Shortcut{
+	Service:     "event",
+	Command:     "+bot-receive-diagnose",
+	Description: "Diagnose why a bot does not receive events (bot-only, read-only)",
+	Risk:        "read",
+	Scopes:      []string{},
+	AuthTypes:   []string{"bot"},
+	HasFormat:   true,
+	Flags: []common.Flag{
+		{Name: "offline", Type: "bool", Desc: "skip network and WebSocket checks; only verify local bot configuration and credential presence"},
+		{Name: "timeout", Type: "int", Default: "5", Desc: "timeout in seconds for network and WebSocket checks (1-60)"},
+		{Name: "event-type", Default: defaultDiagnoseEventType, Desc: "event type to diagnose (default im.message.receive_v1)"},
+	},
+	DryRun: func(_ context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		return common.NewDryRunAPI().
+			Desc("Diagnose bot event receive readiness without changing remote state").
+			Set("command", "event +bot-receive-diagnose").
+			Set("event_type", runtime.Str("event-type")).
+			Set("offline", runtime.Bool("offline")).
+			Set("timeout_sec", runtime.Int("timeout"))
+	},
+	Validate: func(_ context.Context, runtime *common.RuntimeContext) error {
+		if runtime.Int("timeout") < 1 || runtime.Int("timeout") > 60 {
+			return output.ErrValidation("--timeout must be an integer between 1 and 60")
+		}
+		if strings.TrimSpace(runtime.Str("event-type")) == "" {
+			return output.ErrValidation("--event-type cannot be empty")
+		}
+		return nil
+	},
+	Execute: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		offline := runtime.Bool("offline")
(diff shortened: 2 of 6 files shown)


xi-ByteD:
Force-pushed `901ed69 → 036ccc3` to address three concerns surfaced during review. Same single-commit layout; 6 files, +912 / -0 (net +150 over the previous push, mostly new tests).

**1. Static `Scopes` declaration removed** — declaring `Scopes: ["im:message:receive_as_bot"]` on the Shortcut was misleading for two reasons:
   - Bot (tenant access token) credentials don't carry an OAuth-style granted-scope list, so the shared `checkScopePrereqs` pre-check is effectively a no-op for bot identity — the CLI cannot actually verify the permission.
   - If a future change teaches the pre-check to fetch bot scopes, a bot that is **missing** the scope would be blocked from running the diagnostic at all — which is exactly the audience this diagnostic is supposed to help.

**2. Required scope is now derived per `--event-type`** — the previous version hard-coded `im:message:receive_as_bot` in a `receive_scope` warn entry, which was wrong for non-IM event types. Replaced with a new `scope_required` advisory:
   - Known event types are mapped to their canonical scope (e.g. `im.message.receive_v1 → im:message:receive_as_bot`, `im.message.message_read_v1 → im:message`, `im.chat.member.* → im:chat:readonly`).
   - Unknown event types fall back to a generic hint that asks the user to confirm the scope in the developer console — no false mapping.

**3. WebSocket readiness signal inverted** — the underlying SDK's `ws.Client.Start()` blocks on `select {}` on the happy path and never returns `nil`, so the old logic treated every healthy run as `warn`. Split the single `context.WithTimeout` into a parent `wsCtx` plus a dedicated `readinessTimer`:
   - `readinessTimer.C` fires (connected + subscribed stayed alive for `--timeout` seconds without error) → **pass**
   - Non-`ctx` error from `Start()` → **fail**
   - Parent ctx canceled (e.g. user Ctrl-C) or SDK returns a ctx error → **warn** with an actionable hint

Three new tests were added to lock down the scope-derivation behavior (`ScopeAdvisoryMatchesKnownEventType`, `ScopeAdvisoryFallsBackForUnknownEventType`, table-driven `TestDeriveScopeForEventType`). `go vet` and `go test ./shortcuts/event/...` are clean locally.

Ready for another pass — thanks!

xi-ByteD:
Follow-up force-push `036ccc3 → d85ce44` — catches a regression I introduced while addressing the earlier review, and closes out this round with a clean quality-gate pass.

### What slipped through the previous push

My previous push removed the static `Scopes` declaration entirely to act on the review feedback that a hard-coded `im:message:receive_as_bot` was misleading. That broke a repo-wide invariant test — [`shortcuts/register_test.go::TestAllShortcutsScopesNotNil`](https://github.com/larksuite/cli/blob/main/shortcuts/register_test.go) requires every Shortcut to declare `Scopes` explicitly (using `[]string{}` when there are no applicable scopes). I only noticed this when re-running `go test ./...` under a re-review.

### Fix

Changed `Scopes` back to `[]string{}` with a comment explaining why a static list is intentionally avoided for this shortcut. This is consistent with the pattern already used by the sibling shortcut `event +subscribe` (same package, `subscribe.go` line 87) and with `drive_task_result` — i.e. it is the established project convention for commands whose required scopes are context-dependent and cannot be pre-declared.

All three of the original review findings remain fully addressed:
1. No hard-coded scope name on the Shortcut definition (the empty slice communicates "we don't verify a specific scope here" without suggesting otherwise).
2. Required scope is still derived per `--event-type` inside `Execute` and surfaced as a `scope_required` advisory.
3. WebSocket readiness signal still uses a dedicated `readinessTimer` so that staying connected for `--timeout` seconds is reported as `pass`, real `Start()` errors as `fail`, and parent-ctx cancellation as `warn`.

### Final quality-gate status (local)

| Gate | Result |
|------|--------|
| `go vet ./...` | clean |
| `go build ./...` | clean |
| `go test ./...` (non-e2e, 57 packages) | all green, exit 0 |
| `shortcuts/event` package tests | 29 tests PASS (incl. 3 new scope-derivation tests) |
| `shortcuts/` invariant tests | `TestAllShortcutsScopesNotNil` PASS |
| e2e packages (`tests/cli_e2e/*`) | require a prebuilt `./lark-cli` binary; fail the same way on `origin/main` and are unrelated to this PR |

### Final shape of the PR

- 1 commit, 6 files, +916 / -0
- Net change from the very first push (+762) is mostly new tests plus the WebSocket split between `wsCtx` (for Ctrl-C propagation) and `readinessTimer` (for the positive-signal branch)

Ready for another pass — thanks!

xi-ByteD:
Follow-up force-push `e0bc2de → 1517abd` — final cleanup after the round-2 review.

What changed in this push:
- Kept the round-2 WebSocket fix: `endpoint_ws` no longer reports `pass` just because no startup error was observed within `--timeout`. The SDK may still be sitting in its internal auto-reconnect loop, so that case remains `warn`.
- Removed one stale inline comment in `bot_receive_diagnose.go` that still described the timeout branch as a healthy signal.
- Updated `skills/lark-event/references/lark-event-bot-receive-diagnose.md` to match the final implementation:
  - `receive scope` → `scope_required`
  - WebSocket wording no longer overstates readiness
  - notes now reflect that `summary.ok=true` means “no hard failure”, not “delivery is confirmed”

Final local checks:
- `go vet ./...` ✅
- `go build ./...` ✅
- `go test ./shortcuts/event/...` ✅
- `go test ./shortcuts/` ✅
- non-e2e repo tests are green for all PR-related packages

One unrelated note: `internal/binding` is still intermittently flaky under full-repo concurrent test runs, but it passes when re-run in isolation; this is pre-existing and not caused by this PR.

Current PR shape:
- 1 commit
- 6 files
- head: `1517abd`

At this point code, tests, and reference docs are aligned. Ready for another pass — thanks!

PR #674, opened 2026-04-27
feat(drive): extend +add-comment to support slides targets

ethan-zhx:
## Summary
  Extend `drive +add-comment` to support slides targets in addition to existing doc/docx flows. This also improves wiki target resolution, validation, dry-run behavior, and skill
  documentation so agents can call the command with clearer and more reliable inputs.

  ## Changes
  - Add slides support to `drive +add-comment`
  - Support resolving wiki URLs to `doc` / `docx` / `slides` targets before creating comments
  - Add validation for slides-specific anchor format:
    - slides require `--block-id <slide-block-type>!<xml-id>`
    - slides do not allow `--full-comment` or `--selection-with-ellipsis`
  - Improve dry-run request generation so it reflects resolved target type and actual API flow
  - Expand unit tests for doc parsing, mode resolution, locate-doc fallback, reply element parsing, and slides comment flows
  - Update `lark-drive` skill docs and examples for doc/docx/slides comment creation

  ## Test Plan
  - [x] Added/updated unit tests in `shortcuts/drive/drive_add_comment_test.go`
  - [ ] Unit tests pass
  - [ ] Manual local verification confirms the `lark xxx` command works as expected

  ## Related Issues
  - None

## Summary by CodeRabbit

* **New Features**
  * Add-comment now supports Slides; slide comments require a compound --block-id (<slide-block-type>!<xml-id>) and include slide_block_type in outputs and requests. Dry-run and execution route slide-specific flows and wiki-resolved slides.

* **Documentation**
  * CLI docs, examples, and flag help updated to include slides, allowed --doc/--type inputs, required block-id format, and unsupported flags (--full-comment, --selection-with-ellipsis) for slides.

* **Tests**
  * Added unit and end-to-end tests for slide parsing, validation, dry-run, execution, responses, and error cases.

* **Bug Fixes**
  * Dry-run now computes comment mode earlier and surfaces wiki resolution errors instead of continuing silently.

### DIFF
--- a/shortcuts/drive/drive_add_comment.go
+++ b/shortcuts/drive/drive_add_comment.go
@@ -64,7 +64,7 @@ const (
 var DriveAddComment = common.Shortcut{
 	Service:     "drive",
 	Command:     "+add-comment",
-	Description: "Add a full-document or local comment to doc/docx/sheet, also supports wiki URL resolving to doc/docx/sheet",
+	Description: "Add a comment to doc/docx/sheet/slides, also supports wiki URL resolving to doc/docx/sheet/slides",
 	Risk:        "write",
 	Scopes: []string{
 		"docx:document:readonly",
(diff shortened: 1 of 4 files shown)


fangshuyu-768 (commented):
One real concern about a flag-validation gap on the wiki→slides path; rest looks solid.

shortcuts/drive/drive_add_comment.go
```diff
@@ -425,11 +465,21 @@ func resolveCommentTarget(ctx context.Context, runtime *common.RuntimeContext, i
 			WikiToken:  docRef.Token,
 		}, nil
 	}
+	if objType == "slides" {
```
fangshuyu-768, line 471: **`--selection-with-ellipsis` is silently dropped on the wiki→slides path.** Validate at line 114 rejects `--selection-with-ellipsis` for slides, but only when `docRef.Kind == "slides"` — wiki URLs go through the `"wiki"` branch and skip that check. `resolveCommentTarget` rejects `--full-comment` for slides at line 454, but doesn't reject `--selection-with-ellipsis`, so the early-return here passes through with `mode == commentModeLocal`. Downstream `executeSlidesComment` (line 801) only reads `--block-id` and ignores the selection flag entirely, so:

- Wiki URL + `--selection-with-ellipsis` only → user gets `"slide comments require --block-id ..."` from `parseSlidesBlockRef`, with no hint that selection is unsupported for slides at all.
- Wiki URL + `--block-id` + `--selection-with-ellipsis` → request succeeds and the selection is silently discarded.

Suggest mirroring the direct-slides validation here, e.g. before this early return:
```go
if objType == "slides" && strings.TrimSpace(runtime.Str("selection-with-ellipsis")) != "" {
    return resolvedCommentTarget{}, output.ErrValidation(
        "wiki resolved to %q, but --selection-with-ellipsis is not applicable for slide comments; use --block-id <slide-block-type>!<xml-id>", objType)
}
```
(Same gap appears in DryRun at line ~176 — `parseSlidesBlockRef`'s error is discarded with `_` and overridden with a placeholder, so a malformed `--block-id` on the wiki→slides path won't surface in dry-run preview, only at Execute time. Smaller issue, but worth aligning.)

fangshuyu-768 (approved):

PR #680, opened 2026-04-27
feat(common): backfill resource URL when create APIs omit it

caojie0621:
## Summary                                                                                                                                                                                 
                                                            
  - Add `common.BuildResourceURL(brand, kind, token)` helper that returns brand-standard URLs (`www.feishu.cn` / `www.larksuite.com`) for known resource kinds (`docx` / `doc` / `sheet` /   `wiki` / `file` / `folder`). The standard host transparently redirects to the tenant domain.
  - Wire it into the create paths so the output always carries a usable `url` field even when the backend response omits one — e.g. degraded MCP responses where `ExtractMCPResult` collapses
   multi-content into a plain string, the drive upload API which never returns a URL, or wiki node create which only returns tokens.                                                         
  - The fallback is non-invasive: it only fills `url` when the backend value is empty, so tenant-specific URLs returned by the API are preserved.
                                                                                                                                                                                             
  ## Affected entries                                       

  | Path | Behavior change |                                                                                                                                                                 
  |---|---|
  | `docs +create --api-version v1` | `doc_url` falls back to `/docx/<doc_id>` when MCP omits it |                                                                                           
  | `docs +create --api-version v2` | `data.document.url` falls back to `/docx/<document_id>` |
  | `sheets +create` | `out.url` falls back to `/sheets/<token>` (also fixes a pre-existing bug where missing url produced a literal `null` field) |                                         
  | `drive +create-folder` | url falls back to `/drive/folder/<token>` |                                                                                                                     
  | `drive +import` | url falls back to `/<docx\|sheets\|base>/<token>` based on resolved `resultType` |                                                                                     
  | `drive +upload` | **newly exposes** `url`, only when `parent_type=explorer` (wiki-hosted files have no standalone URL) |                                                                 
  | `wiki +node-create` | **newly exposes** `url` = `/wiki/<node_token>` |                                                                                                                   
                                                                                                                                                                                             
  `drive +create-shortcut` is intentionally not covered: shortcut URL form depends on the underlying file kind, which the create payload does not carry — better to omit than to guess wrong.
                                                                                                                                                                                             
  ## Test plan                                                                                                                                                                               
                                                            
  - [x] `go test ./shortcuts/common/ ./shortcuts/doc/ ./shortcuts/sheets/ ./shortcuts/drive/ ./shortcuts/wiki/` — all green                                                
  - [x] New `BuildResourceURL` helper has 16 table-driven cases covering brand/kind/token edge cases
  - [x] `wiki_node_create_test.go` extended to assert the new `url` field on output                                       

## Summary by CodeRabbit

* **New Features**
  * Creation, import, and upload operations now populate brand-standard resource URLs (Feishu/Lark) for docs, sheets, wikis, folders, files, mindnotes and slides when upstream responses omit them; tenant-provided URLs are preserved.

* **Bug Fixes**
  * Avoided missing or invalid URL outputs by trimming tokens and normalizing kinds; wiki-parent uploads intentionally omit fragile fallbacks to prevent broken links.

* **Tests**
  * Added extensive tests covering URL fallbacks, token trimming, kind normalization (case-insensitive), and preservation behavior.

### DIFF
--- a/shortcuts/common/resource_url.go
+++ b/shortcuts/common/resource_url.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package common
+
+import (
+	"strings"
+
+	"github.com/larksuite/cli/internal/core"
+)
+
+// BuildResourceURL returns a brand-standard, user-facing URL for a freshly
+// created Lark resource. It is intended as a fallback when the create API does
+// not return a URL field (e.g. drive +upload, wiki +node-create) or when the
+// returned URL is empty (e.g. degraded MCP responses for docs +create v1).
+//
+// The returned URL points at the brand's standard host (www.feishu.cn /
+// www.larksuite.com), which transparently redirects to the tenant-specific
+// domain. It is NOT a guess at the tenant's vanity domain.
+//
+// Returns "" when token is empty or kind is unrecognized — callers should
+// only set the field when the result is non-empty so that "" never overrides
+// a real URL the backend already returned.
+func BuildResourceURL(brand core.LarkBrand, kind, token string) string {
+	token = strings.TrimSpace(token)
+	if token == "" {
+		return ""
+	}
+
+	host := "https://www.feishu.cn"
+	if brand == core.BrandLark {
+		host = "https://www.larksuite.com"
+	}
+
+	switch strings.ToLower(strings.TrimSpace(kind)) {
+	case "docx":
+		return host + "/docx/" + token
+	case "doc":
+		return host + "/doc/" + token
+	case "sheet":
+		return host + "/sheets/" + token
+	case "bitable":
+		return host + "/base/" + token
+	case "wiki":
+		return host + "/wiki/" + token
+	case "file":
+		return host + "/file/" + token
+	case "folder":
+		return host + "/drive/folder/" + token
+	case "mindnote":
+		return host + "/mindnote/" + token
+	case "slides":
+		return host + "/slides/" + token
+	default:
+		return ""
+	}
+}
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -51,6 +51,7 @@ func executeCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	}
 
 	augmentDocsCreatePermission(runtime, data)
+	fallbackDocsCreateURLV2(runtime, data)
 	runtime.OutRaw(data, nil)
 	return nil
 }
@@ -82,5 +83,25 @@ func augmentDocsCreatePermission(runtime *common.RuntimeContext, data map[string
 	}
 	if grant := common.AutoGrantCurrentUserDrivePermission(runtime, docID, "docx"); grant != nil {
 		data["permission_grant"] = grant
+	}
+}
+
+// fallbackDocsCreateURLV2 fills data.document.url with a brand-standard URL
+// when the OpenAPI response did not include one. Backfills only when missing,
+// so any tenant-specific URL the backend returned is preserved.
+func fallbackDocsCreateURLV2(runtime *common.RuntimeContext, data map[string]interface{}) {
+	doc, _ := data["document"].(map[string]interface{})
+	if doc == nil {
+		return
+	}
+	if strings.TrimSpace(common.GetString(doc, "url")) != "" {
+		return
+	}
+	docID := strings.TrimSpace(common.GetString(doc, "document_id"))
+	if docID == "" {
+		return
+	}
+	if u := common.BuildResourceURL(runtime.Config.Brand, "docx", docID); u != "" {
+		doc["url"] = u
 	}
 }
--- a/shortcuts/drive/drive_import.go
+++ b/shortcuts/drive/drive_import.go
@@ -119,8 +119,12 @@ var DriveImport = common.Shortcut{
 		if status.Token != "" {
 			out["token"] = status.Token
 		}
-		if status.URL != "" {
-			out["url"] = status.URL
+		if statusURL := strings.TrimSpace(status.URL); statusURL != "" {
+			out["url"] = statusURL
+		} else if status.Token != "" {
+			if u := common.BuildResourceURL(runtime.Config.Brand, normalizeDriveImportKindForURL(resultType, spec.DocType), status.Token); u != "" {
+				out["url"] = u
+			}
 		}
 		if status.JobErrorMsg != "" {
 			out["job_error_msg"] = status.JobErrorMsg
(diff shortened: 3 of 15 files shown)


shortcuts/common/resource_url.go
```diff
@@ -0,0 +1,53 @@
...
+	host := "https://www.feishu.cn"
+	if brand == core.BrandLark {
+		host = "https://www.larksuite.com"
+	}
+
+	switch strings.ToLower(strings.TrimSpace(kind)) {
+	case "docx":
+		return host + "/docx/" + token
+	case "doc":
+		return host + "/docs/" + token
```
fangshuyu-768: **Inconsistent with the rest of the codebase.** The legacy-doc kind here produces `/docs/<token>` (plural), but every other place in the repo parses legacy doc URLs as `/doc/<token>` (singular):

- `shortcuts/drive/drive_apply_permission.go:38` — `{"/doc/", "doc"}`
- `shortcuts/drive/drive_apply_permission.go:62` — error message lists `/doc/` as a supported path
- `shortcuts/drive/drive_add_comment.go:364` — `extractURLToken(raw, "/doc/")`
- `shortcuts/doc/helpers.go:31` — `extractDocumentToken(raw, "/doc/")`

A URL produced by `BuildResourceURL(brand, "doc", token)` therefore would not round-trip through this project's own parsers (e.g. you couldn't paste it into `drive +apply-permission --token <url>`). Suggest changing this branch to `host + "/doc/" + token` and updating the test fixture, so all sites agree.

fangshuyu-768 (commented):
Nice helper and clean fallback strategy — the create paths definitely benefit from a stable `url` field. A few inline comments below; one is a real inconsistency with how the rest of the repo parses legacy `doc` URLs, and the others are about hardening the contract and shoring up test coverage.

shortcuts/common/resource_url.go
```diff
@@ -0,0 +1,53 @@
...
+		return host + "/sheets/" + token
+	case "bitable":
+		return host + "/base/" + token
+	case "wiki":
+		return host + "/wiki/" + token
+	case "file":
+		return host + "/file/" + token
+	case "folder":
+		return host + "/drive/folder/" + token
+	default:
```
fangshuyu-768, line 54: Optional: kinds `mindnote` and `slides` are recognized elsewhere in the repo (e.g. `drive_apply_permission.go:62`, `drive_apply_permission.go:96`) but are missing here. If/when a create path for either is added, `BuildResourceURL` will silently return `""` and the caller will think the URL was "already set by the backend". Worth either adding them now (`/mindnotes/<token>`, `/slides/<token>`) or leaving a TODO so the next contributor doesn't get a silent miss.

shortcuts/drive/drive_import.go
```diff
@@ -121,6 +121,10 @@ var DriveImport = common.Shortcut{
 		}
 		if status.URL != "" {
 			out["url"] = status.URL
+		} else if status.Token != "" {
+			if u := common.BuildResourceURL(runtime.Config.Brand, resultType, status.Token); u != "" {
```
fangshuyu-768: `resultType` is passed straight into `BuildResourceURL`, but its values come from two sources with subtly different contracts:

1. `spec.DocType` (user `--type` flag, validated to `docx|sheet|bitable`) — matches `BuildResourceURL` cases.
2. `status.DocType` = `common.GetString(result, "type")` (server response, see `drive_import_common.go:331`) — not normalized.

If the import API ever returns something like `"sheets"` or `"sheet_v2"`, the switch in `BuildResourceURL` silently falls through to `default` and the user gets no `url` despite `status.Token` being set. Suggest either normalizing `status.DocType` against the known set before this call, or adding a defensive map (e.g. `"sheets" → "sheet"`).

shortcuts/doc/docs_create_v2.go
```diff
@@ -83,4 +83,9 @@ func augmentDocsCreatePermission(runtime *common.RuntimeContext, data map[string
 	if grant := common.AutoGrantCurrentUserDrivePermission(runtime, docID, "docx"); grant != nil {
 		data["permission_grant"] = grant
 	}
+	if strings.TrimSpace(common.GetString(doc, "url")) == "" {
+		if u := common.BuildResourceURL(runtime.Config.Brand, "docx", docID); u != "" {
+			doc["url"] = u
+		}
+	}
 }
```
fangshuyu-768, line 107: The function is named `augmentDocsCreatePermission` and its doc comment still says *"grants full_access to the current CLI user when the document was created with bot identity"* — but it now also backfills `doc.url`. Either rename to something like `augmentDocsCreateOutput` and update the comment, or move the URL backfill into a sibling helper called from `executeCreateV2`, so the function's contract matches its name. Future readers will assume "permission" functions only touch permissions.

shortcuts/sheets/sheet_create.go
```diff
@@ -108,7 +109,11 @@ var SheetCreate = common.Shortcut{
 		out := map[string]interface{}{
 			"spreadsheet_token": token,
 			"title":             title,
-			"url":               spreadsheet["url"],
+		}
+		if url, _ := spreadsheet["url"].(string); strings.TrimSpace(url) != "" {
+			out["url"] = url
+		} else if u := common.BuildResourceURL(runtime.Config.Brand, "sheet", token); u != "" {
+			out["url"] = u
```
fangshuyu-768, line 117: Test coverage gap: this is now a two-branch fallback (real URL → built URL → omitted), but `shortcuts/sheets/sheet_create_test.go` doesn't cover any of the three. The same gap exists for `docs +create` v1/v2, `drive +create-folder`, `drive +import`, and `drive +upload` — only `wiki_node_create_test.go` got a new assertion. Suggest adding at least one test per path that stubs an empty/missing `url` in the API response and asserts the helper-built URL appears, otherwise a future regression in `BuildResourceURL` (e.g. the `/docs/` vs `/doc/` issue above) won't be caught at the call site.

shortcuts/drive/drive_upload.go
```diff
@@ -146,6 +146,11 @@ var DriveUpload = common.Shortcut{
 			"file_name":  fileName,
 			"size":       fileSize,
 		}
+		if target.ParentType == driveUploadParentTypeExplorer {
+			if u := common.BuildResourceURL(runtime.Config.Brand, "file", fileToken); u != "" {
```
fangshuyu-768, line 153: Minor: `parent_type` only has two values today (`explorer`, `wiki` — see `drive_upload_common.go:24-25`), so this branch is correct. But there's no comment explaining *why* wiki is excluded; a one-liner like `// wiki-hosted files have no standalone /file/<token> URL — only the wiki node URL, which the upload response doesn't carry.` would prevent someone from "fixing" the missing `else` later.

fangshuyu-768 (approved):

PR #722, opened 2026-04-29
feat(sheets): add sheet management shortcuts

caojie0621:
## Summary

  This PR adds sheet management shortcuts for Sheets so agents can create, copy,
  delete, and update sheets without hand-writing raw API payloads. It also fixes
  the related validation gaps and consolidates the `lark-sheets` reference docs
  from per-command files into grouped capability-based docs.

  ## Changes

  - Add `+create-sheet`, `+copy-sheet`, `+delete-sheet`, and `+update-sheet` in `shortcuts/sheets/sheet_manage.go`
  - Add unit tests and dry-run E2E coverage for request shapes and validation behavior
  - Consolidate `skills/lark-sheets/references` from 43 per-command files into 10 grouped docs:
    - `lark-sheets-spreadsheet-management.md`
    - `lark-sheets-sheet-management.md`
    - `lark-sheets-cell-data.md`
    - `lark-sheets-cell-style-and-merge.md`
    - `lark-sheets-cell-images.md`
    - `lark-sheets-row-column-management.md`
    - `lark-sheets-filter-views.md`
    - `lark-sheets-dropdown.md`
    - `lark-sheets-float-images.md`
    - `lark-sheets-formula.md`

  ## Test Plan

  - [x] `make unit-test`
  - [x] `go test ./shortcuts/sheets`

## Summary by CodeRabbit

* **New Features**
  * Added sheet-level commands: create, copy, update, delete sheets
  * Expanded cell commands: read, write, append, find, replace
  * Added cell styling & merge: set-style, batch-style, merge/unmerge cells
  * Row/column management: add/insert/update/move/delete dimensions
  * Media support: upload images for floating/cell image workflows

* **Documentation**
  * Reorganized and consolidated Sheets reference docs; bumped skill version to 1.2.0


shortcuts/sheets/sheet_manage.go
```diff
@@ -0,0 +1,489 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/larksuite/cli/internal/validate"
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+var sheetProtectLockValues = []string{"LOCK", "UNLOCK"}
+
+func sheetCollectionPath(token string) string {
```
fangshuyu-768: The official "operate sheets" API uses a single `POST /open-apis/sheets/v2/spreadsheets/{spreadsheet_token}/sheets_batch_update` endpoint for add/copy/delete sheet operations. I could not find a documented v3 `.../spreadsheets/{token}/sheets`, `.../copy`, or `DELETE .../sheets/{sheet_id}` route. As written, `+create-sheet`, `+copy-sheet`, and `+delete-sheet` are all building requests against undocumented endpoints and the dry-run/unit tests are locking in the wrong request shape.

shortcuts/sheets/sheet_manage.go
```diff
@@ -0,0 +1,489 @@
...
+	if runtime.Changed("title") {
+		body["title"] = runtime.Str("title")
+	}
+	if runtime.Changed("index") {
+		body["index"] = runtime.Int("index")
+	}
+	return body
+}
+
+func buildSheetCreateCopyOutput(token string, data map[string]interface{}) map[string]interface{} {
```
fangshuyu-768: Even after switching to the documented `sheets_batch_update` API, this output helper will still be wrong: the documented response shape is `data.replies[0].addSheet.properties` / `copySheet.properties` / `deleteSheet.{result,sheetId}`, not `data.sheet`. In other words, the current fake responses in `sheet_manage_test.go` are masking a real response-shape mismatch, and `sheet_id` would come back empty against the documented payload.

shortcuts/sheets/lark_sheets_sheet_management.go
```diff
@@ -0,0 +1,489 @@
...
+	Service:     "sheets",
+	Command:     "+create-sheet",
+	Description: "Create a sheet in an existing spreadsheet",
+	Risk:        "write",
+	Scopes:      []string{"sheets:spreadsheet:write_only", "sheets:spreadsheet:read"},
+	AuthTypes:   []string{"user", "bot"},
+	Flags: []common.Flag{
+		{Name: "url", Desc: "spreadsheet URL"},
+		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
+		{Name: "title", Desc: "sheet title"},
```
fangshuyu-768, line 505: `title` looks required for sheet creation in the documented add-sheet payload (`requests[].addSheet.properties.title`). Right now validation only checks the title when it is explicitly provided, so `lark-cli sheets +create-sheet --spreadsheet-token shtDummy --dry-run` succeeds with an empty body. That means we defer a missing-required-field error to the backend instead of catching it up front.

shortcuts/sheets/lark_sheets_sheet_management.go
```diff
@@ -0,0 +1,489 @@
...
+	Description: "Copy a sheet within a spreadsheet",
+	Risk:        "write",
+	Scopes:      []string{"sheets:spreadsheet:write_only", "sheets:spreadsheet:read"},
+	AuthTypes:   []string{"user", "bot"},
+	Flags: []common.Flag{
+		{Name: "url", Desc: "spreadsheet URL"},
+		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
+		{Name: "sheet-id", Desc: "source sheet ID", Required: true},
+		{Name: "title", Desc: "new sheet title"},
+		{Name: "index", Type: "int", Desc: "new sheet index (0-based)"},
```
fangshuyu-768, line 558: I do not see an `index` field in the documented `copySheet.destination` payload for the sheet-operations API. The doc says the copied sheet is created immediately after the source sheet, and only `destination.title` is configurable. If that is still the current contract, exposing `--index` here is misleading because the flag has no backend equivalent once we switch to the documented batch-update endpoint.

shortcuts/sheets/lark_sheets_sheet_management.go
caojie0621: title is intentionally not required here.

  The current contract for +create-sheet is:

  - --title is optional
  - if --title is provided, it must pass local validation
  - if --index is provided, it must also pass local validation

  So I did not make title required. I reverted that part and kept the validation scoped to explicitly provided --title values only.

shortcuts/sheets/lark_sheets_sheet_management.go
caojie0621: You're right that copySheet.destination itself does not support index in the documented v2 sheet-operations payload. The --index flag here is intentionally implemented as a client-side two-step flow rather than a direct field mapping: 1. send copySheet to create the copy next to the source sheet 2. if --index is provided, send a follow-up updateSheet request to move the copied sheet to the requested position So --index is not forwarded into copySheet.destination; it is applied by a second operation after the copy succeeds. That is why the flag still has a real backend effect even after switching to the documented batch-update endpoint.

fangshuyu-768 (approved):

PR #730, opened 2026-04-30
fix: preserve filenames for file uploads

### LINKED ISSUE #729: Approval attachment uploads lose filename and show as unknown file
## Summary

When creating a Feishu/Lark approval instance with an `attachmentV2` widget, files uploaded through `lark-cli api POST /open-apis/approval/v4/files/upload` are accepted and can be attached to the approval, but the approval UI displays them as `unknown file` because the original filename/extension is not preserved in the created approval instance.

There is also a mismatch between `instances/preview` and `instances`:

- `instances/preview` accepts object-shaped `attachmentV2.value` entries that include `code`, `name`, and `type`.
- `instances` rejects the same object-shaped values with `Widget value is not string`.

This makes it hard for agents/scripts to create approval instances with user-friendly attachment names.

## Environment

- `lark-cli version 1.0.22`
- macOS
- Command path: `lark-cli api ...`
- Approval API: `/open-apis/approval/v4/files/upload`, `/open-apis/approval/v4/instances/preview`, `/open-apis/approval/v4/instances`

## Reproduction

1. Upload a PDF attachment and include `name` and `type` in multipart form fields:

```bash
lark-cli api POST /open-apis/approval/v4/files/upload \
  --as bot \
  --file path/to/invoice.pdf \
  --data '{"type":"attachment","name":"invoice.pdf"}' \
  --format json
```

The upload succeeds and returns a file `code`.

2. Use the returned `code` in an approval `attachmentV2` widget:

```json
{
  "id": "widget_attachment",
  "type": "attachmentV2",
  "value": ["<uploaded-file-code>"]
}
```

3. Create the approval instance:

```bash
lark-cli api POST /open-apis/approval/v4/instances \
  --as bot \
  --data - \
  --format json < approval_request.json
```

The instance is created successfully, but the approval UI shows the attachment as `unknown file` instead of `invoice.pdf`.

## Preview vs Create Mismatch

The preview endpoint accepts object-shaped values:

```json
{
  "id": "widget_attachment",
  "type": "attachmentV2",
  "value": [
    {
      "code": "<uploaded-file-code>",
      "name": "invoice.pdf",
      "type": "attachment"
    }
  ]
}
```

This request succeeds against:

```bash
lark-cli api POST /open-apis/approval/v4/instances/preview --as bot --data -
```

But the create endpoint rejects the same payload:

```bash
lark-cli api POST /open-apis/approval/v4/instances --as bot --data -
```

Error:

```text
Widget value is not string. index= 5, widget type= attachmentV2.
```

## Expected Behavior

One of the following would make the workflow reliable:

1. `files/upload` preserves the uploaded filename/extension so approval UI displays `invoice.pdf` instead of `unknown file`.
2. `instances` accepts the same object-shaped `attachmentV2.value` that `instances/preview` accepts, including `name` and `type`.
3. `lark-cli` documents the limitation and provides a recommended way to create approval attachments with visible filenames.

## Actual Behavior

- Uploaded attachments can be attached to approval instances.
- Approval UI displays the attachment as `unknown file`.
- Supplying a filename in `files/upload --data '{"name":"..."}'` does not appear to affect the displayed filename.
- Supplying object-shaped attachment values passes preview but fails create.

## Workaround

For image-like attachments, converting a PDF to PNG and uploading it as `type=image` avoids the worst UX because the image is previewable, but this is not a good replacement for preserving the original filename for PDFs and other documents.

rendylong:
## Summary

Fix generic `--file` uploads so multipart filenames are preserved for local files.

`BuildFormdata` previously opened a local file, read it into memory, and passed a `bytes.Reader` to the Lark SDK. The SDK only preserves multipart filenames when the form field is a concrete `*os.File`; anonymous readers fall back to `unknown-file`.

This change passes the opened local file directly to `Formdata.AddFile`, allowing the SDK to use `filepath.Base(f.Name())` as the multipart filename.

Fixes #729

## Tests

- `go test ./internal/cmdutil`
- `go test ./cmd/api ./cmd/service`
- `make build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file upload performance by streamlining data processing, reducing memory consumption during uploads and enabling better handling of larger files.

### DIFF
--- a/internal/cmdutil/fileupload.go
+++ b/internal/cmdutil/fileupload.go
@@ -112,12 +112,11 @@ func BuildFormdata(fileIO fileio.FileIO, fieldName, filePath string, isStdin boo
 		if err != nil {
 			return nil, output.ErrValidation("cannot open file: %s", filePath)
 		}
-		defer f.Close()
-		data, err := io.ReadAll(f)
-		if err != nil {
-			return nil, output.ErrValidation("--file: failed to read %s: %v", filePath, err)
-		}
-		fd.AddFile(fieldName, bytes.NewReader(data))
+		// Keep the concrete file reader instead of copying it into a
+		// bytes.Reader. The Lark SDK preserves multipart filenames only when
+		// it receives an *os.File; replacing it with an anonymous reader makes
+		// the SDK fall back to "unknown-file".
+		fd.AddFile(fieldName, f)
 	}
 
 	// Add top-level JSON keys as text form fields.
--- a/internal/cmdutil/fileupload_test.go
+++ b/internal/cmdutil/fileupload_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -276,6 +277,10 @@ func TestBuildFormdata(t *testing.T) {
 		if fd == nil {
 			t.Fatal("expected non-nil Formdata")
 		}
+
+		if got := formdataFieldTypeName(t, fd, "photo"); got != "*os.File" {
+			t.Fatalf("formdata file field type = %s, want *os.File", got)
+		}
 	})
 
 	t.Run("file not found", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


liangshuo-1:
Hi @rendylong, thanks for the fix and the feedback! 🙏

  We'd like to address this at the oapi-sdk-go layer instead — the SDK currently still reads the whole file into memory internally, so we'll take this chance to optimize it
  end-to-end. Give us a bit of time and we'll follow up here with progress.

PR #744, opened 2026-05-04
fix: handle negative truncate lengths

hiSandog:
## Summary
- return an empty string for non-positive truncate limits
- prevent negative lengths from panicking in truncate helpers
- add regression coverage for negative limits and ellipsis zero limits

## Validation
- env GOCACHE=/tmp/cli-gocache GOMODCACHE=/tmp/cli-gomodcache go test ./internal/util
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * String truncation now returns an empty string when given a non-positive limit, preventing unexpected output for zero or negative values.
* **Tests**
  * Unit tests updated to cover zero and negative truncation limits, ensuring consistent behavior across cases.

### DIFF
--- a/internal/util/strings.go
+++ b/internal/util/strings.go
@@ -5,6 +5,9 @@ package util
 
 // TruncateStr truncates s to at most n runes, safe for multi-byte (e.g. CJK) characters.
 func TruncateStr(s string, n int) string {
+	if n <= 0 {
+		return ""
+	}
 	r := []rune(s)
 	if len(r) <= n {
 		return s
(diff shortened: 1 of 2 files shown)


liuxinyanglxy:
Hi @hiSandog, the change itself looks fine to me. The two failing checks (deadcode and unit-test) don't seem related to your diff — likely caused by the branch being a few commits behind main. Could you rebase onto main? Should clear up after that. Thanks!

hiSandog:
Thanks for checking. I rebased this branch onto the latest main in ac6e49b and verified the touched package with `env GOCACHE=/tmp/cli-gocache GOMODCACHE=/tmp/cli-gomodcache go test ./internal/util`.

liuxinyanglxy (approved):

PR #750, opened 2026-05-06
feat(doc): guard v2 markdown against custom-tag corruption; warn on whole-paragraph style

herbertliu:
## Summary

Two new pre-flight checks for the v2 docs create/update path that prevent silent data loss and surface rendering pitfalls before content is sent to the API.

## Changes

- **`shortcuts/doc/docs_update_check.go`**: two new check functions
  - `CheckV2MarkdownCustomTags`: detects Lark custom tags (`<grid>`, `<column>`, `<lark-table>`, `<text color=...>`) in markdown-mode content and returns an error message. Called from `validateCreateV2` and `validateUpdateV2`.
  - `checkDocsUpdateWholeParagraphStyle`: warns when a line consists entirely of `*…*` or `**…**` markers (Lark renders these as literal characters, not styled paragraphs). Integrated into `docsUpdateWarnings`.
- **`shortcuts/doc/docs_create_v2.go`** / **`shortcuts/doc/docs_update_v2.go`**: call `CheckV2MarkdownCustomTags` in Validate when `--doc-format markdown` is set.
- **`shortcuts/doc/docs_update_check_test.go`**: 16 new table-driven tests (8 per check).

## Motivation

**Case 15** (`<grid>`/`<lark-table>` corruption): the v2 markdown parser does not understand Lark custom tags but attempts to process them — collapsing grid columns to 1, discarding lark-table content entirely, escaping `<text color=...>` to literal characters. This caused complete structural collapse of a weekly report document. Blocking early in Validate is the only reliable defense before MCP-layer support lands.

**Case 14** (whole-paragraph style markers): a line like `*为什么？*` on its own renders as the literal string `*为什么？*` in Lark, not as italic text. This is a different failure mode from the existing bold+italic check (#569) and is worth a dedicated warning.

## Test Plan

- [x] Unit tests pass (`go test ./shortcuts/doc/...`)
- [x] `go vet ./...` clean
- [x] `gofmt -l` clean

## Related

- Case 15: v2 markdown mode corrupts Lark custom tags (grid/column/lark-table/text)
- Case 14: whole-paragraph style markers not rendered by Lark

## Summary by CodeRabbit

* **New Features**
  * Markdown validation now detects unsupported v2 custom tags and surfaces errors during document create and update flows.
  * Added detection and warnings for whole‑paragraph Markdown style issues (e.g., entire-line bold/italic).

* **Tests**
  * Added tests covering custom‑tag detection, whole‑paragraph style warnings, code‑block/inline exceptions, and related edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/750)

### DIFF
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -27,6 +27,11 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
+	if runtime.Str("doc-format") == "markdown" {
+		if msg := CheckV2MarkdownCustomTags(runtime.Str("content")); msg != "" {
+			return common.FlagErrorf("%s", msg)
+		}
+	}
 	return nil
 }
 
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -28,7 +28,7 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
 	if runtime.Str("doc-format") == "markdown" {
-		if msg := CheckV2MarkdownCustomTags(runtime.Str("content")); msg != "" {
+		if msg := checkV2MarkdownCustomTags(runtime.Str("content")); msg != "" {
 			return common.FlagErrorf("%s", msg)
 		}
 	}
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -34,12 +34,20 @@ import (
 //     _**text**_   *__text__*
 //     Lark stores only one of the two emphases (usually italic), silently
 //     dropping the other. The user wanted both; they will get one.
+//
+//  3. Whole-paragraph style markers are not rendered by Lark. A line that
+//     consists entirely of *…* or **…** (no other content) is treated as a
+//     literal string rather than an italic or bold paragraph. The markers
+//     appear verbatim in the document instead of applying style.
 func docsUpdateWarnings(mode, markdown string) []string {
 	var warnings []string
 	if w := checkDocsUpdateReplaceMultilineMarkdown(mode, markdown); w != "" {
 		warnings = append(warnings, w)
 	}
 	if w := checkDocsUpdateBoldItalic(markdown); w != "" {
+		warnings = append(warnings, w)
+	}
+	if w := checkDocsUpdateWholeParagraphStyle(markdown); w != "" {
 		warnings = append(warnings, w)
 	}
 	return warnings
@@ -278,4 +286,70 @@ func leadingRun(s string, c byte) string {
 		i++
 	}
 	return s[:i]
+}
+
+// wholeParagraphStyleRe matches a line whose entire content is a single
+// emphasis span: *…* or **…** (at least one non-whitespace char inside).
+// CJK and ASCII content both match; we deliberately do NOT match ***…***
+// because that is already covered by checkDocsUpdateBoldItalic.
+var wholeParagraphStyleRe = regexp.MustCompile(`^\*{1,2}[^*\n]+\*{1,2}$`)
+
+// checkDocsUpdateWholeParagraphStyle warns when a markdown paragraph
+// consists entirely of *italic* or **bold** markers. Lark's markdown
+// parser treats such lines as literal strings (the markers appear verbatim)
+// rather than styled paragraphs. This is distinct from inline emphasis
+// embedded in mixed-content lines, which Lark handles correctly.
+func checkDocsUpdateWholeParagraphStyle(markdown string) string {
+	if markdown == "" {
+		return ""
+	}
+	sanitized := stripMarkdownCodeRegions(markdown)
+	for _, line := range strings.Split(sanitized, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if wholeParagraphStyleRe.MatchString(trimmed) {
+			return "line consisting entirely of *…* or **…** markers will not be rendered as italic/bold by Lark; " +
+				"the markers appear as literal characters. " +
+				"Mix the styled text with surrounding prose, or use Lark XML: <em>…</em> / <b>…</b>."
+		}
+	}
+	return ""
+}
+
+// v2MarkdownUnsupportedTags lists Lark custom XML tags that the v2
+// markdown parser does not understand. When present in markdown-mode
+// content, the parser silently strips or corrupts them — collapsing grid
+// columns, discarding lark-table rows, or escaping text-color tags to
+// literal characters. Use --doc-format xml to preserve these constructs.
+var v2MarkdownUnsupportedTags = []struct {
+	prefix  string
+	display string
+}{
+	{"<grid", "<grid>"},
+	{"<column", "<column>"},
+	{"<lark-table", "<lark-table>"},
+	{"<text color=", "<text color=...>"},
+	{"<text colour=", "<text colour=...>"},
+}
+
+// CheckV2MarkdownCustomTags returns a non-empty error message when content
+// contains Lark custom tags that the v2 markdown parser silently corrupts.
+// Callers in Validate should return this as an error; callers in DryRun
+// may choose to surface it as a warning instead.
+func CheckV2MarkdownCustomTags(content string) string {
+	if content == "" {
+		return ""
+	}
+	var found []string
+	for _, t := range v2MarkdownUnsupportedTags {
+		if strings.Contains(content, t.prefix) {
+			found = append(found, t.display)
+		}
+	}
+	if len(found) == 0 {
+		return ""
+	}
+	return "--doc-format markdown does not support Lark custom tags (" +
+		strings.Join(found, ", ") + "); " +
+		"the v2 API will silently strip or corrupt them. " +
+		"Switch to --doc-format xml (the default) to preserve these blocks."
 }
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -345,9 +345,14 @@ func CheckV2MarkdownCustomTags(content string) string {
 	if content == "" {
 		return ""
 	}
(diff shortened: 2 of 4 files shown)


shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,75 @@ func leadingRun(s string, c byte) string {
...
+		}
+	}
+	if len(found) == 0 {
+		return ""
+	}
+	return "--doc-format markdown does not support Lark custom tags (" +
+		strings.Join(found, ", ") + "); " +
+		"the v2 API will silently strip or corrupt them. " +
+		"Switch to --doc-format xml (the default) to preserve these blocks."
+}
```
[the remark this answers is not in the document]
herbertliu: Fixed in eed95ba: added `scanned := stripMarkdownCodeRegions(content)` before the tag scan loop, identical to the preprocessing used by `checkDocsUpdateWholeParagraphStyle`. Tags inside fenced code blocks and inline code spans no longer trigger the hard validation error. Added 4 regression test cases covering: grid in backtick fence, lark-table in fence, tag in inline code span, and tag outside fence still flagged when same tag appears inside fence.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
 	}
 	return s[:i]
 }
+
+// wholeParagraphStyleRe matches a line whose entire content is a single
+// emphasis span: *…* or **…** (at least one non-whitespace char inside).
+// CJK and ASCII content both match; we deliberately do NOT match ***…***
+// because that is already covered by checkDocsUpdateBoldItalic.
+// wholeParagraphStyleRe matches a line whose entire content is exactly one
```
fangshuyu-768, line 292: The comment block starting at line 291 and this one (lines 295–298) overlap significantly — both describe what `wholeParagraphStyleRe` matches and why `***…***` is excluded. Consider merging them into a single concise comment block to avoid redundancy and maintenance drift.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
...
+// *italic* or **bold** span. Requirements: opener and closer must be the same
+// number of asterisks (1 or 2), content has no * or newline, and at least one
+// non-whitespace non-asterisk character is present (prevents `*   *` matches).
+var wholeParagraphStyleRe = regexp.MustCompile(
+	`^(?:\*[^*\n]*[^\s*\n][^*\n]*\*|\*\*[^*\n]*[^\s*\n][^*\n]*\*\*)$`,
```
fangshuyu-768, line 309: The regex only matches `*italic*` / `**bold**` but not `_italic_` / `__bold__`. If Lark's markdown parser treats whole-paragraph underscore emphasis the same way (rendering markers verbatim), this check would miss those cases. Worth confirming Lark's behavior for underscore-style emphasis and adding coverage if needed.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
...
+// checkDocsUpdateWholeParagraphStyle warns when a markdown paragraph
+// consists entirely of *italic* or **bold** markers. Lark's markdown
+// parser treats such lines as literal strings (the markers appear verbatim)
+// rather than styled paragraphs. This is distinct from inline emphasis
+// embedded in mixed-content lines, which Lark handles correctly.
+func checkDocsUpdateWholeParagraphStyle(markdown string) string {
+	if markdown == "" {
+		return ""
+	}
+	sanitized := stripMarkdownCodeRegions(markdown)
```
fangshuyu-768, line 324: `stripMarkdownCodeRegions` handles fenced code blocks (```) and inline code, but 4-space indented code blocks in Markdown are not stripped. A line like `    *italic text*` (indented as a code block) could trigger a false-positive warning. Could you confirm whether `stripMarkdownCodeRegions` covers indented code blocks, and if not, whether that's an acceptable trade-off?

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
...
+	for _, line := range strings.Split(sanitized, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if wholeParagraphStyleRe.MatchString(trimmed) {
```
fangshuyu-768, line 328: The function returns on the first matching line, so if a document has multiple whole-paragraph emphasis lines, the user only sees one warning. Consider collecting all matching line numbers (or at least a count) and reporting them together, e.g. `"3 lines consisting entirely of *…* or **…** markers…"`. This gives users a clearer picture of how many lines need attention.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
...
+			return "line consisting entirely of *…* or **…** markers will not be rendered as italic/bold by Lark; " +
+				"the markers appear as literal characters. " +
+				"Mix the styled text with surrounding prose, or use Lark XML: <em>…</em> / <b>…</b>."
```
fangshuyu-768: The suggestion `"use Lark XML: <em>…</em> / <b>…</b>"` — could you confirm that `<em>` and `<b>` are valid Lark XML tags? If the actual Lark XML format uses different tag names (e.g. `<text>` with a `style` attribute), this suggestion would mislead users into trying unsupported syntax.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +287,80 @@ func leadingRun(s string, c byte) string {
...
+	{"<lark-table", "<lark-table>"},
+	{"<text color=", "<text color=...>"},
+	{"<text colour=", "<text colour=...>"},
+}
+
+// CheckV2MarkdownCustomTags returns a non-empty error message when content
+// contains Lark custom tags that the v2 markdown parser silently corrupts.
+// Callers in Validate should return this as an error; callers in DryRun
+// may choose to surface it as a warning instead.
+func CheckV2MarkdownCustomTags(content string) string {
```
fangshuyu-768: This function is only called from `docs_create_v2.go` and `docs_update_v2.go`, which are in the same package. Per Go convention, it doesn't need to be exported — `checkV2MarkdownCustomTags` would be consistent with the unexported `checkDocsUpdateWholeParagraphStyle` in the same file. Was there a reason to export it (e.g. planned external usage)?

shortcuts/doc/docs_create_v2.go
```diff
@@ -27,6 +27,11 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
+	if runtime.Str("doc-format") == "markdown" {
```
fangshuyu-768, line 30: Here the custom-tags check has no `content != ""` guard, while `docs_update_v2.go:106` adds `&& content != ""` before calling `CheckV2MarkdownCustomTags`. Although `CheckV2MarkdownCustomTags("")` safely returns `""`, the inconsistency between the two call sites is a bit confusing. Consider either adding the guard here for symmetry, or removing it in `docs_update_v2.go` since the function handles empty input.

fangshuyu-768 (commented):
Thanks for the PR! Overall the approach is solid — pre-flight checks to prevent silent data loss are a great idea. I have a few suggestions and questions below.

shortcuts/doc/docs_update_check.go
herbertliu: Fixed in f5f036d. Merged into a single concise block that states the shape once and documents both exclusions (underscore-style and ***…***).

shortcuts/doc/docs_update_check.go
herbertliu: Confirmed: Lark's markdown parser renders underscore-style whole-paragraph emphasis (_italic_ / __bold__) correctly — the markers are not shown as literals. The warning applies only to asterisk-style whole-paragraph lines, so underscore forms are correctly excluded from the regex. Added a comment in f5f036d documenting this decision.

shortcuts/doc/docs_update_check.go
herbertliu: Confirmed and documented in f5f036d. `stripMarkdownCodeRegions` does not strip 4-space indented code blocks (CommonMark §4.4). A line like `    *italic*` could theoretically trigger a false positive, but this is an accepted trade-off — a full CommonMark indented-block parser is not worth the code surface for content this check targets (AI-generated markdown). The limitation is now noted in the comment above `wholeParagraphStyleRe`.

shortcuts/doc/docs_update_check.go
herbertliu: Fixed in f5f036d. The function now collects all matching line numbers and reports them together, e.g. "3 lines consisting entirely of *…* or **…** markers…".

shortcuts/doc/docs_update_check.go
herbertliu: Good catch — updated in f5f036d. The suggestion now reads "switch to --doc-format xml" instead of suggesting specific XML tags, since the correct Lark XML representation for inline emphasis uses `<text>` with style attributes rather than `<em>`/`<b>`, and spelling that out accurately would be verbose.

shortcuts/doc/docs_update_check.go
herbertliu: Fixed in f5f036d. Renamed to `checkV2MarkdownCustomTags` (unexported). No external usage planned — the function is only ever called from the same package.

shortcuts/doc/docs_create_v2.go
herbertliu: Fixed in f5f036d. Removed the `&& content != ""` guard from `validateUpdateV2` to match `validateCreateV2` — both now rely on the callee's internal empty-string check.

SunPeiYang996:
Thanks for the detailed investigation and the test coverage here.

For the custom XML tag case, markdown is expected to support embedded custom XML tags such as `<grid>`, `<column>`, and related tags. The current behavior is a compatibility gap in the document API path, and we plan to handle the tag compatibility on the server side going forward. Because of that, I do not think the CLI should block this content locally.

For the whole-paragraph style warning, I am currently unable to reproduce the issue. The behavior looks correct based on the screenshot above, so I would prefer not to add this warning for now.

I am going to close this PR for now. If there is another reproducible case or a better approach, I would be happy to continue discussing it.

PR #784, opened 2026-05-08
docs: refine lark-base data analysis SOP

zgz2048:
## Summary
This PR refines the `lark-base` skill guidance based on failures observed in realistic, scaled Base data-analysis simulations. The main goal is to turn the previous record-read guidance into a higher-signal data-analysis execution contract that helps agents choose between record reads, filtered views, and `data-query` correctly.

## Why this matters
In the simulated evaluation tasks, agents could often access the right Base but still failed because they used unsafe query paths: reading a default page and extrapolating, using local `jq` over partial results, relying on an existing view without validating its filter/sort, or returning internal IDs instead of user-readable business values. These are not product API gaps; they are agent execution-policy gaps. This PR makes those failure modes explicit and gives agents a compact routing contract before they run Base queries.

## Changes
- Replace the older record-read SOP with `lark-base-data-analysis-sop.md`, covering data-query, filtered views, TopN, aggregation, relationship traversal, and post-query target validation.
- Add hard rules for global queries: avoid default-page conclusions, avoid local `jq`/shell statistics beyond small result sets, push filter/sort/aggregate work to Base, and validate existing views before relying on them.
- Clarify when to use `record-search`, filtered views, `record-list`, `record-get`, and `data-query`.
- Document the display-value rule: internal IDs, `record_id`, link IDs, open IDs, and code fields are connection keys, not final answers unless explicitly requested.
- Encourage preserving high-value reusable query views as persistent views and telling the user where to find them.
- Update `SKILL.md`, `lark-base-record.md`, and `lark-base-data-query.md` references to point at the new data-analysis SOP.

## Value
This should improve agent reliability on realistic Base/BI tasks where the user asks for business answers rather than raw records. It specifically reduces failures around partial-page reasoning, local-only aggregation, accidental view misuse, missing dimension-table lookups, and unclear boundaries between one-off analysis and reusable views.

## Test Plan
- [x] Manual local verification: `rg` confirms there are no remaining references to `lark-base-record-read-sop.md` or `lark-base-query-sop.md`.
- [x] Manual local verification: `git diff --check` passes.
- [ ] Unit tests pass — not run; documentation-only skill change.
- [ ] Manual local verification confirms the `lark xxx` command works as expected — not applicable; no CLI command behavior changed.

## Related Issues
- None

## Summary by CodeRabbit

* **Documentation**
  * Updated Base skill execution contract and guidance: added requirement to consult the data-query & analysis SOP for query tasks and detailed a pre-answer checklist.
  * Added unified SOP covering query selection, TopN/aggregation patterns, multi-table relationship handling, traceability, pagination/has_more and limit expectations.
  * Removed the old record-read SOP and consolidated its guidance into the new unified documentation.


kongenpei (dismissed):

skills/lark-base/references/lark-base-data-query.md
```diff
@@ -397,6 +416,17 @@ value 使用预定义关键字机制，第一个元素为字符串常量名称
    - 每个 value 是 CellValue 对象，实际值在 `value` 字段中，如 `{"value": "北京"}` 或 `{"value": 12345.00}`
    - 失败时结果在 `data.error` 中，包含具体错误码和信息
 
+## 与记录读取组合
+
+`+data-query` 不返回原始记录或 link 字段明细。需要最终输出业务实体名称、负责人、门店、供应商等明细时，按以下方式组合：
```
kongenpei: 这里过拟合了？

zgz2048:
Addressed the review feedback in commit `84c60e7`.

I changed the `+data-query` follow-up guidance so `+record-search` is no longer listed as an equivalent precise lookup path:

- known `record_id` -> `+record-get`
- structured business keys such as IDs/status/date/amount -> temporary filtered view + `+record-list --view-id`
- `+record-search` only when the candidate condition is a text display-value keyword, with `search_fields` and `select_fields`

This keeps text search distinct from structured exact filtering.

skills/lark-base/references/lark-base-data-analysis-sop.md
```diff
@@ -0,0 +1,88 @@
...
+- link 单元格通常是关联表 `record_id` 数组，不是用户可读名称。
+- 先用 `+field-list` 确认 link 字段的 `link_table`、业务唯一键和展示字段。
+- 从驱动表拿到候选记录后，用关联 `record_id` 到关联表 `+record-get` 批量读取展示字段。
+- 多跳关系逐跳建立 `record_id/key -> 展示字段` 映射；最终输出业务名称或用户要求字段。
+
+禁止：
+
+- 把 link `record_id` 当最终答案。
+- 用 `+record-search` 搜 link `record_id`。
+- 凭 ID 猜名称、负责人、门店、供应商。
```
kongenpei, line 68: 同上

kongenpei (approved):

PR #787, opened 2026-05-08
feat(mail): HTML lint lib + Larksuite-native autofix + lark-mail skill

bubbmon233:
## Summary

Adds an HTML lint library + Larksuite-native autofix to `lark-cli mail`, plus the `skills/lark-mail/` skill bundle (2 reference docs, 5 HTML templates, the `+lint-html` shortcut, and writing-path lint integration across all 6 compose shortcuts).

### 1. Lint library (`shortcuts/mail/lint/`)

3-tier rule set:

- **Error**: drop dangerous tags (`<script>` / `<iframe>` / `<form>` / `<input>` / `<link>` / `<object>` / `<embed>`), `on*` event handlers, and `javascript:` / `vbscript:` / `file:` URLs.
- **Warning + autofix**: rewrite HTML4-era tags (`<font>` / `<center>` / `<marquee>` / `<blink>`).
- **Larksuite-native autofix**: rewrite `<p>` / `<ul>` / `<ol>` / `<li>` / `<blockquote>` / `<a>` to mail-editor native markup so AI can write the simplest HTML and still produce native-quality rendering.
- Inline-style and URL-scheme allow-list filtering.
- `<style>` block passthrough (server adds CSS scope class).

### 2. `+lint-html` shortcut (preview / CI)

Read-only HTML preview tool. Default envelope returns only `cleaned_html`; `--show-lint-details` adds full `warnings[]` / `errors[]`. `--strict` exits non-zero on any finding (CI gate).

### 3. Writing-path lint in the 6 compose shortcuts

`+send` / `+draft-create` / `+reply` / `+reply-all` / `+forward` / `+draft-edit` body op all run lint before drafting:

- `lint_applied_count` / `original_blocked_count` — always present.
- `lint_applied[]` / `original_blocked[]` — only with `--show-lint-details`.
- `compose_hint` — points AI consumers to the HTML writing guide.

### 4. `skills/lark-mail/` skill bundle

- 5 pre-rendered Larksuite-native HTML templates: weekly newsletter, personal weekly report, team weekly report, market research report, résumé.
- 2 reference docs:
  - `references/lark-mail-html.md` — writing rules + format primitives + template-usage flow.
  - `references/lark-mail-lint-html.md` — `+lint-html` usage + return-value contract + 9 examples.
- `SKILL.md` updates linking the new docs and templates.

### 5. Sealed conventions

Fixed writing conventions enforced by the lint library, the Larksuite mail-editor data model, or the upstream service-side sanitiser.

- **@user mention chip** — `id="at-user-N"` is the only hard requirement; do not write `data-user-id`.
- **Highlight palette** — 3 colors (pink milestones, yellow follow-ups, green completed); black text, no bold / padding / border-radius.
- **Brand color palette** — main black, 3 levels of grey, Lark blue / deep blue, alert red, emergency orange, light pink / light grey backgrounds, border grey.
- **URL scheme allow-list** — `http(s):` / `mailto:` / `cid:` / `data:image/*` only.
- **Inline style allow-list** — font-* / color / background-color / text-* / line-height / letter-spacing / vertical-align / margin* / padding* / width / height / display / border* / list-style* / white-space / word-break / overflow / transition / cursor / opacity.
- **Tag allow-list** — `<p>` / `<div>` / `<span>` / `<a>` / `` / `<table>` (with `<thead>` / `<tbody>` / `<tfoot>` / `<tr>` / `<td>` / `<th>` / `<caption>` / `<colgroup>` / `<col>`) / `<ul>` / `<ol>` / `<li>` / `<blockquote>` / `<h1>`-`<h6>` / `<b>` / `<i>` / `<em>` / `<strong>` / `<u>` / `<s>` / `<sub>` / `<sup>` / `<pre>` / `<code>` / `<style>`.
- **Writing-style floor** — subject ≤ 50 chars; decision-first; lists instead of "一、二、三" / "①②③"; emoji only as status tags; greeting / sign-off ≤ 1 paragraph each.

## Tests

- `shortcuts/mail/lint/...` — unit tests for every rule.
- `shortcuts/mail/mail_lint_html_test.go` — `+lint-html` envelope contract.
- `shortcuts/mail/mail_lint_writepath_test.go` — writing-path envelope contract.
- 5 templates verified via `+draft-create` smoke test.

## Test plan

- [x] `go test ./shortcuts/mail/lint/... ./shortcuts/mail/...`
- [x] All 5 templates render correctly via `+draft-create` smoke
- [x] Default vs `--show-lint-details` envelope verified for both `+lint-html` and `+draft-create`

## Summary by CodeRabbit

* **New Features**
  * Added `+lint-html` shortcut for local HTML email linting (auto-fix by default; strict mode returns non-zero on findings).
  * Compose shortcuts now run automatic HTML sanitization during send/draft flows and accept `--show-lint-details` to expose findings.
  * `--auto-fix` / `--strict` options supported; cleaned HTML returned when auto-fix enabled.

* **Documentation**
  * New comprehensive email HTML/CSS writing guidelines and `+lint-html` usage guide.

* **Templates**
  * Added multiple HTML email templates (job application, newsletters, reports, weekly reports).

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/787)


shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+// preserves user intent for niche tags (e.g. `<details>` / `<summary>`) that
+// browsers + Feishu native renderer already handle.
+func classifyTag(tag string) (kind, ruleID string) {
+	tag = strings.ToLower(tag)
+	if allowedTags[tag] {
+		return "allow", ""
+	}
+	if id, ok := blockedTags[tag]; ok {
+		return "error", id
+	}
```
chanthuang, line 161: **Issue:** `<svg>` and `<math>` tags are not in `blockedTags` and fall through to `return "allow", ""`. Both are well-known XSS vectors:

- `<svg onload="alert(1)">` — the `onload` handler gets stripped by `processAttributes`, but `<svg><foreignObject><body onload="...">` may bypass this depending on how `golang.org/x/net/html` handles SVG namespace switching.
- `<math><mtext><table><mglyph><style><!--</style>` — this is a classic mXSS (mutation XSS) pattern exploiting parser-context differences between HTML and MathML.
- `<svg><script>alert(1)</script></svg>` — the inner `<script>` is caught by the walker, but the SVG element itself survives and could carry other active content.

Neither `<svg>` nor `<math>` are in the PR description's tag allow-list, so allowing them by default appears unintentional.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Add `<svg>`, `<math>`, and their namespace-sensitive children (`<foreignObject>`, `<mtext>`, `<mglyph>`, `<annotation-xml>`) to `blockedTags`. If SVG inline images are needed, they should go through `` (which is already allowed, though see the separate comment on SVG data URIs).

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+	case allowedURLSchemes[scheme]:
+		return "ok", ""
+	case scheme == "data":
+		// data:image/* is whitelisted; anything else (e.g. data:text/html;...)
+		// is rejected. The check tolerates any subtype under image/* (png /
+		// jpeg / gif / svg+xml / webp) so users embedding base64 thumbnails
+		// don't trip the rule.
+		rest = strings.TrimSpace(rest)
+		if strings.HasPrefix(strings.ToLower(rest), "image/") {
+			return "ok", ""
```
chanthuang, line 237: **Issue:** `data:image/svg+xml` URIs pass the allow-list because the check only verifies `strings.HasPrefix(rest, "image/")`. SVG data URIs can contain embedded `<script>` tags, event handlers, and `<foreignObject>` — they are a well-documented XSS vector (OWASP, PortSwigger). When rendered in an `` tag, most modern browsers sandbox SVG and block script execution, but when used in CSS `background-image` or other contexts, the behavior is less predictable.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Either block `data:image/svg+xml` specifically:
```go
if strings.HasPrefix(strings.ToLower(rest), "image/") {
    if strings.HasPrefix(strings.ToLower(rest), "image/svg") {
        return "error", RuleAttrDataURLBlocked
    }
    return "ok", ""
}
```
Or, more conservatively, restrict data URIs to raster formats only (`image/png`, `image/jpeg`, `image/gif`, `image/webp`).

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+	// scheme (e.g. "java\tscript:..."). The html-parser already strips
+	// stray whitespace at attribute boundaries; this is defence-in-depth
+	// for older clients that paste from Word with U+0009 / U+0020 inside
+	// the scheme prefix.
+	value = strings.Map(func(r rune) rune {
+		if r < 0x20 || r == 0x7F {
+			return -1
+		}
+		return r
+	}, value)
```
chanthuang, line 216: **Issue:** The control-byte stripping only removes chars `< 0x20` and `0x7F`, but does not strip Unicode zero-width / invisible characters that can obfuscate URL schemes:

- `java\u200Bscript:` (zero-width space U+200B inside "javascript") → extracted scheme is `"java\u200Bscript"` which does NOT match `blockedURLSchemes["javascript"]`, so it falls to the default `"warn"` case instead of `"error"`.
- Similar bypasses via U+200C (ZWNJ), U+200D (ZWJ), U+FEFF (BOM), U+00A0 (NBSP), U+2028/U+2029 (line/paragraph separator), or bidi override characters (U+202A-U+202E).

Whether downstream renderers actually execute `java\u200Bscript:` depends on the mail client, but this is a defense-in-depth gap — the existing test `TestRun_WhitespaceObfuscatedJavaScriptScheme` only covers ASCII whitespace (`\t`, `\n`), not Unicode invisible chars.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Extend the `strings.Map` filter to also strip Unicode categories Cf (format chars), Zs (space separators), and Zl/Zp (line/paragraph separators):
```go
import "unicode"

value = strings.Map(func(r rune) rune {
    if r < 0x20 || r == 0x7F {
        return -1
    }
    if unicode.Is(unicode.Cf, r) || unicode.Is(unicode.Zs, r) ||
       unicode.Is(unicode.Zl, r) || unicode.Is(unicode.Zp, r) {
        return -1
    }
    return r
}, value)
```

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+		// data:image/* is whitelisted; anything else (e.g. data:text/html;...)
+		// is rejected. The check tolerates any subtype under image/* (png /
+		// jpeg / gif / svg+xml / webp) so users embedding base64 thumbnails
+		// don't trip the rule.
+		rest = strings.TrimSpace(rest)
+		if strings.HasPrefix(strings.ToLower(rest), "image/") {
+			return "ok", ""
+		}
+		return "error", RuleAttrJSURLBlocked
+	case blockedURLSchemes[scheme]:
```
chanthuang, line 240: **Issue:** When a `data:text/html` or `file:` URL is blocked, the rule ID is `RuleAttrJSURLBlocked` ("ATTR_JS_URL_BLOCKED"). This is semantically incorrect — a `data:text/html` URI is not a JavaScript URL, and `file:` is a local file access scheme. The finding's `RuleID` is part of the public envelope contract (per the S2 contract reference in the package comment), so consumers pattern-matching on rule IDs will misclassify these.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Introduce distinct rule IDs:
```go
RuleAttrDataURLBlocked = "ATTR_DATA_URL_BLOCKED"
RuleAttrFileURLBlocked = "ATTR_FILE_URL_BLOCKED"
```
And use them in the respective `switch` cases.

shortcuts/mail/lint/linter.go
```diff
@@ -0,0 +1,1065 @@
...
+				Hint:      hintForWarnTag(tagName),
+			}
+			if opts.Strict {
+				finding.Severity = SeverityError
+				rep.Blocked = append(rep.Blocked, finding)
+			} else {
+				rep.Applied = append(rep.Applied, finding)
+			}
+			rewriteWarnTag(n, tagName)
+			// Recurse into the rewritten node by falling through; the
```
chanthuang: **Issue:** When `opts.Strict = true` AND `opts.AutoFix = true`, the finding is classified as `SeverityError` and added to `rep.Blocked`, but `rewriteWarnTag` is still called — the tag is **rewritten** (e.g., `<font>` → `<span>`), not **removed**. This is inconsistent with how true error-tier tags are handled (line 131: `parent.RemoveChild(n)`).

In strict mode, all findings are errors. A `<font>` tag classified as error should arguably be removed entirely (like `<script>`), not rewritten to `<span>`. The current behavior means strict mode reports it as an error but still produces a "fixed" output, which sends mixed signals to CI consumers using `--strict`.

**Severity:** 🚨 Should Fix

**Suggested Fix:** When `opts.Strict` is true, skip the rewrite and remove the tag's children (re-parent them) or remove the tag entirely, consistent with the error-tier treatment. Or, if the current behavior is intentional (strict = "report as error but still fix"), document this explicitly in the `Options.Strict` godoc and add a test asserting that strict-promoted warnings are rewritten, not removed.

skills/lark-mail/assets/templates/research--market-report.html
```diff
@@ -0,0 +1,270 @@
+<!--
+=============================================================================
+SUBJECT 模板（lark-cli mail --subject 用）：
+  [调研主题] 市场调研报告 ([YYYY-MM-DD])
+字段说明：
+  · [调研主题]：调研对象赛道，例 "AI Mail Agent" / "向量数据库" / "前端构建工具"
+  · [YYYY-MM-DD]：调研完成日期（ISO 格式）
+=============================================================================
+-->
+<style>
```
chanthuang, line 10: **Issue:** This template relies entirely on a `<style>` block with CSS classes (`display:flex`, `linear-gradient`, `box-shadow`, `gap`, `flex-wrap`, etc.). Two problems:

1. **Doc inconsistency:** `skill-template/domains/mail.md` line 288 states `<style>` blocks "会被直接删除" (will be deleted), but the linter actually passes `<style>` through (it's not in `blockedTags`). Either the doc or the code is wrong.
2. **Server-side risk:** If the Feishu mail server-side sanitizer strips `<style>` blocks (as the doc claims), this template degrades to unstyled HTML — all CSS classes lose their definitions, `display:flex` layouts break, and the card/gradient design disappears. All other templates correctly use only inline styles.
3. **CSS properties outside allow-list:** `box-shadow`, `flex`, `gap`, `linear-gradient`, `flex-wrap` are NOT in `allowedStyleProps` — they only work because they're in a `<style>` block (bypassing inline-style filtering), which makes this template fragile.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Rewrite this template to use inline styles only (like the other 4 templates), or explicitly add `<style>` to `allowedTags` and update `mail.md` to document that `<style>` is allowed. If `<style>` is intentionally passed through (as the linter code suggests), the domain doc must be corrected.

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+		// renders eventually support them.
+		return "warn", RuleAttrUnsafeSchemeBlocked
+	}
+}
+
+// urlAttributes lists attributes whose value is a URL and must therefore
+// pass classifyURLValue. Lower-case canonical names.
+var urlAttributes = map[string]bool{
+	"href":       true,
+	"src":        true,
```
chanthuang, line 254: **Severity:** 💡 Suggestion

**Issue:** `srcset` is missing from the URL attribute list. Since `` passes through `allowedTags`, an `` would not have its `srcset` value checked by `classifyURLValue`. While `srcset` has a more complex value syntax (comma-separated URL + descriptor pairs), the `javascript:` prefix would still be present in the raw value.

Also, the PR description's "Inline style allow-list" mentions `letter-spacing` but it's not in `allowedStyleProps` — minor doc/code discrepancy.

**Suggested Fix:** Add `"srcset": true` to `urlAttributes`. For `letter-spacing`, either add it to `allowedStyleProps` or remove it from the PR description's allow-list.

shortcuts/mail/mail_lint_writepath_test.go
```diff
@@ -0,0 +1,352 @@
...
+	if ob == nil {
+		t.Error("original_blocked is nil — must be empty slice in detail mode")
+	}
+}
+
+// =====================================================================
+// End-to-end: +draft-create writing path emits envelope with lint fields.
+// =====================================================================
+
+// TestMailDraftCreate_WritePathLintEnvelopeDefault verifies +draft-create's
```
chanthuang, line 126: **Severity:** 💡 Suggestion

**Issue:** The writepath E2E tests only cover `+draft-create`. The other 5 compose shortcuts (`+send`, `+reply`, `+reply-all`, `+forward`, `+draft-edit`) were all wired up with lint integration in this PR but lack dedicated writepath tests. While the shared helper functions (`runWritePathLint`, `applyLintToEnvelope`) are tested in isolation, the integration wiring in each shortcut's `Execute` function (flag registration, correct placement in the output pipeline, both draft/send branches) is not verified.

**Suggested Fix:** Add at least one E2E test per shortcut confirming the lint envelope fields appear in the output. For shortcuts with two branches (`+forward`/`+reply`/`+reply-all`/`+send` — draft-saved vs. sent), test both branches.

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+	"text-indent":      true,
+	// Quote-block / native Feishu styles spec §4.4 "通过" (the editor-kit
+	// branch keeps them). Whitespace + word-break are part of the existing
+	// `<pre>` / quote wrapper styles in mail_quote.go (e.g. `bodyDivStyle`).
+	"white-space":     true,
+	"word-break":      true,
+	"word-wrap":       true,
+	"overflow":        true,
+	"overflow-wrap":   true,
+	"vertical-align":  true,
```
chanthuang, line 293: **Issue:** From an email security perspective, the combination of allowed CSS properties enables several classic email abuse patterns that enterprise email gateways (Proofpoint, Mimecast, Barracuda) actively flag:

1. **Invisible phishing overlay**: `opacity:0` on an `<a>` tag creates a clickable but invisible link. Combined with `cursor:pointer`, the user clicks what they think is one element but actually hits a hidden link underneath.
2. **Hidden content**: `display:none` hides content from visual rendering while it remains in the DOM — used for spam keyword stuffing, hidden tracking URLs, or payload hiding.
3. **Zero-size text**: `font-size:0` or `font-size:1px` makes text invisible — a classic spam technique to embed keywords or hidden URLs.
4. **Camouflaged text**: `color` can be set identical to `background-color` (e.g., white-on-white), hiding text from readers but not from the DOM.
5. **Tracking pixel construction**: `width:0;height:0` or `width:1px;height:1px` on `` with an external `src` creates a tracking pixel — the lint passes this through without warning.

These are not theoretical — they are documented in RFC 8058 (One-Click Unsubscribe), M3AAWG best practices, and OWASP email security guidelines.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Add lint warnings (not errors — these have legitimate uses) for suspicious CSS patterns:
- `opacity:0` or `opacity` values very close to 0
- `display:none`
- `font-size:0` or `font-size:1px`
- `` with `width` or `height` <= 2px (tracking pixel heuristic)

This doesn't require blocking — just surfacing them as `SeverityWarning` findings so consumers and users are aware.

shortcuts/mail/lint/linter.go
```diff
@@ -0,0 +1,1065 @@
...
+//
+// Other attributes pass through unchanged. The cli's existing
+// `validateInlineCIDs` (helpers.go:2226) handles `cid:`-specific checks; the
+// lint must not duplicate that responsibility (S2 contract «MUST reuse» row).
+func processAttributes(n *xhtml.Node, rep *Report, opts Options) {
+	keep := n.Attr[:0]
+	for _, attr := range n.Attr {
+		name := strings.ToLower(attr.Key)
+
+		// 1. on*-handlers → always drop, error-tier.
```
chanthuang, line 165: **Issue:** The attribute scanner does not check for **link text vs. href mismatch** — one of the most common phishing patterns in email. Example:

```html
<a href="https://evil-site.com">https://your-company.com/dashboard</a>
```

The visible text looks like a legitimate internal URL, but the actual `href` points elsewhere. Enterprise email security gateways (Microsoft Defender for Office 365, Google Safe Browsing, Proofpoint URL Defense) specifically flag this pattern. Since this lint library aims to be a "safety floor" for the email writing path, detecting obvious URL-text mismatches would significantly strengthen its anti-phishing value.

**Severity:** 💡 Suggestion

**Suggested Fix:** When processing `<a>` tags, if the text content of the anchor looks like a URL (starts with `http://` or `https://`), compare its domain with the `href` domain. If they differ, emit a warning finding like `LINK_TEXT_HREF_DOMAIN_MISMATCH`. This is a heuristic — it won't catch all phishing — but it catches the most common pattern with very low false-positive rate.

skills/lark-mail/assets/templates/research--market-report.html
```diff
@@ -0,0 +1,270 @@
...
+.research-root { font-family:-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif; max-width:800px; margin:0 auto; color:#1a1a1a; line-height:1.6; background-color:#f8f9fa; padding:20px; }
+.gradient-header { background:linear-gradient(135deg, #1a73e8, #4285f4); border-radius:12px; padding:32px; color:white; text-align:center; }
+.card { background-color:white; border-radius:8px; padding:20px; margin:16px 0; box-shadow:0 1px 3px rgba(0,0,0,0.1); }
+.stat-row { display:flex; gap:10px; margin:16px 0; }
+.stat-card { flex:1; background-color:white; border-radius:8px; padding:14px; text-align:center; box-shadow:0 1px 3px rgba(0,0,0,0.1); }
```
chanthuang, line 15: **Issue:** `display:flex` and `gap` have **zero support** in Outlook (all versions — Outlook uses the Word HTML rendering engine, not a browser engine). This template's card-grid layout will collapse into a vertical stack with no gap spacing when opened in Outlook, which is the dominant email client in enterprise environments.

Specific incompatibilities in this template:
| CSS Property | Outlook | Gmail (web) | Apple Mail | Yahoo |
|---|---|---|---|---|
| `display:flex` | ❌ | ⚠️ partial | ✅ | ❌ |
| `gap` | ❌ | ❌ | ✅ | ❌ |
| `linear-gradient` | ❌ | ❌ | ✅ | ⚠️ |
| `box-shadow` | ❌ | ✅ | ✅ | ✅ |
| `border-radius` | ❌ (Windows) | ✅ | ✅ | ✅ |
| `max-width` on div | ❌ | ✅ | ✅ | ✅ |

The email industry standard for multi-column layouts is `<table>` with inline styles — it's ugly to write but renders consistently across all clients. The other 4 templates in this PR correctly avoid these properties.

**Severity:** 🚨 Should Fix

**Suggested Fix:** Rewrite the card-grid sections (`.stat-row`, `.player-row`) using `<table>` with `<td>` for each card. Replace `linear-gradient` with a solid color fallback. Move all class-based styles to inline `style` attributes. Use [Can I Email](https://www.caniemail.com/) as the compatibility reference.

shortcuts/mail/lint/rules.go
```diff
@@ -0,0 +1,349 @@
...
+	"h3":         true,
+	"h4":         true,
+	"h5":         true,
+	"h6":         true,
+	"sub":        true,
+	"sup":        true,
+	"section":    true,
+	"article":    true,
+	"header":     true,
+	"footer":     true,
```
chanthuang, line 101: **Issue:** Semantic HTML5 tags (`<section>`, `<article>`, `<header>`, `<footer>`, `<nav>`, `<main>`, `<figure>`, `<figcaption>`) are in `allowedTags`, but email clients handle them very inconsistently:

- **Outlook** (Word engine): renders them as inline spans with no block behavior — layout completely breaks.
- **Gmail**: strips `<article>`, `<nav>`, `<main>` entirely; keeps `<section>`/`<header>`/`<footer>` but removes their semantic meaning.
- **Yahoo Mail**: strips most semantic tags.
- **AOL**: strips all HTML5 semantic tags.

Email HTML is effectively a subset of HTML4 + CSS2.1. The industry best practice (per Litmus, Email on Acid, MJML framework) is to use only `<div>`, `<span>`, `<table>`, `<p>`, `<h1>`-`<h6>`, and formatting tags. Including semantic tags in the allow-list creates a false sense of safety — authors will use them thinking they're "allowed," but the rendered result in major email clients will be broken or stripped.

**Severity:** 💡 Suggestion

**Suggested Fix:** Either move `<section>`, `<article>`, `<header>`, `<footer>`, `<nav>`, `<main>`, `<figure>`, `<figcaption>` to `warnAutofixTags` (rewrite to `<div>`) or add a lint hint when these tags are used, noting that they have poor email client support. The reference doc `lark-mail-html.md` already doesn't mention them in any examples — implicitly acknowledging they shouldn't be used.

shortcuts/mail/lint/linter.go
```diff
@@ -0,0 +1,1065 @@
...
+//                     </div>
+//                     This is the only rewrite that changes tree shape; the
+//                     others only touch attributes.
+func applyFeishuNativeStyles(parent *xhtml.Node, rep *Report, opts Options) {
+	if !opts.AutoFix {
+		return
+	}
+	child := parent.FirstChild
+	for child != nil {
+		next := child.NextSibling
```
chanthuang, line 528: **Issue:** The `applyFeishuNativeStyles` pass adds native inline styles to `` tags — but the lint does not enforce `alt` attribute presence on ``. Missing `alt` text is both an **accessibility failure** (WCAG 2.1 Level A, criterion 1.1.1) and a **practical email UX problem** — most email clients block external images by default and display alt text as a fallback. An `` without `alt` shows as a blank rectangle or broken icon until the recipient manually loads images.

This is not a niche concern: Google Workspace, Outlook, and Apple Mail all block images by default for external senders. Enterprise security policies often enforce image blocking. Alt text is the only way recipients see *anything* until they click "load images."

**Severity:** 💡 Suggestion

**Suggested Fix:** In `processAttributes` or as a post-pass, emit a `SeverityWarning` finding when an `` tag lacks an `alt` attribute (or has `alt=""`). Rule ID suggestion: `IMG_MISSING_ALT_TEXT`.

skills/lark-mail/references/lark-mail-html.md
```diff
@@ -0,0 +1,311 @@
...
+
+**CRITICAL 邮件是重要的对外交流渠道，请你保证书写语言凝练扼要**
+**CRITICAL 电子邮件的 HTML 不是 Web 开发的 HTML，请你务必遵守本文档中提及的常用邮件格式书写规范**
+**CRITICAL 请务必使用 shortcut 来进行邮件内容编辑 （`+send` / `+draft-create` / `+reply` / `+reply-all` / `+forward`）或 `+draft-edit` 的 body op，严禁自行拼接 EML**
+
+你可以参考 **官方模板库** [`../assets/templates/`](../assets/templates) — 提供部分场景模板，可供参考
+
+> 请注意，邮件内容编辑相关的 shortcut 内置 HTML lint 工具，处于安全考虑和格式适配，你输入的 HTML 可能会被自动调整
+
+## 风格底线
```
chanthuang, line 13: **Issue:** Two inconsistencies between this doc and `skill-template/domains/mail.md`:

1. **Subject line length**: This doc says "邮件标题小于50字" (line 13), but `mail.md` says "标题 ≤ 30 字" (line 302). These are contradictory. Industry standard is typically 30-50 chars depending on mobile vs. desktop — but the two docs in the same PR must agree.

2. **`opacity` example** (line 209 of this doc): The writing guide includes `<span style="opacity:0.5">半透明文字</span>` as a recommended pattern. While semi-transparent text is fine, showing `opacity` as a first-class recommended feature invites misuse (`opacity:0` for invisible content). Consider whether this example belongs in an email writing guide, given the phishing risk described in the separate security comment.

**Severity:** 🚨 Should Fix (for the inconsistency), 💡 Suggestion (for the opacity example)

**Suggested Fix:** Align both documents on one subject-length limit. If 50 is the intent, update `mail.md`; if 30, update this doc. For the opacity example, consider adding a note that `opacity:0` will trigger a lint warning (if the invisible-content detection suggested in the security comment is implemented).

PR #822, opened 2026-05-11
feat(doc): add v2 XML content guards for bare ampersands and deprecated tags

herbertliu:
## Summary

Add pre-flight static checks for the v2 XML document path to catch two categories of silent failures before the API call is made.

## Changes

- **`shortcuts/doc/docs_update_check.go`**: Add `CheckV2XMLBareAmpersand` (hard error) and `CheckV2XMLWarnings` (non-fatal warnings) with table-driven tests
- **`shortcuts/doc/docs_create_v2.go`**: Integrate bare-ampersand check in `validateCreateV2` and XML warnings in `executeCreateV2`
- **`shortcuts/doc/docs_update_v2.go`**: Same integration in `validateUpdateV2` / `executeUpdateV2`
- **`shortcuts/doc/docs_update_check_test.go`**: Add `TestCheckV2XMLBareAmpersand` and `TestCheckV2XMLWarnings`

## Checks

**`CheckV2XMLBareAmpersand`** — hard error (returned from Validate):
- Fires when content contains a `&` that is not a recognised XML entity (`&amp;`, `&lt;`, `&gt;`, `&apos;`, `&quot;`, `&#N;`, `&#xH;`).
- The v2 XML parser rejects such requests outright; catching this early gives a clear error instead of an opaque API failure.

**`CheckV2XMLWarnings`** — non-fatal warnings (printed to stderr before the API call):
1. `<quote-container>` — v2 silently drops the block; the correct tag is `<blockquote>`.
2. `<column width="N">` with an integer value — has no effect in v2; the correct attribute is `width-ratio="0.5"` (float 0–1).

Both checks only fire when `--doc-format` is `xml` (the default).

## Test Plan

- [x] `go test ./shortcuts/doc/...` — all pass, 66.4% coverage
- [x] `go vet ./...` — clean
- [x] `gofmt -l .` — clean

## Summary by CodeRabbit

* **New Features**
  * Enhanced v2 document validation: submissions (when not using markdown) are blocked for malformed bare ampersands.
  * Emits user-facing warnings to stderr for unsupported XML constructs (e.g., quote-container) and integer column-width attributes before processing.

* **Tests**
  * Added unit tests covering bare-ampersand detection and XML warning conditions.
  * Added CLI dry-run end-to-end tests verifying create/update guard and warning behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/822?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -28,7 +28,7 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
-	if runtime.Str("doc-format") != "markdown" && runtime.Str("content") != "" {
+	if runtime.Str("doc-format") != "markdown" {
 		if msg := CheckV2XMLBareAmpersand(runtime.Str("content")); msg != "" {
 			return common.FlagErrorf("%s", msg)
 		}
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -5,6 +5,7 @@ package doc
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/larksuite/cli/shortcuts/common"
@@ -27,6 +28,11 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
+	if runtime.Str("doc-format") != "markdown" {
+		if msg := CheckV2XMLBareAmpersand(runtime.Str("content")); msg != "" {
+			return common.FlagErrorf("%s", msg)
+		}
+	}
 	return nil
 }
 
@@ -44,6 +50,12 @@ func dryRunCreateV2(_ context.Context, runtime *common.RuntimeContext) *common.D
 
 func executeCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	body := buildCreateBody(runtime)
+
+	if runtime.Str("doc-format") != "markdown" {
+		for _, w := range CheckV2XMLWarnings(runtime.Str("content")) {
+			fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", w)
+		}
+	}
 
 	data, err := doDocAPI(runtime, "POST", "/open-apis/docs_ai/v1/documents", body)
 	if err != nil {
--- a/shortcuts/doc/docs_create_v2.go
+++ b/shortcuts/doc/docs_create_v2.go
@@ -28,7 +28,7 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
-	if runtime.Str("doc-format") != "markdown" {
+	if runtime.Str("doc-format") != "markdown" && runtime.Str("content") != "" {
 		if msg := CheckV2XMLBareAmpersand(runtime.Str("content")); msg != "" {
 			return common.FlagErrorf("%s", msg)
 		}
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -317,13 +317,20 @@ func CheckV2XMLBareAmpersand(content string) string {
 // hypothetical attributes or element names that start with "quote-container".
 var quoteContainerTagRe = regexp.MustCompile(`<quote-container(?:\s|>|/)`)
 
-// columnIntWidthRe matches a <column … width="N" …> attribute where N is a
-// plain integer (not a float). The pattern requires:
-//   - whitespace before "width" to exclude attributes like data-width
-//   - the value to be enclosed in quotes with digits immediately before the
-//     closing quote, so width="0.5" does NOT match (the dot prevents \d+
-//     from consuming the full value up to the quote).
-var columnIntWidthRe = regexp.MustCompile(`<column\b[^>]*\swidth\s*=\s*["']\d+["']`)
+// columnWidthAttrRe matches a <column … width="…" …> attribute, capturing the
+// opening quote, the numeric value, and the closing quote. Go's RE2 does not
+// support backreferences, so we verify opening == closing quote in Go code.
+var columnWidthAttrRe = regexp.MustCompile(`<column\b[^>]*\swidth\s*=\s*(['"])([0-9.]+)(['"])`)
+
+// isIntWidth reports whether s is a plain integer with no decimal point.
+func isIntWidth(s string) bool {
+	for _, c := range s {
+		if c == '.' {
+			return false
+		}
+	}
+	return true
+}
 
 // CheckV2XMLWarnings returns a list of non-fatal warnings for v2 XML content.
 // These describe constructs that are silently dropped or ignored by the v2 API
@@ -347,10 +354,13 @@ func CheckV2XMLWarnings(content string) []string {
 			"<quote-container> is not supported in v2 XML and will be silently dropped; "+
 				"use <blockquote> instead.")
 	}
-	if columnIntWidthRe.MatchString(content) {
-		warnings = append(warnings,
-			"<column width=\"N\"> with an integer value has no effect in v2 XML; "+
-				"use width-ratio=\"0.5\" (float 0–1) to set column width.")
+	for _, m := range columnWidthAttrRe.FindAllStringSubmatch(content, -1) {
+		if m[1] == m[3] && isIntWidth(m[2]) {
+			warnings = append(warnings,
+				`<column width="N"> with an integer value has no effect in v2 XML; `+
+					`use width-ratio="0.5" (float 0–1) to set column width.`)
+			break
+		}
 	}
 	return warnings
 }
--- a/shortcuts/doc/docs_update_check.go
+++ b/shortcuts/doc/docs_update_check.go
@@ -310,10 +310,19 @@ func CheckV2XMLBareAmpersand(content string) string {
 		"Escape it as &amp; (and < as &lt;, > as &gt; where needed)."
 }
 
+// quoteContainerTagRe matches the opening of a <quote-container> element
+// (tag name followed by whitespace, >, or />) to avoid false positives on
+// hypothetical attributes or element names that start with "quote-container".
+var quoteContainerTagRe = regexp.MustCompile(`<quote-container(?:\s|>|/)`)
+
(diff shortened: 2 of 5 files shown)


shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +279,68 @@ func leadingRun(s string, c byte) string {
...
+	return "content contains a bare & character that is not a valid XML entity reference; " +
+		"the v2 XML parser will reject the request. " +
+		"Escape it as &amp; (and < as &lt;, > as &gt; where needed)."
+}
+
+// columnIntWidthRe matches a <column … width="N" …> attribute where N is a
+// plain integer. In v2 XML the valid attribute is width-ratio (a float 0–1),
+// not width. An integer width silently has no effect on column sizing.
+var columnIntWidthRe = regexp.MustCompile(`<column\b[^>]*\bwidth="(\d+)"`)
+
```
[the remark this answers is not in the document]
herbertliu: Fixed in 9fa94ab: replaced strings.Contains with quoteContainerTagRe (`<quote-container(?:\s|>|/)`) to avoid matching attribute names that merely start with "quote-container". Tightened columnIntWidthRe to require whitespace before "width" (preventing data-width false positives) and accept both single and double quotes with optional spaces around "=". Added test cases for all four new scenarios.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +279,77 @@ func leadingRun(s string, c byte) string {
...
+// columnIntWidthRe matches a <column … width="N" …> attribute where N is a
+// plain integer. The pattern requires whitespace before "width" to avoid
+// matching unrelated attributes such as data-width, and accepts optional
+// whitespace around "=" and either single or double quotes, so forms like
+// width='50' or width = "50" are also detected. Go's RE2 engine does not
+// support backreferences, so mismatched quote pairs (e.g. width="50') are
+// also matched — that is acceptable for a non-blocking warning.
+var columnIntWidthRe = regexp.MustCompile(`<column\b[^>]*\swidth\s*=\s*['"]?\d+['"]?`)
+
+// CheckV2XMLWarnings returns a list of non-fatal warnings for v2 XML content.
```
fangshuyu-768, line 338: **`columnIntWidthRe` produces false positives on float values.**

The regex `<column\b[^>]*\swidth\s*=\s*['"]?\d+['"]?` will match `width="0.5"` because `\d+` greedily matches the `0` before the dot, producing a spurious warning for a valid float value.

Suggested fix: use `FindStringSubmatch` to capture the attribute value and then check in Go code whether it's a pure integer (no decimal point), e.g.:

```go
var columnWidthRe = regexp.MustCompile(`<column\b[^>]*\swidth\s*=\s*(['"])([^'"]+)\1`)

// then in CheckV2XMLWarnings:
for _, m := range columnWidthRe.FindAllStringSubmatch(content, -1) {
    val := m[2]
    if _, err := strconv.Atoi(val); err == nil {
        // pure integer → warn
    }
}
```

Alternatively, if the regex approach is preferred, add a negative lookahead equivalent by excluding values containing a dot: `\d+(?!\.\d)` — but since Go's RE2 doesn't support lookahead, the code-based check is the way to go.

shortcuts/doc/docs_update_check.go
```diff
@@ -279,3 +279,77 @@ func leadingRun(s string, c byte) string {
...
+// return this as a hard error.
+//
+// Go's regexp package does not support lookahead, so we detect bare ampersands
+// by replacing all valid entity references with a placeholder and then
+// checking whether any & remains.
+func CheckV2XMLBareAmpersand(content string) string {
+	if content == "" || !strings.Contains(content, "&") {
+		return ""
+	}
+	// Replace every valid entity with its same-length placeholder so positional
```
fangshuyu-768: **Comment says "same-length placeholder" but `"ENTITY"` is not same-length.**

The comment states: _"Replace every valid entity with its same-length placeholder so positional byte offsets are preserved"_, but `"ENTITY"` (6 bytes) is not the same length as `&lt;` (4), `&amp;` (5), `&#65;` (5), etc. The parenthetical "(not required here, but avoids false positives)" also contradicts the stated purpose.

Since positional offsets aren't actually used anywhere, suggest simplifying the comment to something like:

```go
// Replace every valid entity with a placeholder so that any remaining &
// must be a bare ampersand.
```

shortcuts/doc/docs_update_check_test.go
```diff
@@ -373,3 +373,125 @@ func TestDocsUpdateWarningsEmpty(t *testing.T) {
...
+
+func containsStr(s, sub string) bool {
+	return len(s) >= len(sub) && (s == sub || len(sub) == 0 ||
+		func() bool {
+			for i := 0; i+len(sub) <= len(s); i++ {
+				if s[i:i+len(sub)] == sub {
+					return true
+				}
+			}
+			return false
```
fangshuyu-768: **Custom `containsStr` helper is unnecessary — use `strings.Contains`.**

This hand-rolled implementation replicates `strings.Contains` with added complexity (nested anonymous function). There's no performance concern here since this is test code.

Suggested replacement:

```go
if !strings.Contains(combined, sub) {
    t.Errorf("expected warning to contain %q, got: %s", sub, combined)
}
```

shortcuts/doc/docs_create_v2.go
```diff
@@ -27,6 +28,11 @@ func validateCreateV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if runtime.Str("parent-token") != "" && runtime.Str("parent-position") != "" {
 		return common.FlagErrorf("--parent-token and --parent-position are mutually exclusive")
 	}
+	if runtime.Str("doc-format") != "markdown" {
```
fangshuyu-768, line 31: **Inconsistent empty-content guard between create and update paths.**

Here in `validateCreateV2` there's no `content != ""` guard before calling `CheckV2XMLBareAmpersand`, but in `validateUpdateV2` (line ~106) there is one:

```go
if runtime.Str("doc-format") != "markdown" && content != "" {
```

While `CheckV2XMLBareAmpersand` handles empty strings internally, the inconsistency makes it look like an oversight. Same applies to the warning checks in `executeCreateV2` vs `executeUpdateV2`.

Suggested: either add the guard here for consistency, or remove it from `validateUpdateV2` and rely on the callee's internal check in both places.

shortcuts/doc/docs_update_check_test.go
```diff
@@ -373,3 +373,125 @@ func TestDocsUpdateWarningsEmpty(t *testing.T) {
...
+		{
+			name:         "column integer width triggers warning",
+			content:      `<grid><column width="50"><p>A</p></column></grid>`,
+			wantContains: []string{"width-ratio"},
+			wantLen:      1,
+		},
+		{
+			name:    "column float width-ratio is fine",
+			content: `<grid><column width-ratio="0.5"><p>A</p></column></grid>`,
+			wantLen: 0,
```
fangshuyu-768, line 459: **Missing test case: mixed valid entities and bare ampersand.**

There's no test for content that contains both valid entities and a bare `&`, e.g. `"a &amp; b & c"`. This is a common real-world scenario and would help verify that the replace-then-check approach works correctly when valid and invalid references coexist.

Suggested addition:
```go
{name: "mixed valid entity and bare ampersand flagged", content: "a &amp; b & c", wantErr: true},
```

shortcuts/doc/docs_update_check_test.go
```diff
@@ -373,3 +373,125 @@ func TestDocsUpdateWarningsEmpty(t *testing.T) {
...
+		{
+			name:    "data-width attribute is not flagged",
+			content: `<column data-width="50"/>`,
+			wantLen: 0,
+		},
+		// single-quoted width should be caught
+		{
+			name:    "column single-quoted integer width triggers warning",
+			content: `<grid><column width='30'/></grid>`,
+			wantLen: 1,
```
fangshuyu-768, line 482: **Missing test case: `width` with float value should NOT trigger warning.**

This is the flip side of the `columnIntWidthRe` false-positive issue. Adding a test for `<column width="0.5">` would both document the intended behavior and catch the regex bug (it currently produces a false positive because `\d+` matches the `0` in `0.5`).

Suggested addition:
```go
{
    name:    "column float width value is fine",
    content: `<grid><column width="0.5"><p>A</p></column></grid>`,
    wantLen: 0,
},
```

fangshuyu-768 (commented):
Thanks for adding these pre-flight checks — the overall approach of catching silent failures early is solid. I have a few concerns below, ordered by impact.

shortcuts/doc/docs_update_check.go
herbertliu: Fixed in 259bc0e. Used `["']\d+["']` — requiring the value to be wrapped in quotes with digits immediately before the closing quote, so `width="0.5"` doesn't match (the dot prevents `\d+` from consuming the full value up to the quote). Avoids backreferences (not supported in Go RE2). Added a test case covering `width="0.5"` → wantLen: 0.

shortcuts/doc/docs_update_check.go
herbertliu: Fixed in 259bc0e. Updated the comment to: _"Replace every valid entity reference with a fixed placeholder so that the subsequent Contains check only fires on truly bare ampersands. ('ENTITY' is not the same length as each entity; byte offsets are not preserved, but that is fine — we only need a yes/no bare-& answer.)"_

shortcuts/doc/docs_update_check_test.go
herbertliu: Fixed in 259bc0e. Removed `containsStr` and replaced all usages with `strings.Contains`.

shortcuts/doc/docs_create_v2.go
herbertliu: Fixed in 259bc0e. Added `&& runtime.Str("content") != ""` guard in `validateCreateV2` to match `validateUpdateV2`, and the same guard in `executeCreateV2`.

shortcuts/doc/docs_update_check_test.go
herbertliu: Fixed in 259bc0e. Added test case `{name: "valid entity mixed with bare ampersand flagged", content: "a &amp; b & c", wantErr: true}`.

shortcuts/doc/docs_update_check_test.go
herbertliu: Fixed in 259bc0e. Added test case `{name: "column float width value is not flagged", content: `...`, wantLen: 0}`.

SunPeiYang996:
Thanks for looking into this and for the thoughtful guardrails here.

For the unescaped ampersand case, for example `<p>A & B</p>`, the server currently handles it in a compatible way and normalizes it to `<p>A &amp; B</p>` automatically. If the CLI blocks this before sending the request, it would reduce the overall success rate for model-generated content.

For `quote-container` and `column width="50"`, these are Doc v1 usages and should not appear in the v2 XML path. In essence, this is a mix of v1 and v2 conventions. If we want to add guidance for this, we would prefer to do it on the server side so it can be maintained consistently over time. The CLI is a client here and should mainly be responsible for routing.

I am going to close this PR for now. If you have a better idea or want to discuss another approach, I would be happy to continue the conversation.

PR #825, opened 2026-05-11
feat(doc): warn before overwrite when document contains whiteboard or file blocks

herbertliu:
## Summary

Add a best-effort pre-flight guard for the v1 `overwrite` mode that warns the user when the current document contains whiteboard or file attachment blocks that cannot be reconstructed from Markdown.

## Changes

- **`shortcuts/doc/docs_update.go`**: In `executeUpdateV1`, pre-fetch the document when `--mode overwrite` and call `warnOverwriteResourceBlocks` / `checkOverwriteResourceBlocks` to detect resource blocks
- **`shortcuts/doc/docs_update_test.go`**: Add `TestCheckOverwriteResourceBlocks` with table-driven cases covering single/multiple whiteboard blocks, file attachments, and combined cases

## Behaviour

When `--mode overwrite` is used and the current document contains `<whiteboard>` or `<file>` blocks:

```
warning: the document contains 2 whiteboard blocks and 1 file attachment block that cannot be
reconstructed from Markdown; overwrite will permanently delete them.
Consider fetching a backup with `docs +fetch` before overwriting.
```

The guard is **best-effort**: if the pre-fetch fails (network error, permission, etc.) the warning is silently skipped and the overwrite proceeds normally. The overwrite is never blocked — this is a warning only.

## Test Plan

- [x] `go test ./shortcuts/doc/...` — all pass
- [x] `go vet ./...` — clean
- [x] `gofmt -l .` — clean

## Summary by CodeRabbit

* **New Features**
  * Added a best-effort pre-check that prints a non-blocking overwrite warning when overwriting documents that contain embedded whiteboards or file attachments; the check is skipped silently if the pre-fetch fails.

* **Tests**
  * Added automated tests covering the overwrite-warning behavior and selection-title validation to ensure correct warnings and error messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/825)

### DIFF
--- a/shortcuts/doc/docs_update.go
+++ b/shortcuts/doc/docs_update.go
@@ -224,9 +224,6 @@ func warnOverwriteResourceBlocks(runtime *common.RuntimeContext) string {
 		return ""
 	}
 	md, _ := result["markdown"].(string)
-	if md == "" {
-		return ""
-	}
 	return checkOverwriteResourceBlocks(md)
 }
 
--- a/shortcuts/doc/docs_update.go
+++ b/shortcuts/doc/docs_update.go
@@ -160,6 +160,16 @@ func executeUpdateV1(_ context.Context, runtime *common.RuntimeContext) error {
 		fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", w)
 	}
 
+	// Overwrite replaces the entire document, silently discarding any
+	// whiteboard or file-attachment blocks that cannot be re-created from
+	// Markdown. Pre-fetch the current content and warn when such blocks
+	// are present so the caller can take a backup before proceeding.
+	if runtime.Str("mode") == "overwrite" {
+		if w := warnOverwriteResourceBlocks(runtime); w != "" {
+			fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", w)
+		}
+	}
+
 	// Surface callout type= hint so users know to switch to background-color/
 	// border-color when they want a colored callout. Non-blocking, advisory.
 	if md := runtime.Str("markdown"); md != "" {
@@ -196,4 +206,56 @@ func buildUpdateArgsV1(runtime *common.RuntimeContext) map[string]interface{} {
 		args["new_title"] = v
 	}
 	return args
+}
+
+// warnOverwriteResourceBlocks pre-fetches the current document and returns a
+// non-empty warning string when the document contains whiteboard or file
+// attachment blocks that would be permanently deleted by an overwrite. Returns
+// an empty string (no warning) when the document is clean or the fetch fails
+// (we never block the overwrite on a best-effort check).
+func warnOverwriteResourceBlocks(runtime *common.RuntimeContext) string {
+	args := map[string]interface{}{
+		"doc_id":           runtime.Str("doc"),
+		"skip_task_detail": true,
+	}
+	result, err := common.CallMCPTool(runtime, "fetch-doc", args)
+	if err != nil {
+		// Fetch failed — silently skip the guard rather than blocking overwrite.
+		return ""
+	}
+	md, _ := result["markdown"].(string)
+	if md == "" {
+		return ""
+	}
+	return checkOverwriteResourceBlocks(md)
+}
+
+// checkOverwriteResourceBlocks scans Markdown for resource block tags that
+// cannot survive an overwrite: <whiteboard …> and <file …>. Returns a
+// warning string listing the counts if any are found, empty string otherwise.
+func checkOverwriteResourceBlocks(markdown string) string {
+	var found []string
+	if n := strings.Count(markdown, "<whiteboard"); n > 0 {
+		if n == 1 {
+			found = append(found, "1 whiteboard block")
+		} else {
+			found = append(found, fmt.Sprintf("%d whiteboard blocks", n))
+		}
+	}
+	if n := strings.Count(markdown, "<file"); n > 0 {
+		if n == 1 {
+			found = append(found, "1 file attachment block")
+		} else {
+			found = append(found, fmt.Sprintf("%d file attachment blocks", n))
+		}
+	}
+	if len(found) == 0 {
+		return ""
+	}
+	return fmt.Sprintf(
+		"the document contains %s that cannot be reconstructed from Markdown; "+
+			"overwrite will permanently delete them. "+
+			"Consider fetching a backup with `docs +fetch` before overwriting.",
+		strings.Join(found, " and "),
+	)
 }
--- a/shortcuts/doc/docs_update.go
+++ b/shortcuts/doc/docs_update.go
@@ -6,6 +6,7 @@ package doc
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


shortcuts/doc/docs_update.go
```diff
@@ -197,3 +207,52 @@ func buildUpdateArgsV1(runtime *common.RuntimeContext) map[string]interface{} {
...
+	md, _ := result["markdown"].(string)
+	return checkOverwriteResourceBlocks(md)
+}
+
+// checkOverwriteResourceBlocks scans Markdown for resource block tags that
+// cannot survive an overwrite: <whiteboard …> and <file …>. Returns a
+// warning string listing the counts if any are found, empty string otherwise.
+func checkOverwriteResourceBlocks(markdown string) string {
+	var found []string
+	if n := strings.Count(markdown, "<whiteboard"); n > 0 {
```
fangshuyu-768: **Substring matching is too coarse — risk of false positives.**

`strings.Count(markdown, "<whiteboard")` and `strings.Count(markdown, "<file")` can produce false matches:

1. **Text content**: If the document contains prose mentioning whiteboards (e.g. a code example `<whiteboard token="..."/>`), it will be counted as a whiteboard block.
2. **`<file` prefix is overly broad**: It could match `<file-view>`, `<file-name>`, or other tags that are not file attachment blocks.

Suggested fix — use a more precise regex:

```go
var (
    whiteboardRe = regexp.MustCompile(`<whiteboard[\s/>]`)
    fileRe       = regexp.MustCompile(`<file[\s/>]`)
)
```

Or better yet, check block types directly via the document blocks API (`/blocks`) instead of scanning the Markdown representation.

shortcuts/doc/docs_update_test.go
```diff
@@ -101,3 +168,35 @@ func TestNormalizeWhiteboardResult(t *testing.T) {
 		}
 	})
 }
+
+func TestValidateSelectionByTitleV1(t *testing.T) {
```
fangshuyu-768, line 172: **Unrelated test added in this PR.**

`TestValidateSelectionByTitleV1` tests `validateSelectionByTitleV1`, which is not part of this PR's changes. Unrelated tests should go in a separate PR to keep the diff focused and the review scoped.

shortcuts/doc/docs_update.go
```diff
@@ -160,6 +160,16 @@ func executeUpdateV1(_ context.Context, runtime *common.RuntimeContext) error {
 		fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", w)
 	}
 
+	// Overwrite replaces the entire document, silently discarding any
+	// whiteboard or file-attachment blocks that cannot be re-created from
+	// Markdown. Pre-fetch the current content and warn when such blocks
+	// are present so the caller can take a backup before proceeding.
+	if runtime.Str("mode") == "overwrite" {
+		if w := warnOverwriteResourceBlocks(runtime); w != "" {
```
fangshuyu-768, line 169: **Extra latency from pre-fetch on every overwrite.**

Every `--mode overwrite` call now incurs an additional `fetch-doc` MCP round-trip, even when the document has no resource blocks. This adds network latency to the critical path.

Consider:
- Documenting the performance impact in the PR description or code comments
- Making the guard opt-in via a flag (e.g. `--warn-resource-loss`) so callers who don't need the warning aren't penalized

shortcuts/doc/docs_update.go
```diff
@@ -197,3 +207,52 @@ func buildUpdateArgsV1(runtime *common.RuntimeContext) map[string]interface{} {
 	}
 	return args
 }
+
+// warnOverwriteResourceBlocks pre-fetches the current document and returns a
+// non-empty warning string when the document contains whiteboard or file
+// attachment blocks that would be permanently deleted by an overwrite. Returns
+// an empty string (no warning) when the document is clean or the fetch fails
+// (we never block the overwrite on a best-effort check).
+func warnOverwriteResourceBlocks(runtime *common.RuntimeContext) string {
```
fangshuyu-768, line 231: **`warnOverwriteResourceBlocks` lacks test coverage.**

`checkOverwriteResourceBlocks` has thorough unit tests, but `warnOverwriteResourceBlocks` (the MCP integration layer) has none. While MCP calls are hard to mock, it would be helpful to either:
- Add a comment explaining why it's not tested (best-effort, depends on external MCP call)
- Provide a mock-based test for the happy path and the silent-failure-on-error path

shortcuts/doc/docs_update.go
```diff
@@ -197,3 +207,52 @@ func buildUpdateArgsV1(runtime *common.RuntimeContext) map[string]interface{} {
...
+	args := map[string]interface{}{
+		"doc_id":           runtime.Str("doc"),
+		"skip_task_detail": true,
```
fangshuyu-768, line 236: **`skip_task_detail: true` lacks a comment.**

The purpose of `skip_task_detail` is not self-evident. Please add a brief comment explaining why it's set (presumably to reduce response payload size and speed up the pre-fetch).

fangshuyu-768 (commented):
Thanks for the PR! The best-effort warning design is sensible. I have a few concerns — the substring matching issue (#1) could lead to false positives and is worth addressing; the rest are suggestions.

shortcuts/doc/docs_update.go
herbertliu: Fixed in 977e057. Replaced `strings.Count` with a regex `<(whiteboard|file)[\s/>]` that requires the tag name to be followed by whitespace, `>`, or `/` — so prose mentioning "whiteboard" or tags like `<file-view>` no longer match.

shortcuts/doc/docs_update_test.go
herbertliu: Acknowledged. The test was added to improve patch coverage for the PR (codecov/patch was at 58.82%, below the 60% threshold) rather than because it is logically related to this change. Happy to move it to a separate follow-up PR if that is preferred, but since the function being tested (`validateSelectionByTitleV1`) lives in the same file and had zero test coverage, it felt like a reasonable addition in-place.

shortcuts/doc/docs_update.go
herbertliu: Documented in 977e057. Added a comment above `warnOverwriteResourceBlocks` explaining that every `--mode overwrite` call incurs one extra fetch-doc round-trip, that the cost is intentional, and why the trade-off is acceptable (best-effort, silent on failure, bounded latency).

shortcuts/doc/docs_update.go
herbertliu: Documented in 977e057. Added a comment to `warnOverwriteResourceBlocks` explaining that it is not unit-tested because it depends on an external MCP call, and that the pure detection logic is fully covered by `checkOverwriteResourceBlocks` tests.

shortcuts/doc/docs_update.go
herbertliu: Fixed in 977e057. Added an inline comment: `// skip_task_detail reduces response payload by omitting per-block task metadata, making the pre-fetch faster and cheaper.`

fangshuyu-768 (approved):

PR #832, opened 2026-05-11
feat(doc): add --width/--height flags to docs +media-insert

### LINKED ISSUE #489: [docs +media-insert] Support --width/--height for image display size
## Summary

`lark-cli docs +media-insert` can upload and insert an image into a Feishu/Lark doc successfully, but it does not provide any way to control the image's **display width/height** in the document.

At the moment, the shortcut appears to bind the uploaded image token plus optional `align` / `caption`, but not image size. In practice this often causes inserted images to render at a very small default size in the document, which makes the shortcut hard to use in article or publishing workflows.

## Reproduce

```bash
npm install -g @larksuite/cli
lark-cli auth login

lark-cli docs +create --title "image size repro" --wiki-space <space_id> --markdown " "
# -> doc_id

lark-cli docs +media-insert --doc <doc_id> --file ./illustration.png --type image --align center
```

Open the doc in Feishu/Lark. The image is inserted, but there is no CLI option to set a display width/height, so the rendered result may be much smaller than expected for normal body illustrations or banners.

## Expected

`docs +media-insert` should support explicit image display sizing.

For example:

```bash
lark-cli docs +media-insert \
  --doc <doc_id> \
  --file ./img.png \
  --type image \
  --align center \
  --width 800 \
  --height 447 \
  --caption "..."
```

Suggested semantics:

- `--width`: desired display width in the Feishu document
- `--height`: desired display height in the Feishu document
- if `--height` is omitted, compute it from `--width` and the source image's native aspect ratio

## Current workaround

After `+media-insert`, callers can make an additional docx block update call (for example via `blocks/batch_update`) to set image width/height explicitly.

This works, but requires each caller to:

1. read native image dimensions
2. compute the scaled height
3. extract the inserted `block_id` / `file_token`
4. make another API call

That logic feels like a natural fit for the shortcut itself rather than something every downstream workflow has to re-implement.

## Why this matters

- `docs +media-insert` is the first-class shortcut for inserting local images into a doc at a precise position
- for image-heavy document workflows, size control is part of the insertion experience, not an edge case
- exposing width/height in the shortcut would remove a lot of repeated downstream glue code

Happy to contribute a PR if this API shape makes sense.

---
Environment: `@larksuite/cli 1.0.11`, macOS, Node 20+, user auth

songyoung77:
## Summary

Add `--width` and `--height` flags to `docs +media-insert` so users and AI agents can control the display size of inserted images. When only one dimension is specified, the other is auto-calculated from the image's native aspect ratio.

## Changes

- Add `--width` and `--height` int flags with validation in `shortcuts/doc/doc_media_insert.go`
- Add `computeMissingDimension()` and `detectImageDimensions()` aspect-ratio helpers in `shortcuts/doc/doc_media_insert.go`
- Wire dimension resolution into Execute (after upload, before batch_update) and DryRun (best-effort preview) in `shortcuts/doc/doc_media_insert.go`
- Add 13 unit tests covering payload construction, validation, and aspect-ratio calculation in `shortcuts/doc/doc_media_insert_test.go`
- Update SKILL.md with new flags and usage examples in `skills/lark-doc/references/lark-doc-media-insert.md`

## Test Plan

- [x] `make unit-test` passed
- [x] `make validate` passed
- [x] local-eval passed (E2E 7/8 — 1 env-only failure: test app lacks `space:document:delete` scope for cleanup; skillave 6/6)
- [x] acceptance-reviewer passed (5/5 cases)
- [x] manual verification: `lark-cli docs +media-insert --doc <token> --file img.png --width 600 --as bot` — confirmed aspect-ratio auto-calc and output fields

## Related Issues

Closes #489

## Summary by CodeRabbit

* **New Features**
  * Optional --width and --height for docs +media-insert (px); if only one is given, the missing dimension is auto-computed from image aspect ratio when detectable.
  * Execute and dry-run now surface resolved image dimensions; dry-run annotates clipboard-only inputs instead of decoding.

* **Bug Fixes / Validation**
  * --width/--height rejected for non-image types; values must be positive. Execute fails if native dimension detection needed but unavailable.

* **Documentation**
  * Command docs updated with sizing examples and supported formats (PNG/JPEG/GIF).

* **Tests**
  * Added unit tests for validation, sizing, and aspect-ratio computation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/832)

### DIFF
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -7,6 +7,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"image"
+	_ "image/gif"
+	_ "image/jpeg"
+	_ "image/png"
+	"io"
 	"path/filepath"
 	"strings"
 
@@ -464,6 +469,35 @@ func resolveDocxDocumentID(runtime *common.RuntimeContext, input string) (string
 	default:
 		return "", output.ErrValidation("docs +media-insert only supports docx documents")
 	}
+}
+
+type imageDimensions struct {
+	width  int
+	height int
+}
+
+func computeMissingDimension(userWidth, userHeight, nativeWidth, nativeHeight int) imageDimensions {
+	if userWidth > 0 && userHeight == 0 {
+		return imageDimensions{
+			width:  userWidth,
+			height: userWidth * nativeHeight / nativeWidth,
+		}
+	}
+	if userHeight > 0 && userWidth == 0 {
+		return imageDimensions{
+			width:  userHeight * nativeWidth / nativeHeight,
+			height: userHeight,
+		}
+	}
+	return imageDimensions{width: userWidth, height: userHeight}
+}
+
+func detectImageDimensions(r io.Reader) (width, height int, err error) {
+	cfg, _, err := image.DecodeConfig(r)
+	if err != nil {
+		return 0, 0, err
+	}
+	return cfg.Width, cfg.Height, nil
 }
 
 func buildBatchUpdateData(blockID, mediaType, fileToken, alignStr, caption string, width, height int) map[string]interface{} {
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -379,15 +379,15 @@ var DocMediaInsert = common.Shortcut{
 				} else {
 					f, openErr := runtime.FileIO().Open(filePath)
 					if openErr != nil {
-						return output.ErrValidation(
-							"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName)
+						return withRollbackWarning(output.ErrValidation(
+							"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName))
 					}
 					nativeW, nativeH, dimErr = detectImageDimensions(f)
 					f.Close()
 				}
 				if dimErr != nil {
-					return output.ErrValidation(
-						"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName)
+					return withRollbackWarning(output.ErrValidation(
+						"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName))
 				}
 				dims := computeMissingDimension(userWidth, userHeight, nativeW, nativeH)
 				finalWidth = dims.width
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -539,15 +539,15 @@ type imageDimensions struct {
 }
 
 func computeMissingDimension(userWidth, userHeight, nativeWidth, nativeHeight int) imageDimensions {
-	if userWidth > 0 && userHeight == 0 {
+	if userWidth > 0 && userHeight == 0 && nativeWidth > 0 {
 		return imageDimensions{
 			width:  userWidth,
-			height: userWidth * nativeHeight / nativeWidth,
+			height: (userWidth*nativeHeight + nativeWidth/2) / nativeWidth,
 		}
 	}
-	if userHeight > 0 && userWidth == 0 {
+	if userHeight > 0 && userWidth == 0 && nativeHeight > 0 {
 		return imageDimensions{
-			width:  userHeight * nativeWidth / nativeHeight,
+			width:  (userHeight*nativeWidth + nativeHeight/2) / nativeHeight,
 			height: userHeight,
 		}
 	}
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -113,10 +113,10 @@ var DocMediaInsert = common.Shortcut{
 		}
 		const maxDimension = 10000
 		if widthChanged && runtime.Int("width") > maxDimension {
-			return output.ErrValidation("--width must not exceed %d", maxDimension)
+			return output.ErrValidation("--width must not exceed %d pixels", maxDimension)
 		}
 		if heightChanged && runtime.Int("height") > maxDimension {
-			return output.ErrValidation("--height must not exceed %d", maxDimension)
+			return output.ErrValidation("--height must not exceed %d pixels", maxDimension)
 		}
 		return nil
 	},
--- a/shortcuts/doc/doc_media_insert.go
+++ b/shortcuts/doc/doc_media_insert.go
@@ -360,25 +360,7 @@ var DocMediaInsert = common.Shortcut{
 		// interface stays a true nil for the --file path. Passing a typed-nil
 		// *bytes.Reader here would make the downstream `if cfg.Content != nil`
 		// check incorrectly take the clipboard branch and crash on Read.
-		uploadCfg := UploadDocMediaFileConfig{
-			FilePath:   filePath,
-			FileName:   fileName,
-			FileSize:   fileSize,
-			ParentType: parentTypeForMediaType(mediaType),
-			ParentNode: uploadParentNode,
-			DocID:      documentID,
-		}
-		if clipboardContent != nil {
-			uploadCfg.Reader = bytes.NewReader(clipboardContent)
-		}
-		fileToken, err := uploadDocMediaFile(runtime, uploadCfg)
-		if err != nil {
-			return withRollbackWarning(err)
-		}
-
-		fmt.Fprintf(runtime.IO().ErrOut, "File uploaded: %s\n", fileToken)
-
-		// Resolve display dimensions for image blocks.
+		// Resolve display dimensions before upload to fail fast on unreadable images.
 		var finalWidth, finalHeight int
 		if mediaType == "image" {
 			userWidth := runtime.Int("width")
@@ -397,15 +379,15 @@ var DocMediaInsert = common.Shortcut{
 				} else {
 					f, openErr := runtime.FileIO().Open(filePath)
 					if openErr != nil {
-						return withRollbackWarning(output.ErrValidation(
-							"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName))
+						return output.ErrValidation(
+							"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName)
 					}
 					nativeW, nativeH, dimErr = detectImageDimensions(f)
 					f.Close()
 				}
 				if dimErr != nil {
-					return withRollbackWarning(output.ErrValidation(
-						"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName))
+					return output.ErrValidation(
+						"unable to detect image dimensions from %s for aspect-ratio calculation; provide both --width and --height", fileName)
 				}
 				dims := computeMissingDimension(userWidth, userHeight, nativeW, nativeH)
 				finalWidth = dims.width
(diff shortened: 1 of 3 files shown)


shortcuts/doc/doc_media_insert.go
```diff
@@ -453,7 +533,45 @@ func resolveDocxDocumentID(runtime *common.RuntimeContext, input string) (string
...
+	return cfg.Width, cfg.Height, nil
+}
+
+func detectImageDimensionsFromPath(fio fileio.FileIO, filePath string) (int, int, error) {
+	f, err := fio.Open(filePath)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer f.Close()
+	return detectImageDimensions(f)
```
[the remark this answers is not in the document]
songyoung77: Good catch — applied the fix in 9386118. While `fio.Open()` calls `SafeInputPath` internally, the explicit check here makes the validation visible at the call site.

shortcuts/doc/doc_media_insert.go
```diff
@@ -453,7 +533,48 @@ func resolveDocxDocumentID(runtime *common.RuntimeContext, input string) (string
 	}
 }
 
-func buildBatchUpdateData(blockID, mediaType, fileToken, alignStr, caption string) map[string]interface{} {
+type imageDimensions struct {
+	width  int
+	height int
+}
+
+func computeMissingDimension(userWidth, userHeight, nativeWidth, nativeHeight int) imageDimensions {
```
fangshuyu-768, line 550: **Potential panic: division by zero.**

If `image.DecodeConfig` returns `nativeWidth == 0` or `nativeHeight == 0` (e.g. corrupted image, unusual format), the integer divisions on lines 545 and 550 will panic at runtime.

Suggested guard:

```go
func computeMissingDimension(userWidth, userHeight, nativeWidth, nativeHeight int) imageDimensions {
    if nativeWidth <= 0 || nativeHeight <= 0 {
        return imageDimensions{width: userWidth, height: userHeight}
    }
    // ... existing logic
}
```

shortcuts/doc/doc_media_insert.go
```diff
@@ -453,7 +533,48 @@ func resolveDocxDocumentID(runtime *common.RuntimeContext, input string) (string
...
+	if userWidth > 0 && userHeight == 0 {
+		return imageDimensions{
+			width:  userWidth,
+			height: userWidth * nativeHeight / nativeWidth,
```
fangshuyu-768: **Integer division truncation.**

Go's integer division truncates toward zero. For example, a 1000×333 native image with `userWidth=999` yields `999*333/1000 = 332`, whereas the exact value is 332.667. A one-pixel off result is usually acceptable, but if precision matters, consider rounding:

```go
height: (userWidth*nativeHeight + nativeWidth/2) / nativeWidth
```

Same applies to line 550.

shortcuts/doc/doc_media_insert.go
```diff
@@ -93,6 +100,17 @@ var DocMediaInsert = common.Shortcut{
 				return output.ErrValidation("--file-view only applies when --type=file")
 			}
 		}
+		widthChanged := runtime.Changed("width")
+		heightChanged := runtime.Changed("height")
+		if (widthChanged || heightChanged) && runtime.Str("type") != "image" {
+			return output.ErrValidation("--width/--height only apply when --type=image")
+		}
+		if widthChanged && runtime.Int("width") <= 0 {
```
fangshuyu-768, line 108: **No upper bound validation for --width/--height.**

Only `> 0` is checked. An extremely large value (e.g. `--width 999999999`) could cause API errors or unexpected behavior. Consider adding a reasonable upper limit, e.g.:

```go
const maxDimension = 10000
if widthChanged && runtime.Int("width") > maxDimension {
    return output.ErrValidation("--width must not exceed %d pixels", maxDimension)
}
```

shortcuts/doc/doc_media_insert_test.go
```diff
@@ -215,6 +216,48 @@ func TestBuildBatchUpdateDataForFile(t *testing.T) {
...
+				"block_id": "blk_2",
+				"replace_file": map[string]interface{}{
+					"token": "file_tok",
+				},
+			},
+		},
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("buildBatchUpdateData(file, 800, 600) = %#v, want %#v", got, want)
+	}
```
fangshuyu-768, line 258: **Incomplete test coverage for `computeMissingDimension`.**

Missing test cases:
1. Both dimensions provided (userWidth > 0 && userHeight > 0) — should return as-is.
2. Neither dimension provided (0, 0) — should return (0, 0).
3. `nativeWidth == 0` or `nativeHeight == 0` — boundary/division-by-zero case.

Also, `detectImageDimensions` and `detectImageDimensionsFromPath` have no unit tests.

fangshuyu-768 (commented):
Thanks for the PR! Overall the implementation is clean and well-structured. I have a few concerns below — the division-by-zero risk (#1) should be addressed before merge, the rest are suggestions.

shortcuts/doc/doc_media_insert.go
```diff
@@ -120,7 +138,24 @@ var DocMediaInsert = common.Shortcut{
 		} else {
 			createBlockData["index"] = "<children_len>"
 		}
-		batchUpdateData := buildBatchUpdateData("<new_block_id>", mediaType, "<file_token>", runtime.Str("align"), caption)
+		// Best-effort dimension computation for dry-run.
+		dryWidth := runtime.Int("width")
+		dryHeight := runtime.Int("height")
+		widthChanged := runtime.Changed("width")
```
fangshuyu-768, line 151: **Repeated `runtime.Changed()` calls.**

`runtime.Changed("width")` and `runtime.Changed("height")` are called separately in `Validate` (lines 103-104), `DryRun` (lines 144-145), and `Execute` (lines 378-379). Consider extracting them once per function into local variables to reduce repetition and potential inconsistency if the flag names ever change.

Minor — feel free to ignore if the current style is preferred.

shortcuts/doc/doc_media_insert.go
```diff
@@ -332,21 +370,63 @@ var DocMediaInsert = common.Shortcut{
...
+			widthChanged := runtime.Changed("width")
+			heightChanged := runtime.Changed("height")
+
+			if widthChanged && heightChanged {
+				finalWidth = userWidth
+				finalHeight = userHeight
+			} else if widthChanged || heightChanged {
+				var nativeW, nativeH int
+				if clipboardContent != nil {
+					nativeW, nativeH, err = detectImageDimensions(bytes.NewReader(clipboardContent))
```
fangshuyu-768: **`err` variable reuse from outer scope.**

Here `err` is assigned via `=` (not `:=`), reusing the outer `err` variable. While logically `err` must be nil at this point (upload succeeded), the implicit dependency on outer state is fragile — a future refactor could introduce a path where `err` is non-nil here.

Consider using a separate variable name for clarity:

```go
var dimErr error
nativeW, nativeH, dimErr = detectImageDimensions(bytes.NewReader(clipboardContent))
if dimErr != nil { ... }
```

Same pattern on line 394.

shortcuts/doc/doc_media_insert.go
songyoung77: Fixed in bc9e52f. Added `nativeWidth > 0` / `nativeHeight > 0` guards — when native dimension is zero, the function now falls through to the passthrough return instead of panicking.

shortcuts/doc/doc_media_insert.go
songyoung77: Fixed in bc9e52f. Changed to rounded integer division: `(a*b + divisor/2) / divisor`. Added a test case verifying rounding behavior (999×333/1000 → 333 instead of truncated 332).

shortcuts/doc/doc_media_insert_test.go
songyoung77: Fixed in bc9e52f. Added 5 test cases: both dimensions provided, neither provided, zero nativeWidth, zero nativeHeight, and rounding behavior.

shortcuts/doc/doc_media_insert.go
songyoung77: Fixed in a7bc2f8. Added `maxDimension = 10000` upper bound — values exceeding it are rejected with a clear validation error.

shortcuts/doc/doc_media_insert.go
songyoung77: Acknowledged. Each of Validate/DryRun/Execute is a separate closure with its own scope, so the `runtime.Changed()` calls are already extracted into local vars within each function. Extracting further would require a shared struct that adds complexity without much benefit given the closures can't share locals. Leaving as-is.

shortcuts/doc/doc_media_insert.go
songyoung77: Fixed in a7bc2f8. Renamed to `dimErr` to avoid reusing the outer `err` variable.

shortcuts/doc/doc_media_insert.go
songyoung77: Updated in 61c9560 — adopted your suggested early-return guard at the top of the function. Cleaner and safer than my original per-branch approach. Thanks!

shortcuts/doc/doc_media_insert.go
songyoung77: Updated error message in 56a37c0 to include `pixels` unit, matching your suggestion exactly.

shortcuts/doc/doc_media_insert.go
```diff
@@ -120,7 +145,24 @@ var DocMediaInsert = common.Shortcut{
...
+		// Best-effort dimension computation for dry-run.
+		dryWidth := runtime.Int("width")
+		dryHeight := runtime.Int("height")
+		widthChanged := runtime.Changed("width")
+		heightChanged := runtime.Changed("height")
+
+		if (widthChanged || heightChanged) && !(widthChanged && heightChanged) {
+			// One dimension provided — try best-effort DecodeConfig for --file path.
+			if filePath != "<clipboard image>" {
+				if nativeW, nativeH, err := detectImageDimensionsFromPath(runtime.FileIO(), filePath); err == nil {
```
fangshuyu-768: Concern: dry-run behavior diverges from Execute here.

When exactly one of `--width` / `--height` is provided, `DryRun` does a best-effort dimension detection here but silently ignores any failure.

That makes the dry-run output diverge from the real execution path: in `Execute`, the same open/decode failure is treated as fatal and the command exits with a validation error. As a result, `--dry-run` may show a request shape that the real command can never execute (for example with unsupported or unreadable local images).

Could we surface a dry-run note/error here, or otherwise make the dry-run behavior match `Execute` more closely?

shortcuts/doc/doc_media_insert.go
```diff
@@ -332,21 +377,64 @@ var DocMediaInsert = common.Shortcut{
...
+		if mediaType == "image" {
+			userWidth := runtime.Int("width")
+			userHeight := runtime.Int("height")
+			widthChanged := runtime.Changed("width")
+			heightChanged := runtime.Changed("height")
+
+			if widthChanged && heightChanged {
+				finalWidth = userWidth
+				finalHeight = userHeight
+			} else if widthChanged || heightChanged {
```
fangshuyu-768, line 374: Small concern about the ordering here: dimension detection happens after the upload.

So if the user provided only one dimension and local detection fails, we only discover that after the remote upload has already completed. Even if block rollback succeeds, this still does unnecessary remote work and may leave cleanup behavior dependent on backend semantics.

It might be cleaner to validate/compute the dimensions first, then upload, since the local file / clipboard bytes are already available at this point.

shortcuts/doc/doc_media_insert.go
songyoung77: Fixed in 0f506a7. DryRun now surfaces warnings when dimension detection fails (both for clipboard images and unreadable files), making the behavior consistent with Execute. This way users get early feedback that they need to provide both --width and --height.

shortcuts/doc/doc_media_insert.go
songyoung77: Good point — moved dimension detection before the upload in 9c181b4. Now if detection fails, we return a validation error immediately without doing any remote work. Also removed the `withRollbackWarning` wrapper since no remote state exists at that point yet.

songyoung77: Follow-up fix in f9bc606: restored `withRollbackWarning` on the dimension detection error returns. Since dimension detection runs after Step 2 (create placeholder block), failures must clean up the placeholder to avoid leaving an empty block in the document. The previous commit (9c181b4) inadvertently dropped the wrapper when moving the detection code.

fangshuyu-768 (approved):

PR #846, opened 2026-05-12
docs(lark-task): clarify list task search routing

BytedanceSearch:
## Summary

- Clarify when task list-style requests should use `task +get-my-tasks` or `task +get-related-tasks` instead of keyword search.
- Document that `task +get-my-tasks` has an implicit assignee/负责人 filter for the current login user.
- Clarify that another person's assignee list should resolve that person to `open_id` and use `task +search --assignee`.
- Keep this PR scoped to `skills/lark-task/**` only.

## Eval Signal

Runs:

- `tests/eval-search/runs/2026-05-12T06-14Z-multientity-sandbox-a970`
- `tests/eval-search/runs/2026-05-13Tfull-critical-info-sandbox`

Observed:

- Task cases 002/003 need list-style task handling and field-based validation instead of relying on search ranking alone.
- Task summaries should validate owner/assignee/follower, completion state, due time, and task GUID before computing priority or completion-rate summaries.

## Validation

- `git diff --check`
- PR diff is scoped to `skills/lark-task/**`.

Docs-only change; no unit tests were run.

### DIFF
--- a/skills/lark-task/references/lark-task-search.md
+++ b/skills/lark-task/references/lark-task-search.md
@@ -6,6 +6,19 @@
 
 Search tasks by keyword and optional filters.
 
+## When not to use `+search`
+
+If the user asks for a list-style task review with no real keyword, prefer the list shortcuts instead of `+search`. Do not turn generic instruction words such as "查找", "总结", "进度", "紧急程度", or date phrases such as "这个月" into `--query`.
+
+Use:
+
+- `task +get-related-tasks` for "我关注的", "我创建的", or "与我相关的" task lists, then apply completion / due-time filtering as needed.
+- `task +get-my-tasks` for "分配给我的" or "我负责的" task lists. The assignee/负责人 filter is implicit: this shortcut only lists tasks assigned to the current login user, and supports additional `--complete`, `--due-start`, and `--due-end` filters.
+- `task +search` when the request includes a task title fragment or domain keyword, for example "发布会", "客户反馈", or "周会纪要".
+- For "张三负责的" / "分配给某人" lists, resolve that person to `open_id` and use `task +search --assignee <open_id>`; `task +get-my-tasks` cannot change the assignee target.
+
+For list-style review requests, the answer should be computed from task fields rather than from search ranking. After fetching candidates, verify the owner/assignee/follower dimension, completion state, due time, and task GUID. If the user asks for "所有" or a completion-rate/priority summary, paginate or use the list shortcut's full result support before calculating counts; do not summarize only the first page.
+
 ## Recommended Commands
 
 ```bash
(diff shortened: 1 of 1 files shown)


skills/lark-drive/references/lark-drive-search.md
```diff
@@ -20,11 +20,15 @@
 > 正确：`lark-cli drive +search --query "方案"`
 > 错误：`lark-cli drive +search 方案`
 > `+search` 不接受位置参数；空 `--query` 或省略 `--query` 表示纯靠 filter 浏览（合法）。
+>
+> **列表型请求不要硬塞关键词**：如果用户只是要求"我这月创建的所有文档"、"最近半年我编辑过的文档"、"按类型分类统计"这类范围浏览 / 汇总请求，且没有给出标题片段或业务关键词，应使用 `--query ""` 搭配 `--mine`、`--created-*`、`--edited-*`、`--doc-types` 等过滤条件。不要把"查找"、"所有文档"、"最近更新过"、"按类型分类统计"这类动作词或统计意图放进 `--query`，否则会把本来应靠 filter 命中的结果过度收窄。
```
arnold9672: 这个描述，感觉质量挺高的。

skills/lark-task/references/lark-task-search.md
```diff
@@ -6,6 +6,16 @@
...
 Search tasks by keyword and optional filters.
 
+## When not to use `+search`
+
+If the user asks for a list-style task review with no real keyword, prefer the list shortcuts instead of `+search`. Do not turn generic instruction words such as "查找", "总结", "进度", "紧急程度", or date phrases such as "这个月" into `--query`.
+
+Use:
+
+- `task +get-related-tasks` for "我关注的", "我创建的", or "与我相关的" task lists, then apply completion / due-time filtering as needed.
+- `task +get-my-tasks` for "分配给我的" or "我负责的" task lists, with `--complete`, `--due-start`, and `--due-end` filters.
```
agicy: 为什么 **分配给我的** 对应附带 `--complete`（对应 **完成状态**）和 `--due`（对应 **截止时间**），用 **负责人** 也许更合适？

zero-my:
Thank you so much for taking the time to work on this, and especially for looking into the bad routing cases around task search. This is a very thoughtful contribution, and I really appreciate the intent behind it.

After looking through the change, I think the main concern is about where this guidance lives.

`skills/lark-task/references/lark-task-search.md` is already the reference doc for the `+search` shortcut itself, so its responsibility should mainly be to explain how to use `+search` correctly once that shortcut has already been chosen. Adding “when not to use `+search`” and cross-shortcut routing guidance here makes the action-level reference start defining dispatch behavior for itself, which feels like the wrong layer.

My understanding is that shortcut selection guidance should live in the higher-level entry doc, `skills/lark-task/SKILL.md`, which already contains the routing context for choosing between `+search` and the list-style shortcuts. Keeping that guidance in one place also makes it easier to maintain as the skill evolves.

Because of that, I don’t think I can approve this PR in its current form.

That said, I do think the motivation here is valuable, and it sounds very possible that you ran into a real bad-case in routing.

If you have a concrete example, it would be greatly appreciated if you could open an issue and include:
- the exact user request,
- the shortcut that was actually chosen, and
- the shortcut you expected instead.

That would help us improve the routing behavior in the right place.

And if you’d like to continue working on this, I think a follow-up PR that updates the routing guidance in `skills/lark-task/SKILL.md` would be a very welcome direction.

Thanks again for the contribution and for digging into this area.

PR #851, opened 2026-05-12
feat(events): register mail message_received in unified pipeline

dingding0418:
## Summary

Register `mail.user_mailbox.event.message_received_v1` as a first-class EventKey under `events/mail/`, so mailbox new-message events can be consumed through the unified `lark-cli event consume` contract instead of the standalone `mail +watch` fork. PreConsume hook auto-opens the per-user server-side subscription and cleanup auto-unsubscribes on graceful shutdown — agents now use the same command shape for IM and mail.

## Changes

- Add `events/mail/register.go` registering `mail.user_mailbox.event.message_received_v1` with `Schema.Native` (SDK type `larkmail.P2UserMailboxEventMessageReceivedV1Data`), `mailbox` param (default `me`), required scopes (`mail:event`, `mail:user_mailbox.event.mail_address:read`), `AuthTypes: [user]`, and `RequiredConsoleEvents` so missing console event subscription fails loudly at startup
- Wire `PreConsume` to call `POST /open-apis/mail/v1/user_mailboxes/<mailbox>/event/subscribe` and return a cleanup that calls `unsubscribe` (with a fresh `context.WithTimeout` so cleanup still reaches the server after the parent ctx is cancelled)
- Wire mail keys into `events/register.go`
- Add `events/mail/register_test.go` (10 tests): registration / native-key invariants, scope+auth metadata, mailbox param shape, default-mailbox path, custom mailbox in URL, slash escape (path-traversal defense), cleanup unsubscribe call, subscribe-failure must return nil cleanup, empty mailbox falls back to `me`, path helper assembly
- Update `skills/lark-event/SKILL.md`: extend description to mention mail; add a `Per-EventKey payload model` section explaining the IM-vs-mail asymmetry; add a typical `event consume → mail +message` pipeline example
- Update `skills/lark-mail/SKILL.md` and `references/lark-mail-watch.md` so the `+watch` row spells out when to use which path

## Note on payload thickness (intentional in this P0)

The mail event is intentionally thin: `event consume mail.user_mailbox.event.message_received_v1` emits `message_id` + `mail_address` + `mailbox_type` + `subscriber` only. Subject/body/attachments need a follow-up `mail +message --message-id <id>` call.

This is asymmetric with IM (whose WS payload natively carries `content`); the asymmetry is a Feishu Open Platform API design choice, not a framework limitation. Closing the asymmetry is scoped to follow-up PRs (see Next Steps).

## Next Steps (follow-up PRs)

- **P1**: add a `Process` function on the EventKey that fetches per-event message metadata via OAPI, so `event consume` emits subject/from/snippet inline (matching `mail +watch`'s default `--msg-format=metadata`). Closes the IM-vs-mail payload asymmetry agents currently see.
- **P2**: extract `mail +watch`'s fetch + filter + format business logic into a shared `internal/mail/eventfmt` package; expose `--msg-format` / `--folders` / `--labels` as Params on the EventKey; `mail +watch` retires to a thin alias forwarding to the unified path with sensible defaults. After P2 the two paths are fully interchangeable.
- **P3**: add a lint rule (`events/lint_test.go`) pinning `larkws` and `event/dispatcher` imports to `events/`, `cmd/event/`, `shortcuts/event/` only — physical defense against future per-domain forks like the one this PR removes the conceptual basis for.

## Test Plan

- [x] `make unit-test` passes (full suite green)
- [x] `go vet ./...` clean
- [x] `golangci-lint v2.1.6 run ./events/...` — 0 issues
- [x] `lark-cli event list` shows `mail.user_mailbox.event.message_received_v1` under `── mail ──`
- [x] `lark-cli event schema mail.user_mailbox.event.message_received_v1` shows `Pre-consume: yes`, scopes, console events, mailbox param, and the SDK-derived output schema
- [x] End-to-end manual: `lark-cli event consume mail.user_mailbox.event.message_received_v1 --as user --max-events 1 --timeout 10m` — PreConsume opens subscription, NDJSON event streams to stdout when a real email arrives, cleanup unsubscribes on graceful exit (`reason: limit`)
- [x] `mail +watch` still works (verified to confirm back-compat)

## Notes

- The existing `mail +watch` shortcut is intentionally left in place — it offers richer payload modes (`--msg-format minimal/metadata/plain_text_full/full`) and folder/label filters that this minimal P0 unified path doesn't yet expose. Both paths share the same server-side subscription record; P1/P2 above explicitly target full feature parity so this duality is temporary.
- Console event subscription `mail.user_mailbox.event.message_received_v1` must be enabled (and the app published) in the developer console for events to flow — same prerequisite as `mail +watch`; the runtime `RequiredConsoleEvents` predicate now warns if it is missing.


bubbmon233:
Thanks for the PR, the unified entry point is a nice direction! While reading through I had a few questions I wanted to raise — happy to discuss.

1. Subscription dimension mismatch: Mail subscriptions are keyed by mailbox. event consume is keyed by event type. When two event consume processes run with different mailboxes, the second one never opens a subscription — it just hangs, receives no mail, and reports no error.

2. Fewer features than mail +watch: event consume mail.xxx does not support --folders / --labels / --msg-format / --print-output-schema.

3. Unsubscribe failure is silent: On exit, if the unsubscribe API call fails, the new code swallows the error with _, _ = .... The cleanup function is func() with no return value, so the caller cannot observe the error. The server-side subscription leaks, and the user is never told.

dingding0418:
Thanks for the careful review — these are all real concerns. Let me set the framing first: this PR and the planned follow-ups are all driven by an **AI-friendly**
  north star. The interface should give callers three things: a single entry point, no silent failures, and no leaking of internal implementation (concepts like the bus
   daemon or `FirstForKey` shouldn't surface to the agent). The three issues you raised happen to land on each of those three principles. Replying to each:

  ---
  
  ### 1. Subscription dimension mismatch — real, framework-level

  **Root cause**: `internal/event/consume/consume.go` keys `FirstForKey` on the EventKey alone, but mail's subscription identity is two-dimensional — `(EventKey,
  mailbox)`. So with two `event consume mail.xxx` processes against the same bus daemon (one `mailbox=A`, one `mailbox=B`), the second one's `PreConsume` is skipped by
  the framework as "already covered" → B's subscribe API is never called. On top of that, the bus dispatch doesn't filter by params → P2 receives P1's events too. The
  result is silent hang + cross-pollination.

  **Wider observation**: mail is the first EventKey in lark-cli where `(EventKey, params)` (rather than EventKey alone) determines the subscription identity. IM doesn't
   have this property so the gap was never exposed. But this is a **general pattern** — drive comments (per file_token), task updates (per task_guid), and future
  calendar / approval EventKeys will all hit the exact same bug. Fixing it once at the framework level benefits every per-resource subscription EventKey we'll ever add.

  **Proposed fix (Option A)**: lift the dedup key from one dimension to two.

  - Add `SubscriptionKey bool` to `ParamDef`: marks which params change subscription identity (`mailbox=true`, `msg-format=false`).
  - Add `Match(params, raw) bool` to `KeyDefinition`: the bus uses it to filter delivery, eliminating cross-pollination (mail's Match compares `event.mail_address`
  against the resolved `params.mailbox`).
  - Hub's `keyCounts` becomes `map[(EventKey, paramsFingerprint)]int`.
  - The Hello protocol gains a `Params map[string]string` field (wire-level change).
  - IM is unchanged (no SubscriptionKey params, no Match → fingerprint is always empty → degenerates back to today's one-dimensional behavior, fully
  backward-compatible).

  `FirstForKey` isn't bypassed — it's just that with the lifted key, P1 (mailbox=A) and P2 (mailbox=B) are each "first" within their own sid, so subscribe API gets
  called per target. The WebSocket and Feishu's subscription table don't move; this is purely about aligning the framework's internal bookkeeping dimensions with the
  subscription table's.

  **Idempotency safety net**: Feishu's `event/subscribe` API is idempotent. Even if the sid computation has a bug, the worst case degrades to "call subscribe once per
  consumer" — still correct, just wasted API quota. Implementation risk is low.

  ### 2. Feature parity gap — same abstraction solves it alongside #1

  `event consume mail.xxx` currently lacks `--folders / --labels / --msg-format / --print-output-schema`, leaving the agent experience fragmented.

  **Key observation**: every missing flag maps cleanly to an `event consume -p param=...`. Once #1's framework upgrade lands, #2 falls out naturally:

  | mail +watch flag | event consume equivalent | SubscriptionKey | Implemented in which hook |
  |---|---|---|---|
  | `--mailbox` | `-p mailbox=...` | true | PreConsume: subscribe against the corresponding mailbox |
  | `--msg-format` | `-p msg-format=metadata\|full\|...` | false | Process: per-event call to messages API to enrich the output |
  | `--folders` / `--folder-ids` | `-p folders=...` | false | Match: fetch metadata via message_id, drop if folder doesn't match |
  | `--labels` / `--label-ids` | `-p labels=...` | false | Match (same pattern) |
  | `--print-output-schema` | `event schema mail.xxx -p msg-format=...` | — | The schema command reflects the chosen format |

  Three hooks, three concerns:
  - **PreConsume** (handles SubscriptionKey-true params) → subscription lifecycle
  - **Match** (handles delivery-filter params) → consumers share the subscription, each filters locally
  - **Process** (handles output-format params) → per-event enrichment

  After this, `mail +watch` becomes a thin alias forwarding to `event consume mail.xxx -p msg-format=metadata -p ...` (or a deprecation notice). The two paths become 
  functionally interchangeable, and from the agent's perspective the unification is real.

  ### 3. Silent unsubscribe failure — fixing now, Option A

  Fully agreed, this is a real bug. The current cleanup is `func()` with no return value, and `_, _ = rt.CallAPI(...)` swallows the error. Worse, the framework prints 
  `[event] cleanup done.` regardless of actual outcome — agents reading that line will assume success when in fact a leak may have occurred.

  **Option A**: change the framework's cleanup signature to return error.

  ```go
  // internal/event/types.go
  PreConsume func(ctx, rt, params) (cleanup func() error, err error)

  // internal/event/consume/consume.go
  case lastForKey:
      fmt.Fprintf(errOut, "[event] running cleanup...\n")
      if err := cleanup(); err != nil {
          fmt.Fprintf(errOut,
              "[event] WARN: cleanup failed: %v "+
              "(server-side subscription may have leaked)\n", err)
      } else {
          fmt.Fprintf(errOut, "[event] cleanup done.\n")
      }

  mail cleanup implementation:

  cleanup := func() error {
      cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
      defer cancel()
      if _, err := rt.CallAPI(cleanupCtx, "POST",
          mailboxEventPath(mailbox, "unsubscribe"), body); err != nil {
          return fmt.Errorf("unsubscribe mailbox=%s: %w", mailbox, err)
      }
      return nil
  }

  Wins: errors are explicit, the WARN line has a uniform format, the misleading "cleanup done" disappears on failure, and any future EventKey using PreConsume gets it
  for free.

  Will add a regression test: fakeClient returns a non-nil unsubErr, asserting cleanup() returns the wrapped error and the framework writes the WARN line.

PR #857, opened 2026-05-13
feat: add markdown +patch shortcut

wittam-01:
## Summary
Add a new `markdown +patch` shortcut that updates a Markdown file in Drive via a fetch-local-replace-overwrite flow, enabling targeted edits without rewriting the entire file. Supports both literal and RE2 regex matching modes.

## Changes
- Add `shortcuts/markdown/markdown_patch.go` implementing the `markdown +patch` command and register it in `shortcuts/markdown/shortcuts.go`
- Extract shared file-name resolution and overwrite-upload helpers into `shortcuts/markdown/helpers.go` and refactor `markdown_overwrite.go` to reuse them
- Add unit tests in `markdown_patch_test.go` covering literal/regex modes, zero-match skip-upload, and flag validation
- Add dry-run and workflow E2E cases under `tests/cli_e2e/markdown/`
- Update README, `lark-markdown` skill docs, and add `skills/lark-markdown/references/lark-markdown-patch.md`

## Test Plan
- [x] Unit tests pass (`make unit-test`)
- [x] Manual local verification confirms the `lark markdown +patch` command works as expected (literal, regex, and zero-match scenarios)

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * Added markdown +patch: local literal or RE2 regex edits for Drive-native .md files; reports match_count, only uploads when matches > 0, and rejects edits that would produce an empty file.

* **Documentation**
  * Skill bumped to v1.1.0 and docs (EN/CN/skill refs) updated with patch semantics, flags, examples, dry-run behavior, and JSON response fields; README updated.

* **Tests**
  * Added unit and e2e tests covering validation, literal/regex modes, dry-run, and lifecycle patching.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/857?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


fangshuyu-768 (commented):
## Review Summary

Thanks for the PR! The `+patch` shortcut is a nice addition. Here are the issues I found:

| Priority | Issue | File |
|----------|-------|------|
| **High** | No HTTP status code check after download — error responses could be treated as Markdown and uploaded, corrupting the file | `helpers.go` |
| **Medium** | `$` injection in regex replacement with no escape mechanism | `markdown_patch.go` |
| **Medium** | Regex compiled twice (Validate + Execute) | `markdown_patch.go` |
| **Medium** | TOCTOU: no concurrency protection between download and upload | `markdown_patch.go` |
| **Low** | Inconsistent required-flag checking style | `markdown_patch.go` |
| **Low** | Download filename discarded then re-fetched via extra API call | `markdown_patch.go` |
| **Low** | `match_count` type roundtrip through float | `markdown_patch.go` |

The **High** priority item (missing HTTP status check) could lead to data corruption and should be addressed before merge.

shortcuts/markdown/markdown_patch.go
```diff
@@ -0,0 +1,235 @@
...
+		ContentSet: runtime.Changed("content"),
+		Regex:      runtime.Bool("regex"),
+	}
+}
+
+func validateMarkdownPatchSpec(runtime *common.RuntimeContext, spec markdownPatchSpec) error {
+	if err := validate.ResourceName(spec.FileToken, "--file-token"); err != nil {
+		return output.ErrValidation("%s", err)
+	}
+	if !runtime.Changed("pattern") {
```
fangshuyu-768, line 197: **[Medium] `$` injection in regex replacement — no way to escape**

`re.ReplaceAllString` interprets `$1`, `$0`, etc. as capture-group references. If a user wants to replace text with a literal `$100`, the `$1` prefix will be silently interpreted as a capture reference, producing unexpected results.

The docs and examples explicitly show `$1` capture substitution, so `ReplaceAllString` is intentional — but there's no escape mechanism or opt-out.

**Suggestion:** Either:
1. Add a `--literal-replacement` flag that uses `re.ReplaceAllLiteralString` instead, or
2. Document this behavior prominently in the `--regex` flag description and the reference doc.

shortcuts/markdown/markdown_patch.go
```diff
@@ -0,0 +1,235 @@
...
+			Body(map[string]interface{}{
+				"request_docs": []map[string]interface{}{
+					{
+						"doc_token": spec.FileToken,
+						"doc_type":  "file",
+					},
+				},
+			}).
+			POST("/open-apis/drive/v1/files/upload_all").
+			Desc("[3a] If the patched Markdown is at most "+sizeThreshold+", overwrite the file with multipart/form-data upload_all").
```
fangshuyu-768, line 80: **[Medium] Regex compiled twice — once in Validate, once in Execute**

`Validate` compiles the regex to check validity:
```go
if _, err := regexp.Compile(spec.Pattern); err != nil {
```

Then `applyMarkdownPatch` compiles it again:
```go
re, err := regexp.Compile(spec.Pattern)
```

**Suggestion:** Store the compiled `*regexp.Regexp` in `markdownPatchSpec` during `Validate`, and pass it into `applyMarkdownPatch` to avoid redundant compilation.

shortcuts/markdown/helpers.go
```diff
@@ -112,6 +113,42 @@ func finalMarkdownFileName(spec markdownUploadSpec) string {
...
+	if err != nil {
+		return nil, "", output.ErrNetwork("download failed: %s", err)
+	}
+	return resp, fileNameFromDownloadHeader(resp.Header, fileToken+".md"), nil
+}
+
+func validateNonEmptyMarkdownSize(size int64) error {
+	if size == 0 {
+		return output.ErrValidation("%s", markdownEmptyContentError)
+	}
```
fangshuyu-768, line 148: **[High] No HTTP status code check after download**

`openMarkdownDownload` returns the raw `*http.Response` without checking `resp.StatusCode`. If the download API returns 404/403/etc., `resp.Body` will contain an error JSON instead of Markdown content.

In `markdown_patch.go` Execute, this error JSON would be treated as valid Markdown, passed through `applyMarkdownPatch`, and potentially uploaded — **overwriting the original file with garbage**.

**Suggestion:** Add a status code check in `openMarkdownDownload` (or at the call site) and return an error for non-2xx responses:
```go
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    body, _ := io.ReadAll(resp.Body)
    resp.Body.Close()
    return nil, "", output.ErrAPI("download failed (HTTP %d): %s", resp.StatusCode, string(body))
}
```

shortcuts/markdown/markdown_patch.go
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			return err
+		}
+
+		mode := markdownPatchModeLiteral
+		if spec.Regex {
+			mode = markdownPatchModeRegex
+		}
+
+		out := map[string]interface{}{
```
fangshuyu-768, line 138: **[Medium] TOCTOU — no concurrency protection between download and upload**

The flow is: download → local replace → upload overwrite. Between download and upload, another client could modify the file, and this patch would silently overwrite their changes, causing data loss.

The docs mention "not server-side atomic patch", but there's no version/etag check at all.

**Suggestion:** 
1. Record the file version from the download response, and pass it as a precondition during upload (optimistic concurrency control), or
2. Add a prominent warning in the reference doc about this race condition risk.

shortcuts/markdown/markdown_patch.go
```diff
@@ -0,0 +1,235 @@
...
+			POST("/open-apis/drive/v1/files/upload_part").
+			Desc("[3c] Upload file parts (repeated) when multipart overwrite is required").
+			Body(map[string]interface{}{
+				"upload_id": "<upload_id>",
+				"seq":       "<chunk_index>",
+				"size":      "<chunk_size>",
+				"file":      "<chunk_binary>",
+			}).
+			POST("/open-apis/drive/v1/files/upload_finish").
+			Desc("[3d] Finalize multipart overwrite upload and return the new version").
```
fangshuyu-768, line 107: **[Low] Inconsistent style for checking required flags**

`--pattern` uses `runtime.Changed("pattern")`, while `--content` uses `spec.ContentSet` (set via `runtime.Changed("content")` in `newMarkdownPatchSpec`). Both work, but the style is inconsistent.

**Suggestion:** Unify the approach — either store `PatternSet bool` in the spec like `ContentSet`, or use `runtime.Changed` for both in `validateMarkdownPatchSpec`.

shortcuts/markdown/markdown_patch.go
fangshuyu-768, line 138: **[Low] Download filename is discarded, then re-fetched via extra API call**

`openMarkdownDownload` returns the filename parsed from `Content-Disposition` as its second return value, but it's always ignored:
```go
resp, _, err := openMarkdownDownload(ctx, runtime, spec.FileToken)
```

Then `resolveMarkdownOverwriteFileName` makes another `batch_query` API call to get the same filename.

**Suggestion:** Either use the filename from the download response header (avoiding the extra `batch_query` call), or remove the unused return value from `openMarkdownDownload`.

shortcuts/markdown/markdown_patch.go
```diff
@@ -0,0 +1,235 @@
...
+			return err
+		}
+
+		specUpload := markdownUploadSpec{
+			FileToken: spec.FileToken,
+		}
+		fileName, err := resolveMarkdownOverwriteFileName(runtime, specUpload)
+		if err != nil {
+			return err
+		}
```
fangshuyu-768, line 164: **[Low] `match_count` type roundtrip through float**

`match_count` is stored as `int` in the map, but `prettyPrintMarkdownPatch` reads it via `common.GetFloat` then casts back to `int`:
```go
fmt.Fprintf(w, "match_count: %d
", int(common.GetFloat(data, "match_count")))
```

This works because JSON unmarshals numbers as float64, but it's a code smell.

**Suggestion:** Use `common.GetInt` if available, or store the value as a consistent type to avoid the int→float→int roundtrip.

shortcuts/markdown/helpers.go
wittam-01: Checked this one again.

`openMarkdownDownload` delegates to `RuntimeContext.DoAPIStream`, and `DoAPIStream` already relies on `APIClient.DoStream` to convert `status >= 400` into an error before the response is returned to the caller. So the patch flow never receives a non-2xx body as Markdown input.

I kept the helper thin here and did not add a second status-code check.

shortcuts/markdown/markdown_patch.go
wittam-01: Intentional behavior. In regex mode we do want Go replacement-template semantics because `$1` capture substitution is part of the documented and tested contract for this shortcut.

I updated the reference doc in the latest push to make the escape path explicit: if callers need a literal `$`, they should write `$$`. I did not add a separate `--literal-replacement` flag in this PR to keep the surface area small.

shortcuts/markdown/markdown_patch.go
wittam-01: Fair point. I looked at this, but I left it as-is for this PR. The first compile keeps regex validation failures in `Validate`, while execution still needs a compiled regexp in the actual patch path. Carrying a compiled `*regexp.Regexp` through the shortcut spec/runtime boundary would make this change larger than the benefit here, so I prefer to keep it as a follow-up cleanup if we want it.

shortcuts/markdown/markdown_patch.go
wittam-01: Agreed on the limitation. I did not add optimistic concurrency in this PR, but I updated the reference doc to make the boundary explicit: this flow is `download -> local replace -> overwrite`, it is not atomic, and concurrent edits can overwrite an intermediate update.

If we want true optimistic concurrency here, that should be a separate capability change in the overwrite path rather than piggybacking on this shortcut addition.

shortcuts/markdown/markdown_patch.go
wittam-01: Agreed that the style is slightly inconsistent. I left it unchanged in this PR to avoid mixing a pure style cleanup into the behavior/documentation follow-up. Happy to normalize the flag-state fields in a separate small cleanup if we want to do that afterward.

shortcuts/markdown/markdown_patch.go
wittam-01: Addressed in the latest push. Since the filename return value was not actually consumed by the patch flow, I removed that unused return value from `openMarkdownDownload` rather than keeping a misleading extra output.

shortcuts/markdown/markdown_patch.go
wittam-01: Addressed in the latest push. I added `common.GetInt` and switched the pretty-print path away from the `GetFloat -> int(...)` roundtrip for `match_count` and the size fields.

fangshuyu-768 (approved):

PR #883, opened 2026-05-14
feat: support wiki node target in markdown +create

wittam-01:
## Summary
  Add `--wiki-token` support to `markdown +create` so Markdown files can be created directly under a Wiki node instead of only Drive folders. The implementation stays scoped to `shortcuts/markdown` and aligns the command behavior with existing Drive wiki-upload semantics.

  ## Changes
  - Add `--wiki-token` support, target validation, and `parent_type=wiki` routing to `markdown +create`
  - Omit the fallback `/file/<token>` URL for wiki-hosted Markdown files, and update command help / skill reference docs
  - Add unit tests, markdown dry-run E2E coverage, and a wiki-parent live workflow test

  ## Test Plan
  - [x] Unit tests pass (`go test ./shortcuts/markdown -count=1`, `go test ./tests/cli_e2e/markdown -count=1`, `make unit-test`)
  - [x] Manual local verification confirms the `lark xxx` command works as expected (`./lark-cli markdown +create --as bot --wiki-token ...` with both `--content` and `--file`, followed by `./lark-cli markdown +fetch`)

  ## Related Issues
  - None

## Summary by CodeRabbit

* **New Features**
  * Create Markdown files under Wiki nodes; dry-run and upload flows include wiki parent info and omit a file URL for Wiki-hosted files.

* **Bug Fixes**
  * Stronger flag validation: rejects empty/whitespace tokens, disallows supplying both wiki and folder targets, and validates resource names for the chosen target.

* **Documentation**
  * Updated docs and examples to show --wiki-token and its mutual-exclusivity with --folder-token.

* **Tests**
  * Added unit and E2E tests covering wiki-target validation, dry-run payloads, uploads, and workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/883?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/markdown/helpers.go
+++ b/shortcuts/markdown/helpers.go
@@ -24,10 +24,16 @@ import (
 const markdownSinglePartSizeLimit = common.MaxDriveMediaUploadSinglePartSize
 const markdownEmptyContentError = "empty markdown content is not supported; cannot create or overwrite an empty file"
 
+const (
+	markdownUploadParentTypeExplorer = "explorer"
+	markdownUploadParentTypeWiki     = "wiki"
+)
+
 type markdownUploadSpec struct {
 	FileToken   string
 	FileName    string
 	FolderToken string
+	WikiToken   string
 	FilePath    string
 	Content     string
 	ContentSet  bool
@@ -45,6 +51,25 @@ type markdownMultipartSession struct {
 	BlockNum  int
 }
 
+type markdownUploadTarget struct {
+	ParentType string
+	ParentNode string
+}
+
+func (spec markdownUploadSpec) Target() markdownUploadTarget {
+	if spec.WikiToken != "" {
+		return markdownUploadTarget{
+			ParentType: markdownUploadParentTypeWiki,
+			ParentNode: spec.WikiToken,
+		}
+	}
+	// An empty explorer parent node uploads to the user's Drive root folder.
+	return markdownUploadTarget{
+		ParentType: markdownUploadParentTypeExplorer,
+		ParentNode: spec.FolderToken,
+	}
+}
+
 func validateMarkdownSpec(runtime *common.RuntimeContext, spec markdownUploadSpec, requireName bool) error {
 	switch {
 	case spec.ContentSet && spec.FileSet:
@@ -53,11 +78,29 @@ func validateMarkdownSpec(runtime *common.RuntimeContext, spec markdownUploadSpe
 		return common.FlagErrorf("specify exactly one of --content or --file")
 	}
 
-	if runtime.Changed("folder-token") && strings.TrimSpace(spec.FolderToken) == "" {
+	if markdownFlagExplicitlyEmpty(runtime, "folder-token") {
 		return common.FlagErrorf("--folder-token cannot be empty; omit it to upload into Drive root folder")
+	}
+	if markdownFlagExplicitlyEmpty(runtime, "wiki-token") {
+		return common.FlagErrorf("--wiki-token cannot be empty; provide a valid wiki node token or omit the flag entirely")
+	}
+	targets := 0
+	if spec.FolderToken != "" {
+		targets++
+	}
+	if spec.WikiToken != "" {
+		targets++
+	}
+	if targets > 1 {
+		return common.FlagErrorf("--folder-token and --wiki-token are mutually exclusive")
 	}
 	if spec.FolderToken != "" {
 		if err := validate.ResourceName(spec.FolderToken, "--folder-token"); err != nil {
+			return output.ErrValidation("%s", err)
+		}
+	}
+	if spec.WikiToken != "" {
+		if err := validate.ResourceName(spec.WikiToken, "--wiki-token"); err != nil {
 			return output.ErrValidation("%s", err)
 		}
 	}
@@ -90,6 +133,10 @@ func validateMarkdownSpec(runtime *common.RuntimeContext, spec markdownUploadSpe
 	}
 
 	return nil
+}
+
+func markdownFlagExplicitlyEmpty(runtime *common.RuntimeContext, flagName string) bool {
+	return runtime.Changed(flagName) && strings.TrimSpace(runtime.Str(flagName)) == ""
 }
 
 func validateMarkdownFileName(name, flagName string) error {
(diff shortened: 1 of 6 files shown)


tests/cli_e2e/markdown/markdown_dryrun_test.go
```diff
@@ -96,6 +123,27 @@ func TestMarkdownCreateDryRun_RejectsEmptyContent(t *testing.T) {
...
+			"--content", "# hello",
+			"--wiki-token", "",
+			"--dry-run",
+		},
+		DefaultAs: "bot",
+	})
+	require.NoError(t, err)
+	result.AssertExitCode(t, 2)
+	assert.Contains(t, result.Stderr, "--wiki-token cannot be empty")
+}
```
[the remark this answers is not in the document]
wittam-01: Fixed in d23181aa: the assertion now checks result.Stdout+result.Stderr while keeping the non-zero exit assertion. Verified with go test ./tests/cli_e2e/markdown -count=1.

tests/cli_e2e/markdown/markdown_workflow_test.go
```diff
@@ -98,3 +99,84 @@ func TestMarkdownLifecycleWorkflow(t *testing.T) {
...
+			DefaultAs: identity,
+		})
+		cleanupCancel()
+		if err == nil && result != nil && result.ExitCode == 0 {
+			return
+		}
+	}
+
+	parentT.Logf("cleanup skipped: could not delete wiki-hosted markdown file %s with either bot or user identity", fileToken)
+}
```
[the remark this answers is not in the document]
wittam-01: Fixed in d23181aa: cleanup now fails the E2E with Fatalf if neither bot nor user can delete the created wiki-hosted Markdown file. Verified with go test ./tests/cli_e2e/markdown -count=1.

shortcuts/markdown/helpers.go
```diff
@@ -92,6 +134,10 @@ func validateMarkdownSpec(runtime *common.RuntimeContext, spec markdownUploadSpe
 	return nil
 }
```
fangshuyu-768, line 137: The original code checked `spec.FolderToken` directly, but the new helper uses `runtime.Str(flagName)` instead. Since `spec.FolderToken` is already trimmed at construction time (`strings.TrimSpace(runtime.Str("folder-token"))`), the values should be equivalent here — but it's worth double-checking that `runtime.Str()` doesn't apply any additional transformations. If there's any divergence, this refactor could subtly change the `--folder-token` empty-check behavior.

shortcuts/markdown/markdown_create.go
```diff
@@ -79,8 +87,10 @@ var MarkdownCreate = common.Shortcut{
 			"file_name":  finalMarkdownFileName(spec),
 			"size_bytes": fileSize,
 		}
-		if u := common.BuildResourceURL(runtime.Config.Brand, "file", result.FileToken); u != "" {
-			out["url"] = u
+		if target := spec.Target(); target.ParentType == markdownUploadParentTypeExplorer {
+			if u := common.BuildResourceURL(runtime.Config.Brand, "file", result.FileToken); u != "" {
```
fangshuyu-768, line 91: When `parent_type=wiki`, the file's permission model differs from Drive. Calling `AutoGrantCurrentUserDrivePermission` on a wiki-hosted file may be a no-op or could even return an error. Should this also be gated behind `target.ParentType == markdownUploadParentTypeExplorer`, similar to the URL logic above?

shortcuts/markdown/helpers.go
```diff
@@ -45,6 +51,24 @@ type markdownMultipartSession struct {
...
+type markdownUploadTarget struct {
+	ParentType string
+	ParentNode string
+}
+
+func (spec markdownUploadSpec) Target() markdownUploadTarget {
+	if spec.WikiToken != "" {
+		return markdownUploadTarget{
+			ParentType: markdownUploadParentTypeWiki,
+			ParentNode: spec.WikiToken,
```
fangshuyu-768, line 63: Minor suggestion: consider adding a brief comment here explaining the fallback behavior — when both `WikiToken` and `FolderToken` are empty, the target defaults to `explorer` with an empty `ParentNode`, which means uploading to the Drive root folder. This implicit semantics may not be obvious to future readers.

tests/cli_e2e/markdown/markdown_workflow_test.go
```diff
@@ -181,3 +181,84 @@ func TestMarkdownLifecycleWorkflow(t *testing.T) {
...
+
+func bestEffortDeleteWikiHostedMarkdownFile(parentT *testing.T, fileToken string) {
+	parentT.Helper()
+
+	request := clie2e.Request{
+		Args: []string{
+			"drive", "+delete",
+			"--file-token", fileToken,
+			"--type", "file",
+			"--yes",
```
fangshuyu-768, line 247: The function name says "best effort" but the implementation calls `parentT.Fatalf` when both cleanup attempts fail, which will fail the test immediately. Consider either:
1. Renaming to `requireDeleteWikiHostedMarkdownFile` to match the actual behavior, or
2. Changing `Fatalf` to `Logf` + `return` if cleanup failure should truly be best-effort.

fangshuyu-768 (commented):
Thanks for the PR! Overall the implementation is clean and well-tested. I have a few comments below — two are worth investigating before merge, the rest are minor suggestions.

shortcuts/markdown/markdown_create.go
wittam-01: Keeping this as-is. In the wiki-parent upload flow the API still returns a file token, not a wiki node token, so granting Drive permission on the returned file token is intentional here. The URL logic is gated because the explorer URL builder is not valid for wiki-hosted output, but the permission target remains the returned file resource.

shortcuts/markdown/helpers.go
wittam-01: Verified. RuntimeContext.Str() is a thin Cobra GetString() accessor and does not apply extra transformations. The constructed spec field still uses strings.TrimSpace(runtime.Str("folder-token")), and markdownFlagExplicitlyEmpty applies the same trim only to detect explicitly empty/whitespace values. So this does not change the previous --folder-token empty-check behavior.

shortcuts/markdown/helpers.go
wittam-01: Addressed in 95f569df. Added a short comment on Target() documenting that an empty explorer parent node means uploading to the Drive root folder.

tests/cli_e2e/markdown/markdown_workflow_test.go
wittam-01: Addressed in 95f569df. Renamed the helper to requireDeleteWikiHostedMarkdownFile so the name matches the Fatalf behavior when cleanup cannot delete the created file.

fangshuyu-768 (approved):

PR #904, opened 2026-05-15
feat(wiki): add +node-get / +node-delete / +space-create shortcuts

liujinkun2025:
## Summary

Fills the wiki node/space CRUD gaps. Three new shortcuts plus a shared async-task helper:

- **`wiki +node-get`** — wraps `wiki.spaces.get_node`. Accepts a raw `node_token` / `obj_token` or a Lark URL (URL path auto-infers `obj_type`; conflicting explicit `--obj-type` is rejected). Optional `--space-id` cross-check. Formatted output: `title / obj_type / obj_token / node_token / space_id / parent_node_token / creator / updated_at / url`.
- **`wiki +node-delete`** — wraps the `space.node` delete API. `high-risk-write` (`--yes` gated), bounded async `delete_node` task polling, auto-resolves `space_id` via `get_node` when `--space-id` is omitted, `--include-children` defaults to `true`. Maps Lark codes `131011` (UI-approval required) and `131003` (subtree too large) to actionable hints; other codes pass through with the original code preserved.
- **`wiki +space-create`** — wraps `spaces.create`. User-only identity (the API rejects tenant/bot tokens; framework `CheckIdentity` enforces it). `--name` required (no empty-name spaces), `--description` optional. Flattened space output, no `url` (the API returns none and there is no space-home URL builder).
- **Refactor** — extracted the duplicated wiki delete-task poll loop / status classification into `wiki_async_task.go` (`pollWikiAsyncTask` / `parseWikiAsyncTaskStatus` / `wikiAsyncTaskStatus`). `+delete-space` now delegates to it via thin wrappers — no behavior change, ~110 lines of duplication removed.

Also renames the originally-proposed `+get-node` to `+node-get` for consistency with the `+node-*` family.

Risk-level note: `spaces.create` carries `danger:true` in the registry, but that flag only drives the `lark-cli schema` browser's red `[danger]` tag — it is **not** a runtime confirmation gate (only `risk:high-risk-write` is). The raw API runs without `--yes` today, so `+space-create` at `Risk:write` is runtime-consistent; a help Tip surfaces the danger flag and points to `+delete-space` for undo.

## Test plan

- [x] `go test ./shortcuts/...` (and full suite excluding `tests/cli_e2e`) green
- [x] `gofmt -l` clean, `go vet ./shortcuts/wiki/...` clean, `./build.sh` OK
- [x] New unit + httpmock tests: URL/token parsing, obj_type inference & conflict, space-id auto-resolve, sync/async/timeout/failure delete paths, error-code hint mapping, `--yes` gating, bot-identity rejection, output flattening
- [x] Manual smoke: `--help`, `--dry-run`, validation failures, identity rejection for each shortcut
- [ ] `golangci-lint` not run (not installed in dev env) — please confirm in CI
- [ ] `tests/cli_e2e/wiki` not run locally (hits real Feishu) — verify in CI/staging

## Summary by CodeRabbit

* **New Features**
  * Added wiki commands: +node-get, +node-delete, +space-create; shortcuts list expanded
  * Drive task result now supports wiki_delete_node scenario

* **Improvements**
  * Centralized async-task polling for wiki operations with normalized status handling, clearer retry/resume hints, and consistent timeout/failure behavior
  * Input validation and user-facing output formatting improved for wiki commands

* **Tests**
  * Extensive tests covering new commands, async polling, error hints, dry-run and mounted executions

* **Documentation**
  * New/updated reference pages for node-get, node-delete, and space-create

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/904?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/drive/drive_task_result.go
+++ b/shortcuts/drive/drive_task_result.go
@@ -20,7 +20,7 @@ import (
 var DriveTaskResult = common.Shortcut{
 	Service:     "drive",
 	Command:     "+task_result",
-	Description: "Poll async task result for import, export, drive move/delete, wiki move, or wiki delete-space operations",
+	Description: "Poll async task result for import, export, drive move/delete, wiki move, wiki delete-space, or wiki delete-node operations",
 	Risk:        "read",
 	// This shortcut multiplexes multiple backend APIs with different scope
 	// requirements, so scenario-specific prechecks are handled in Validate.
@@ -28,8 +28,8 @@ var DriveTaskResult = common.Shortcut{
 	AuthTypes: []string{"user", "bot"},
 	Flags: []common.Flag{
 		{Name: "ticket", Desc: "async task ticket (for import/export tasks)", Required: false},
-		{Name: "task-id", Desc: "async task ID (for drive task_check, wiki_move, or wiki_delete_space tasks)", Required: false},
-		{Name: "scenario", Desc: "task scenario: import, export, task_check, wiki_move, or wiki_delete_space", Required: true},
+		{Name: "task-id", Desc: "async task ID (for drive task_check, wiki_move, wiki_delete_space, or wiki_delete_node tasks)", Required: false},
+		{Name: "scenario", Desc: "task scenario: import, export, task_check, wiki_move, wiki_delete_space, or wiki_delete_node", Required: true},
 		{Name: "file-token", Desc: "source document token used for export task status lookup", Required: false},
 	},
 	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
@@ -40,9 +40,10 @@ var DriveTaskResult = common.Shortcut{
 			"task_check":        true,
 			"wiki_move":         true,
 			"wiki_delete_space": true,
+			"wiki_delete_node":  true,
 		}
 		if !validScenarios[scenario] {
-			return output.ErrValidation("unsupported scenario: %s. Supported scenarios: import, export, task_check, wiki_move, wiki_delete_space", scenario)
+			return output.ErrValidation("unsupported scenario: %s. Supported scenarios: import, export, task_check, wiki_move, wiki_delete_space, wiki_delete_node", scenario)
 		}
 
 		// Validate required params based on scenario
@@ -54,7 +55,7 @@ var DriveTaskResult = common.Shortcut{
 			if err := validate.ResourceName(runtime.Str("ticket"), "--ticket"); err != nil {
 				return output.ErrValidation("%s", err)
 			}
-		case "task_check", "wiki_move", "wiki_delete_space":
+		case "task_check", "wiki_move", "wiki_delete_space", "wiki_delete_node":
 			if runtime.Str("task-id") == "" {
 				return output.ErrValidation("--task-id is required for %s scenario", scenario)
 			}
@@ -108,6 +109,11 @@ var DriveTaskResult = common.Shortcut{
 				Desc("[1] Query wiki delete-space task result").
 				Set("task_id", taskID).
 				Params(map[string]interface{}{"task_type": "delete_space"})
+		case "wiki_delete_node":
+			dry.GET("/open-apis/wiki/v2/tasks/:task_id").
+				Desc("[1] Query wiki delete-node task result").
+				Set("task_id", taskID).
+				Params(map[string]interface{}{"task_type": "delete_node"})
 		}
 
 		return dry
@@ -136,6 +142,8 @@ var DriveTaskResult = common.Shortcut{
 			result, err = queryWikiMoveTask(runtime, taskID)
 		case "wiki_delete_space":
 			result, err = queryWikiDeleteSpaceTask(runtime, taskID)
+		case "wiki_delete_node":
+			result, err = queryWikiDeleteNodeTask(runtime, taskID)
 		}
 
 		if err != nil {
@@ -236,7 +244,7 @@ func validateDriveTaskResultScopes(ctx context.Context, runtime *common.RuntimeC
 	switch scenario {
 	case "import", "export", "task_check":
 		required = []string{"drive:drive.metadata:readonly"}
-	case "wiki_move", "wiki_delete_space":
+	case "wiki_move", "wiki_delete_space", "wiki_delete_node":
 		required = []string{"wiki:space:read"}
 	}
 
@@ -538,5 +546,66 @@ func queryWikiDeleteSpaceTask(runtime *common.RuntimeContext, taskID string) (ma
 		"failed":     failed,
 		"status":     resolvedStatus,
 		"status_msg": label,
+	}, nil
+}
+
+// queryWikiDeleteNodeTask returns the normalized status of an async wiki
+// delete-node task. For historical reasons the gateway stashes delete-node
+// status under the generic `simple_task_result` key (NOT `delete_node_result`),
+// and that object only carries `status` — there is no `status_msg`, so the
+// label falls back to the status code. Mirrors queryWikiDeleteSpaceTask;
+// intentionally duplicated here (rather than importing the wiki package) to
+// keep drive from depending on shortcuts/wiki.
+func queryWikiDeleteNodeTask(runtime *common.RuntimeContext, taskID string) (map[string]interface{}, error) {
+	if err := validate.ResourceName(taskID, "--task-id"); err != nil {
+		return nil, output.ErrValidation("%s", err)
+	}
+
+	data, err := runtime.CallAPI(
+		"GET",
+		fmt.Sprintf("/open-apis/wiki/v2/tasks/%s", validate.EncodePathSegment(taskID)),
+		map[string]interface{}{"task_type": "delete_node"},
+		nil,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	task := common.GetMap(data, "task")
+	if task == nil {
+		return nil, output.Errorf(output.ExitAPI, "api_error", "wiki task response missing task")
+	}
+
+	resolvedTaskID := common.GetString(task, "task_id")
+	if resolvedTaskID == "" {
+		resolvedTaskID = taskID
+	}
+
+	result := common.GetMap(task, "simple_task_result")
+	var status string
+	if result != nil {
+		status = common.GetString(result, "status")
+	}
+
+	// Keep in sync with wiki.parseWikiAsyncTaskStatus / wikiAsyncTaskStatus
+	// classification (intentionally duplicated to avoid a drive→wiki import —
+	// see the doc comment above). If the success/failed/processing rules change
+	// there, mirror the change here.
+	lowered := strings.ToLower(strings.TrimSpace(status))
+	ready := lowered == "success"
+	failed := lowered == "failure" || lowered == "failed"
+
+	resolvedStatus := strings.TrimSpace(status)
+	if resolvedStatus == "" {
+		resolvedStatus = "processing"
+	}
+
+	return map[string]interface{}{
+		"scenario":   "wiki_delete_node",
+		"task_id":    resolvedTaskID,
+		"ready":      ready,
+		"failed":     failed,
+		"status":     resolvedStatus,
+		"status_msg": resolvedStatus,
 	}, nil
 }
--- a/shortcuts/wiki/wiki_async_task.go
+++ b/shortcuts/wiki/wiki_async_task.go
@@ -1,0 +1,207 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package wiki
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/larksuite/cli/internal/output"
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+// Shared async-task polling for wiki delete operations. The wiki delete
+// endpoints (DELETE /spaces/{id}, DELETE /spaces/{id}/nodes/{token}) may
+// return either an empty task_id (sync completion) or a task_id that must
+// be polled against /wiki/v2/tasks/{task_id}?task_type=<...>.
+//
+// For historical reasons /wiki/v2/tasks/{task_id} stashes the status under a
+// different key per task type: delete-space uses `delete_space_result`, while
+// delete-node uses the generic `simple_task_result` (the gateway's reusable
+// "future async tasks share this" field). move tasks use `move_result` and are
+// handled separately in wiki_move.go. Every key still exposes a `status`, so
+// the poll loop / classification is factored out here and the caller passes
+// the right result key.
+//
+// Note: `simple_task_result` only carries `status` (no `status_msg`), so for
+// delete-node StatusLabel() falls back to the status code — which is fine.
+
+const (
+	wikiAsyncStatusSuccess    = "success"
+	wikiAsyncStatusFailure    = "failure"
+	wikiAsyncStatusProcessing = "processing"
+
+	wikiAsyncTaskTypeDeleteSpace = "delete_space"
+	wikiAsyncTaskTypeDeleteNode  = "delete_node"
+
+	wikiAsyncResultDeleteSpace = "delete_space_result"
+	// wikiAsyncResultSimpleTask is the generic result key the gateway uses for
+	// delete-node (and intends to reuse for future async task types). It is
+	// NOT `delete_node_result` — that key does not exist in the response.
+	wikiAsyncResultSimpleTask = "simple_task_result"
+)
+
+// wikiAsyncTaskStatus is the unified poll-response shape used by every wiki
+// delete task. The taskID is captured so error/resume hints can name it.
+type wikiAsyncTaskStatus struct {
+	TaskID    string
+	Status    string
+	StatusMsg string
+}
+
+// normalizedStatus collapses whitespace and case so "  SUCCESS  " classifies
+// the same as "success". Ready()/Failed() (control flow) derive from this;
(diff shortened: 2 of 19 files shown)


shortcuts/wiki/wiki_async_task.go
```diff
@@ -0,0 +1,176 @@
...
+			label, taskID, nextCommand,
+		)
+		var exitErr *output.ExitError
+		if errors.As(lastErr, &exitErr) && exitErr.Detail != nil {
+			if strings.TrimSpace(exitErr.Detail.Hint) != "" {
+				hint = exitErr.Detail.Hint + "\n" + hint
+			}
+			return lastStatus, false, output.ErrWithHint(exitErr.Code, exitErr.Detail.Type, exitErr.Detail.Message, hint)
+		}
+		return lastStatus, false, output.ErrWithHint(output.ExitAPI, "api_error", lastErr.Error(), hint)
```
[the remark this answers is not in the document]
liujinkun2025: Fixed in `2cfd336`. The poll-exhaustion path now hand-builds the `*output.ExitError` (preserving `Detail.Code`/`ConsoleURL`/`Risk`/nested `Detail`) instead of the lossy `ErrWithHint`, matching `wrapWikiNodeDeleteAPIError`. Added a regression assertion in `TestPollWikiDeleteSpaceTaskWrapsPollFailuresWithHint` that the upstream Lark code (131006) survives a fully failed poll.

shortcuts/wiki/wiki_async_task.go
```diff
@@ -0,0 +1,176 @@
...
+
+// StatusCode returns a never-empty status value for the output envelope. If
+// the backend response omits delete_*_result.status (or sends whitespace),
+// fall back to "processing" so the documented timeout-shape stays accurate.
+func (s wikiAsyncTaskStatus) StatusCode() string {
+	if status := strings.TrimSpace(s.Status); status != "" {
+		return status
+	}
+	return wikiAsyncStatusProcessing
+}
```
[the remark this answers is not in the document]
liujinkun2025: Won't-fix (intentional). Coercing unknown non-terminal statuses to `processing` would break the existing `TestWikiDeleteSpaceTaskStatusClassification`, which deliberately asserts an unknown status like `running` is surfaced verbatim via `StatusLabel()`. The design intent is: unknown status → not classified as failure (`failed=false`, correct), but show the backend's raw string rather than masking it as a generic "processing". The boolean classification is already consistent; only the human label is verbatim, which is desired. A safer reshape is out of scope here.

liujinkun2025:
### Review follow-up (force-pushed amend `2cfd336`)

**CodeRabbit actionables**
- 🟡 `wiki_async_task.go` `ErrWithHint` drops structured fields → **fixed** (see in-thread reply; hand-built ExitError + regression test).
- 🟡 `wiki_async_task.go` `StatusCode()`/`Failed()` mismatch → **won't-fix** (see in-thread reply; coercing breaks the intentional `TestWikiDeleteSpaceTaskStatusClassification` contract).

**CodeRabbit nitpicks**
- `TestWikiShortcutsIncludeAllCommands` → **removed entirely**. It's a pre-existing (#610) circular change-detector: the count/index assertion is updated in lockstep with the code it "guards", so it can't catch the bug it claims to. `TestWikiShortcutsIncludesSpaceListNodeListNodeCopy` (membership-based) remains.
- `wikiSpaceCreateOutput` reuse `parseWikiSpaceRecord` → **won't-fix**. There is no shared *formatter* — `parseWikiSpaceRecord` returns a struct without a `Description` field. Composing it would add an error path and a struct→map rebuild to reproduce the same flat output; the direct raw-map read is simpler and already commented.

**Two additional bugs found during review (also fixed in this amend)**
1. **`+node-delete` async polling read the wrong field.** The gateway stores the delete-node task status under the generic `simple_task_result` key, **not** `delete_node_result`. The old constant meant every async node-delete would poll forever and falsely report `timed_out` even on success. Fixed the constant + the masking test stub.
2. **`+node-delete` timeout `next_command` was a dead end.** It points users at `drive +task_result --scenario wiki_delete_node`, but that scenario didn't exist (`drive +task_result` rejected it as unsupported). Added the `wiki_delete_node` scenario (Validate/DryRun/Execute/scope + `queryWikiDeleteNodeTask` reading `simple_task_result`) with tests.

Also added the missing skill reference docs (`+node-get` / `+node-delete` / `+space-create`) and SKILL.md shortcuts/permission table entries.

skills/lark-wiki/SKILL.md
```diff
@@ -58,8 +58,11 @@ Shortcut 是对常用操作的高级封装（`lark-cli wiki +<verb> [flags]`）
 | [`+node-create`](references/lark-wiki-node-create.md) | Create a wiki node with automatic space resolution |
 | [`+delete-space`](references/lark-wiki-delete-space.md) | Delete a wiki space, polling the async delete task when needed |
 | [`+space-list`](references/lark-wiki-space-list.md) | List all wiki spaces accessible to the caller |
+| [`+space-create`](references/lark-wiki-space-create.md) | Create a wiki space (user identity only) |
 | [`+node-list`](references/lark-wiki-node-list.md) | List wiki nodes in a space or under a parent node (supports pagination) |
 | [`+node-copy`](references/lark-wiki-node-copy.md) | Copy a wiki node to a target space or parent node |
+| [`+node-get`](references/lark-wiki-node-get.md) | Get a wiki node's details by node_token / obj_token / Lark URL |
+| [`+node-delete`](references/lark-wiki-node-delete.md) | Delete a wiki node, polling the async delete task when needed |
```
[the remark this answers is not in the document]
liujinkun2025: Won't-fix — this is intentional, not an omission. There is **no registered `nodes.delete` raw meta API**: `internal/registry/meta_data.json` only exposes `nodes.copy` / `nodes.create` / `nodes.list` (+ `spaces.create`). `+node-delete` is a **shortcut-only** capability wrapping a platform OpenAPI that the CLI registry does not surface.

- The `## API Resources` / 权限表 sections document the raw CLI API surface (`lark-cli wiki <resource> <method>` via `lark-cli schema`). Adding `nodes.delete` there would falsely imply `lark-cli wiki nodes delete` works, which it does not.
- The shortcut's required scope **is** documented, in its reference page `references/lark-wiki-node-delete.md` (`Required Scope: wiki:node:create`) — the correct home for a shortcut-only capability.
- Precedent: `nodes.move` is likewise shortcut-only (no raw API) and is deliberately not listed in API Resources for the same reason.

The verification script here only checked whether "delete" is mentioned; it didn't distinguish a registered raw API from a shortcut-only wrapper.

liujinkun2025:
Proactive convergence (not a CodeRabbit ask): dropped the synthesized `url` field from `+node-get` output. `wiki/v2/spaces/get_node` returns no URL, and the `BuildResourceURL` fallback (`www.feishu.cn/wiki/<node_token>`) is a non-canonical link that is misleading in a read/confirm command. Sibling read shortcuts (`+node-list`, `+node-copy`) likewise emit no url; `node_token`/`obj_token` remain as the precise identifiers. Updated the shortcut output, pretty renderer, test assertion, and reference doc accordingly. `+node-create`'s url (a legitimate created-resource fallback) is unchanged.

shortcuts/wiki/wiki_async_task.go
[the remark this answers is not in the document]
liujinkun2025: won't resolve

fangshuyu-768 (commented):
Thanks for the PR — solid work overall. Left a few comments on things worth considering before merge. Summary of findings:

1. **[Medium] `wikiNodeGetURLObjTypes` map iteration** — Go map order is non-deterministic; consider an ordered slice or a runtime prefix-conflict check.
2. **[Low] `queryWikiDeleteNodeTask` vs `parseWikiAsyncTaskStatus`** — duplicated logic; add a sync-maintenance note.
3. **[Needs confirmation] `+space-create` Risk level** — `write` vs `high-risk-write` for an irreversible operation.
4. **[Low] `looksLikeWikiNodeToken` heuristic** — only checks `wik` prefix; worth documenting the assumption.
5. **[Nitpick] `wikiSpaceCreateOutput` not reusing `parseWikiSpaceRecord`** — risk of field drift.
6. **[Low] `formatWikiTimestamp` uses local timezone** — output varies by server timezone.

shortcuts/wiki/wiki_node_get.go
```diff
@@ -0,0 +1,364 @@
...
+// /wiki/ is handled separately because node_tokens take no obj_type.
+//
+// INVARIANT: the prefixes must be mutually exclusive (no prefix may be a
+// prefix of another). tokenAndObjTypeFromWikiURL ranges this map, and Go map
+// iteration order is randomized — overlapping prefixes would make the match
+// non-deterministic. The trailing slash keeps them disjoint today (e.g.
+// "/docx/" does not start with "/doc/"); preserve that when adding entries.
+var wikiNodeGetURLObjTypes = map[string]string{
+	"/docx/":     "docx",
+	"/doc/":      "doc",
```
fangshuyu-768, line 30: The comment correctly identifies the invariant that prefixes must be mutually exclusive, but Go map iteration order is non-deterministic. If someone later adds an overlapping prefix (e.g. `/doc-new/`), the match in `tokenAndObjTypeFromWikiURL` becomes non-deterministic with no runtime error.

Consider either:
1. Converting this to an ordered slice of `{prefix, objType}` pairs (similar to `wikiNodeDeleteObjTypes`), which makes iteration deterministic and the match order explicit; or
2. Adding an `init()` or test that validates the mutual-exclusion invariant at startup.

The current prefixes are indeed disjoint, so this is not a bug today — just a maintenance hazard.

shortcuts/drive/drive_task_result.go
```diff
@@ -540,3 +548,60 @@ func queryWikiDeleteSpaceTask(runtime *common.RuntimeContext, taskID string) (ma
 		"status_msg": label,
 	}, nil
 }
+
+// queryWikiDeleteNodeTask returns the normalized status of an async wiki
+// delete-node task. For historical reasons the gateway stashes delete-node
```
fangshuyu-768, line 553: The comment explains the intentional duplication (avoiding a drive→wiki import), which is a reasonable architectural choice. However, the success/failed/processing classification logic here mirrors `wiki.parseWikiAsyncTaskStatus` quite closely, and they need to stay in sync if the status classification ever changes.

Consider adding a `// Keep in sync with wiki.parseWikiAsyncTaskStatus` annotation so future maintainers know to update both sites together.

shortcuts/wiki/wiki_space_create.go
```diff
@@ -0,0 +1,120 @@
...
+// WikiSpaceCreate wraps wiki.spaces.create. The raw API only takes two
+// optional string fields, so the shortcut's value is flag ergonomics
+// (no hand-written --params JSON), output flattening (data.space.* lifted
+// to the top level), and a dry-run preview.
+//
+// The API only accepts a user access token (no tenant/bot), so AuthTypes is
+// user-only — the framework's CheckIdentity rejects --as bot for us.
+var WikiSpaceCreate = common.Shortcut{
+	Service:     "wiki",
+	Command:     "+space-create",
```
fangshuyu-768, line 24: The PR description explains that `danger:true` in the schema browser is not a runtime confirmation gate, and that `Risk:write` is runtime-consistent with the raw API behavior. However, creating a space is not easily reversible (it requires a separate `+delete-space` call), and `+node-delete` uses `high-risk-write` with `--yes` gating.

Should `+space-create` also be `high-risk-write` with a `--yes` confirmation? If the intent is to keep it at `write` because the raw API has no confirmation, that's fine — but worth confirming this is deliberate rather than an oversight.

shortcuts/wiki/wiki_node_get.go
```diff
@@ -0,0 +1,364 @@
...
+			}
+		} else {
+			spec.SourceKind = "raw-obj"
+			// A raw obj_token needs an explicit obj_type: get_node would
+			// otherwise default to "doc" and fail confusingly for docx /
+			// sheet / bitable / ... Fail fast with the same upfront contract
+			// as +node-delete instead of deferring to an opaque API error.
+			if spec.ObjType == "" {
+				return wikiNodeGetSpec{}, output.ErrValidation(
+					"--obj-type is required for a raw obj_token %q (one of: %s); or pass a typed Lark URL (e.g. /docx/<token>) so it can be inferred",
```
fangshuyu-768, line 219: This heuristic only checks for a `wik` prefix. If Lark ever introduces other token types starting with `wik` that are *not* node_tokens, this check would misclassify them. The current behavior is reasonable for today's token scheme.

Consider adding a brief comment noting this is a heuristic based on the current Lark token naming convention, so it's clear this isn't a guaranteed invariant.

shortcuts/wiki/wiki_space_create.go
```diff
@@ -0,0 +1,120 @@
...
+		Name:        strings.TrimSpace(runtime.Str("name")),
+		Description: strings.TrimSpace(runtime.Str("description")),
+	}
+	if spec.Name == "" {
+		return wikiSpaceCreateSpec{}, output.ErrValidation("--name is required and cannot be blank")
+	}
+	return spec, nil
+}
+
+// wikiSpaceCreateOutput flattens data.space into the top-level envelope. It
```
fangshuyu-768, line 108: This function manually extracts the same fields that `parseWikiSpaceRecord` already handles. The comment explains the rationale (round-tripping `description`), but if `parseWikiSpaceRecord`'s field set changes, this output will silently drift.

Consider composing with `parseWikiSpaceRecord(raw)` and overlaying `description`:
```go
func wikiSpaceCreateOutput(raw map[string]interface{}) map[string]interface{} {
    out := parseWikiSpaceRecord(raw)
    out["description"] = common.GetString(raw, "description")
    return out
}
```
This keeps `+space-create` output consistent with other wiki space outputs and avoids future drift.

shortcuts/wiki/wiki_node_get.go
```diff
@@ -0,0 +1,364 @@
...
+
+	objEditRaw := common.GetString(raw, "obj_edit_time")
+	out["obj_edit_time"] = objEditRaw
+	out["obj_create_time"] = common.GetString(raw, "obj_create_time")
+	out["node_create_time"] = common.GetString(raw, "node_create_time")
+	out["updated_at"] = formatWikiTimestamp(objEditRaw)
+
+	return out
+}
+
```
fangshuyu-768, line 333: `time.Unix(secs, 0).Format(time.RFC3339)` uses the server's local timezone, so the output varies depending on where the CLI runs. For a CLI tool that may be used across timezones, this can be confusing.

Consider using UTC for consistency:
```go
return time.Unix(secs, 0).UTC().Format(time.RFC3339)
```
Or at minimum, document that the timestamp is in local time. RFC3339 does include the timezone offset, so the output is technically unambiguous — but the varying offset across runs may still surprise users.

liujinkun2025: Fixed in `7d8071f` — `formatWikiTimestamp` now uses `.UTC()`, so output is a stable `2023-11-14T22:13:20Z`-style string regardless of host timezone. Tightened the two assertions (`TestFormatWikiTimestamp` + the mounted-execute test) to the exact UTC value so a regression to local time would fail.

shortcuts/drive/drive_task_result.go
liujinkun2025: Addressed in `7d8071f` — added a `// Keep in sync with wiki.parseWikiAsyncTaskStatus ...` annotation on the classification block so future maintainers update both sites together. Kept the duplication (drive→wiki import avoidance) as you noted.

shortcuts/wiki/wiki_node_get.go
liujinkun2025: Addressed in `7d8071f` — `looksLikeWikiNodeToken` now documents that it's a heuristic on the current Lark token-naming convention (not a guaranteed invariant), and that misclassification's worst case is a confusing API error with no data risk.

shortcuts/wiki/wiki_node_get.go
liujinkun2025: Won't-fix (deliberate). The prefixes are disjoint today and the INVARIANT comment documents the requirement. Converting to an ordered slice or adding an init/test guard is more machinery than a 7-entry static, mutually-exclusive table warrants; the comment is the right-sized mitigation for a non-bug maintenance note. Will revisit if an overlapping prefix is ever genuinely needed.

shortcuts/wiki/wiki_space_create.go
liujinkun2025: Won't-fix. The suggested snippet does not compile: `parseWikiSpaceRecord` returns `(*wikiSpaceRecord, error)` (a struct, with an error), not a `map`, and the struct has no `Description` field. Composing it would require a struct→map mapper that doesn't exist plus an added error path, to reproduce the same flat output. The direct raw-map read is simpler and the rationale (round-tripping `description`) is already commented. A true single-source fix would mean adding `Description` to the shared `wikiSpaceRecord` + a shared formatter — out of scope for this PR.

shortcuts/wiki/wiki_space_create.go
liujinkun2025: Confirmed deliberate, keeping `Risk:write`. Rationale: (1) runtime-consistent with the raw `spaces.create` API, which has no `risk:high-risk-write` and thus no `--yes` gate; (2) space creation is additive and reversible (`+delete-space`), unlike `+node-delete` which is destructive/irreversible — that's why the latter is `high-risk-write`; (3) consistent with the sibling `+node-create` (`write`). Gating a non-destructive create would actually create an inconsistency with both the raw API and `+node-create`.

fangshuyu-768 (approved):

PR #934, opened 2026-05-18
feat(sidecar): support multi-client identity isolation in server-demo

hGrany:
## Summary

When multiple CLI sandbox environments share a single sidecar instance, user tokens (UAT) are not isolated — the last user to log in overwrites previous users' tokens, causing **identity cross-contamination** (user A operates Feishu as user B).

This PR adds per-client HMAC key isolation to `server-demo`, enabling each client environment to maintain its own Feishu identity.

## Changes

| File | Type | Description |
|------|------|-------------|
| `auth_bridge.go` | New | Management-plane endpoints (`login`/`poll`/`status`) with explicit `clientName → feishuOpenId` binding and persistent mapping |
| `handler.go` | Modified | Multi-key HMAC verification: tries the shared key first, then iterates per-client keys to identify the request origin |
| `main.go` | Modified | `--keys-dir` flag for client key directory; reuses existing `proxy.key` on restart instead of regenerating (fixes a race condition with multiple instances) |

## Design Decisions

1. **Why use HMAC key for client identification instead of a header field?**
   - The HMAC key is the existing trust anchor. Using it for identification introduces no new trust assumptions and prevents a malicious client from spoofing another client's identity.

2. **Why separate management-plane and data-plane keys?**
   - Management plane (login flow) needs a unified entry point — all clients use the shared `proxy.key`.
   - Data plane (API proxy) needs per-client isolation to inject the correct user token.

3. **Why no fallback when a client has no user mapping?**
   - This is authentication — silently falling back to another user's token is a security violation. Unmapped clients receive an explicit error prompting them to log in.

4. **Why reuse `proxy.key` instead of regenerating on each start?**
   - When multiple sidecar instances start concurrently (e.g. one per Lark app), they all write to the same `proxy.key`. The last writer wins, leaving other instances with an in-memory key that doesn't match the file. Reusing the existing key eliminates this race.

## Backward Compatibility

- Wire protocol (`sidecar` package) is **unchanged** — no client-side (`lark-cli-sidecar`) changes needed.
- Existing single-client deployments continue to work: the shared key is tried first, and if it matches, the request proceeds as before.
- TAT (bot token) resolution is unaffected — it always uses the shared credential provider.

## Risk Assessment

| Risk | Level | Mitigation |
|------|-------|------------|
| Key iteration performance | Low | Typical deployment < 50 clients; HMAC verification < 1μs each |
| Hot-reload of new keys | Low | Automatic rescan on cache miss; no restart needed |
| Mapping file loss | Low | Users simply re-run the login flow to rebuild binding |
| No-fallback breaking change | Intentional | Unmapped clients must authenticate — this is by-design security behavior |

Signed-off-by: Gao Yang <grany@yeah.net> (topwin.tech)

## Summary by CodeRabbit

* **New Features**
  * Multi-tenant auth sidecar server with per-client key isolation, HMAC-protected management bridge, token injection, and graceful CLI startup.
  * OAuth device-flow login with persisted client→user mapping and token management; secure proxy forwarding with audit logging.

* **Documentation**
  * Comprehensive README for the multi-tenant demo and added “See also” linking to it.

* **Tests**
  * Extensive test suite covering proxy signing/verification, allowlists, sanitization, redirects, and multi-tenant workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/934?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/sidecar/server-demo/README.md
+++ b/sidecar/server-demo/README.md
@@ -195,3 +195,9 @@ its identity flipped (bot↔user) or its auth-header redirected (e.g. into
 | `allowlist.go` | Target host / identity allowlists |
 | `audit.go` | Log path/error sanitization |
 | `handler_test.go` | Unit tests for all of the above |
+
+## See also
+
+- [server-multi-tenant-demo](../server-multi-tenant-demo/) — extends this demo
+  with per-client HMAC key isolation, OAuth device-flow login, and persistent
+  client → user mapping for multi-tenant deployments
--- a/sidecar/server-demo/auth_bridge.go
+++ b/sidecar/server-demo/auth_bridge.go
@@ -26,6 +26,7 @@ import (
 	larkauth "github.com/larksuite/cli/internal/auth"
 	"github.com/larksuite/cli/internal/core"
 	"github.com/larksuite/cli/internal/credential"
+	"github.com/larksuite/cli/internal/vfs"
 )
 
 // authBridge handles /_sidecar/auth/* management endpoints.
@@ -77,7 +78,7 @@ func (ab *authBridge) loadUserMap() {
 	if ab.mapFile == "" {
 		return
 	}
-	data, err := os.ReadFile(ab.mapFile)
+	data, err := vfs.ReadFile(ab.mapFile)
 	if err != nil {
 		return
 	}
@@ -91,8 +92,14 @@ func (ab *authBridge) saveUserMap() {
 	if ab.mapFile == "" {
 		return
 	}
-	data, _ := json.MarshalIndent(ab.userMap, "", "  ")
-	_ = os.WriteFile(ab.mapFile, data, 0600)
+	data, err := json.MarshalIndent(ab.userMap, "", "  ")
+	if err != nil {
+		ab.logger.Printf("AUTH_BRIDGE_ERROR action=save_user_map error=%q", err.Error())
+		return
+	}
+	if err := vfs.WriteFile(ab.mapFile, data, 0600); err != nil {
+		ab.logger.Printf("AUTH_BRIDGE_ERROR action=save_user_map error=%q", err.Error())
+	}
 }
 
 // verifyManagementHMAC checks a simplified HMAC for management endpoints.
@@ -138,7 +145,7 @@ func sha256Hex(data []byte) string {
 
 // ServeHTTP routes management API requests.
 func (ab *authBridge) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	body, err := io.ReadAll(r.Body)
+	body, err := io.ReadAll(io.LimitReader(r.Body, 64*1024))
 	if err != nil {
 		jsonError(w, http.StatusBadRequest, "failed to read body")
 		return
@@ -496,7 +503,7 @@ func loadCachedScopes() string {
 		return ""
 	}
 	dir := filepath.Join(configDir, "cache", "auth_login_scopes")
-	entries, err := os.ReadDir(dir)
+	entries, err := vfs.ReadDir(dir)
 	if err != nil {
 		return ""
 	}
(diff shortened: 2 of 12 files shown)


sang-neo03:
hi @hGrany ：
  Two things after reading the diff end-to-end:
  
  **1. Tests for the multi-key path**

  The new HMAC multi-key logic is correct (`handler.go` L611-619 explicitly skips shared-key collisions and duplicate keyHex with `KEYS_SCAN_SKIP` logs), but 530 LOC landed without
  any test coverage. Three small unit tests would lock down the isolation guarantee that's the whole point of this PR:

  - `loadClientKeys`: shared-key collision and duplicate keyHex are both skipped + the `KEYS_SCAN_SKIP` log fires
  - `verifyWithClientKeys`: only the matching client key resolves to its `clientName` (and a stranger key → empty + error)
  - `loadUserMap` / `saveUserMap`: round-trip persistence across restart

  Without these the "isolation by HMAC" contract lives only in your head and the PR description.

  **2. Placement: extract to a dedicated example**

  Right now `server-demo` reads as "the minimum sidecar you can compile and run" — useful as a starting template. Adding 530 lines of OAuth device-flow bridge + persistent `client →
  openId` mapping pushes it toward "production-ready multi-tenant gateway", which is a legitimate target but a different one. Mixing the two collapses the value of having a minimal
  reference.
  
  Suggest extracting the new code into `examples/multi-tenant-sidecar/` (or `sidecar/server-multi-tenant-demo/`) and leaving `server-demo` as the single-tenant starting template,
  with cross-links in both READMEs. That way readers picking up the simple case still get a simple example, and operators who actually need multi-tenant isolation get a full
  reference without having to mentally subtract features.

hGrany:
@coderabbitai review

sang-neo03:
Thanks @hGrany — extraction to `server-multi-tenant-demo/` and the three core tests (`SkipsSharedKeyCollision`, `SkipsDuplicateKeyHex`, `VerifyWithClientKeys_MatchesCorrectClient`,
  `UserMap_RoundTripPersistence`) all landed. Good follow-through, and the extra 20+ tests around HMAC / SSRF / redirect / auth-header allowlist make the contract much more visible. Happy
  with this part.

  Before I can sign off though, the CodeRabbit findings need attention — I'd treat the following two as merge blockers:

  **1. DoS surface on the management plane** — `/login`, `/poll`, `/status` all use `io.ReadAll(r.Body)` without a cap. Please wrap with `http.MaxBytesReader` (these endpoints only take
  small JSON, a few KB is plenty).

  **2. Path traversal via `--keys-dir` / `--key-file`** — both are user-supplied paths but bypass `validate.SafeInputPath`. In a sidecar this is the trust boundary; the repo's path-guard
  contract should apply here too. Same goes for swapping the direct `os.ReadFile` / `os.ReadDir` / `os.WriteFile` calls to `vfs.*` — it's both the repo convention and what makes these
  paths mockable.
  
  The remaining items are smaller but worth folding into the same pass:
  - `saveUserMap` silently discarding write errors (data loss masking)
  - Allowlist acceptance test hitting a real external host with no timeout (CI flakiness)
  - `TestRun_RejectsSelfProxy` missing `t.Setenv("LARKSUITE_CLI_CONFIG_DIR", t.TempDir())` for config isolation
  - Ignored `os.MkdirAll` error in test fixtures
  - README code fence languages (MD040)

  Once those are pushed I'll do another pass. Overall direction (HMAC-key-as-identity, no-fallback for unmapped clients, `proxy.key` reuse to fix the startup race) reads sound — just want
  the boundary hygiene to match.

hGrany:
Thanks @sang-neo03 — all seven items were already addressed in the last two commits (`c6959d1` + `1abdeb4`) that landed before your comment:

- Management body reads are capped at 64 KB via `io.LimitReader`
- All three CLI path flags go through `validate.SafeInputPath`; `os.*` swapped to `vfs.*` throughout
- `saveUserMap` errors are logged
- Allowlist test uses `httptest.NewTLSServer` (fully offline)
- `TestRun_RejectsSelfProxy` uses `t.Setenv` + temp config dir
- `os.MkdirAll` error checked in fixture
- README fences tagged with `text`

Waiting for your next look.

sang-neo03 (approved):

PR #935, opened 2026-05-18
fix(sheets): add missing JSON unmarshal error handling in dry-run mode

KhanCold:
## Summary

Two DryRun functions in the sheets shortcuts called `json.Unmarshal` without checking the return value. This looks like a bug, but `Validate` already parses and validates the same `--style` / `--data` JSON before `DryRun` runs, so the error is structurally impossible at this point.

Use `_ =` assignment + comment to silence the unchecked-error lint warning and make the safety invariant explicit to future readers.

## Changes

- `SheetSetStyle.DryRun`: document why `json.Unmarshal` of `--style` can be safely ignored
- `SheetBatchSetStyle.DryRun`: document why `json.Unmarshal` of `--data` can be safely ignored (via `validateBatchStyleData()`)

## Note

The initial attempt added `if err != nil { return common.FlagErrorf(...) }` inside `DryRun`, but `DryRun` must return `*common.DryRunAPI` (not `error`), so that would not compile. The correct fix is to acknowledge the already-validated invariant instead.

## Summary by CodeRabbit

* **Style**
  * Internal code cleanup to improve maintainability in sheet functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/935?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/sheets/lark_sheets_cell_style_and_merge.go
+++ b/shortcuts/sheets/lark_sheets_cell_style_and_merge.go
@@ -95,7 +95,7 @@ var SheetSetStyle = common.Shortcut{
 		}
 		r := normalizePointRange(runtime.Str("sheet-id"), runtime.Str("range"))
 		var style interface{}
-		json.Unmarshal([]byte(runtime.Str("style")), &style)
+		_ = json.Unmarshal([]byte(runtime.Str("style")), &style) // Validate already parses and validates this JSON.
 		return common.NewDryRunAPI().
 			PUT("/open-apis/sheets/v2/spreadsheets/:token/style").
 			Body(map[string]interface{}{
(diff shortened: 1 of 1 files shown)


fangshuyu-768:
This PR does not compile. The `DryRun` function signature returns `*common.DryRunAPI`, but `common.FlagErrorf` returns `error`. The type mismatch causes build failures:

```
shortcuts/sheets/lark_sheets_cell_style_and_merge.go:99:11: cannot use common.FlagErrorf("--style must be valid JSON: %v", err) (value of type error) as *cmdutil.DryRunAPI value in return statement
shortcuts/sheets/lark_sheets_cell_style_and_merge.go:170:11: cannot use common.FlagErrorf("--data must be valid JSON: %v", err) (value of type error) as *cmdutil.DryRunAPI value in return statement
```

Since `Validate` already guarantees valid JSON before `DryRun` is called, a possible fix is to explicitly discard the error with a comment:

```go
// Safe to ignore: Validate already verified JSON validity
_ = json.Unmarshal([]byte(runtime.Str("style")), &style)
```

KhanCold:
@fangshuyu-768 Thanks for the detailed feedback. You were absolutely right — the original commit still had the `if err != nil { return common.FlagErrorf(...) }` paths that don't compile because DryRun returns `*common.DryRunAPI`, not `error`.

I've amended the commit to use `_ = json.Unmarshal(...)` with explicit comments documenting the safety invariant (Validate already guarantees valid JSON before DryRun runs). Force-pushed as bb9bd12. PTAL, thanks!

fangshuyu-768 (approved):

PR #936, opened 2026-05-18
fix: mark base field update high risk

zgz2048:
## Summary
- Mark `base +field-update` as `high-risk-write`.
- Add a metadata test locking the risk level.
- Update the field update execution test to pass `--yes` under the high-risk confirmation gate.
- Update lark-base skill references so agents include `--yes` for field updates.

## Why
`field update` is implemented with PUT semantics and sends a complete field definition, not a narrow patch. Changing a Base field can change the column type or configuration, which may affect how the entire column of existing data is interpreted, displayed, or remains usable. It should require explicit confirmation like other high-risk Base schema changes.

The skill docs need to match this behavior; otherwise agents may keep generating `+field-update` commands without `--yes` and hit the confirmation gate.

## Validation
- `go test ./shortcuts/base`

## Summary by CodeRabbit

* **Bug Fixes**
  * Field update operations now report the correct high-risk classification.

* **Documentation**
  * Clarified safety guidance: field-update is a high-risk operation and must be confirmed with --yes; examples and guidance updated accordingly.

* **Tests**
  * Added and adjusted tests to verify the risk classification and confirmation behavior for field updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/936?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/base/base_execute_test.go
+++ b/shortcuts/base/base_execute_test.go
@@ -432,7 +432,7 @@ func TestBaseFieldExecuteUpdate(t *testing.T) {
 			"data": map[string]interface{}{"id": "fld_x", "name": "Amount", "type": "number"},
 		},
 	})
-	if err := runShortcut(t, BaseFieldUpdate, []string{"+field-update", "--base-token", "app_x", "--table-id", "tbl_x", "--field-id", "fld_x", "--json", `{"name":"Amount","type":"number"}`}, factory, stdout); err != nil {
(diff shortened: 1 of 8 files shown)


kongenpei (approved):
LGTM. Verified locally with:

- `go test ./shortcuts/base`
- `go test ./shortcuts/common ./shortcuts/base`

Non-blocking: `shortcuts/base/field_update.go` tips still show `+field-update` examples without `--yes`, while the command is now `high-risk-write`. Consider updating those examples so copied help output succeeds.

PR #953, opened 2026-05-18
fix(drive): clarify wiki-owned delete failures

afengzi:
## Summary

- reject wiki node tokens early in `drive +delete` validation with a more specific explanation
- add guidance to `1061004 forbidden` delete failures for wiki-owned nodes, including the expected alternatives
- cover the validation and enriched error message paths with tests

## Validation

- `go test -count=1 ./shortcuts/drive`
- `go test -race -gcflags="all=-N -l" -count=1 ./shortcuts/drive`
- `go vet ./shortcuts/drive`
- `make build`

Refs #566

## Summary by CodeRabbit

* **Bug Fixes**
  * `drive +delete` now displays clearer error messages when attempting to delete wiki-owned nodes, including guidance on alternative deletion methods.
  * Added input validation to prevent deletion attempts on wiki-owned nodes with an explanatory error message.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/drive/drive_delete.go
+++ b/shortcuts/drive/drive_delete.go
@@ -5,6 +5,7 @@ package drive
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -88,7 +89,7 @@ var DriveDelete = common.Shortcut{
 			nil,
 		)
 		if err != nil {
-			return err
+			return enrichDriveDeleteError(err)
 		}
 
 		if spec.FileType == "folder" {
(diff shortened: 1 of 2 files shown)


liujinkun2025:
Thanks for improving the wiki-owned delete UX — the enriched `1061004` hint is a nice touch.

One concern about `looksLikeWikiNodeToken`: the `wiki`/`wikcn` prefix check isn't reliable. Newer wiki node tokens no longer carry a `wiki`/`wikcn` prefix, so from a token alone the CLI can't reliably tell whether it points to a wiki node or a regular Drive/space document. In practice this guard will miss most real wiki node tokens, so I wouldn't rely on it for correctness — the runtime `1061004` handling you added already covers all token formats and is what users will actually hit.

Longer term, instead of rejecting wiki nodes up front, we're considering supporting deletion of wiki-type nodes directly in `drive +delete`. Happy to track that as a follow-up.

PR #960, opened 2026-05-19
fix(wiki): surface real node url for +node-create / +node-copy

liujinkun2025:
## Summary

The create-node and copy-node OpenAPI responses carry a real `url` field (present in practice, even though it is not in Feishu's documented node schema). Both shortcuts ignored it:

- `+node-create` synthesized a link via `BuildResourceURL` (host + `/wiki/` + token).
- `+node-copy` emitted no URL at all.

This change makes both shortcuts prefer the canonical URL returned by the API:

- Parse `url` into the shared `wikiNodeRecord`.
- Add a `wikiNodeURL(brand, node)` helper that returns the response `url` when present and falls back to `BuildResourceURL` synthesis only when it is blank/absent.
- Wire `augmentWikiNodeCreateOutput` (+node-create) and `+node-copy` output/pretty rendering to the helper.

`+node-get` / `+space-create` are intentionally untouched — they build their own output maps and deliberately omit URL, so the new struct field does not leak into them.

## Test plan

- [x] `go test ./shortcuts/wiki/...` passes
- [x] `gofmt` clean, `go vet ./shortcuts/...` clean
- [x] `./build.sh` succeeds
- [x] Updated existing `+node-create` e2e test: mock returns a real `url`; asserts it wins over synthesized fallback
- [x] Updated existing `+node-copy` e2e test: mock returns a real `url`; asserts copy surfaces it
- [x] Added `TestWikiNodeURL` covering prefer-response / fallback / blank / nil / no-token
- [ ] `golangci-lint` not run locally (not installed); relying on CI

## Summary by CodeRabbit

* **New Features**
  * Wiki node create and copy outputs now include a user-facing URL when available; if the API omits it, a synthesized URL is used as a fallback. The pretty output shows the URL only when present.

* **Tests**
  * Tests updated and added to verify URL selection, precedence (API URL preferred), fallback behavior, and output formatting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/960?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/wiki/wiki_node_copy.go
+++ b/shortcuts/wiki/wiki_node_copy.go
@@ -89,6 +89,9 @@ var WikiNodeCopy = common.Shortcut{
 		fmt.Fprintf(runtime.IO().ErrOut, "Copied to node %s in space %s\n",
 			common.MaskToken(node.NodeToken), common.MaskToken(node.SpaceID))
 		out := wikiNodeCopyOutput(node)
+		if u := wikiNodeURL(runtime.Config.Brand, node); u != "" {
+			out["url"] = u
+		}
 		runtime.OutFormat(out, nil, func(w io.Writer) {
 			renderWikiNodeCopyPretty(w, out)
 		})
@@ -105,6 +108,9 @@ func renderWikiNodeCopyPretty(w io.Writer, out map[string]interface{}) {
 	fmt.Fprintf(w, "  obj_token:         %s\n", valueOrDash(out["obj_token"]))
 	if parent, _ := out["parent_node_token"].(string); parent != "" {
 		fmt.Fprintf(w, "  parent_node_token: %s\n", parent)
+	}
+	if url, _ := out["url"].(string); url != "" {
+		fmt.Fprintf(w, "  url:               %s\n", url)
 	}
 }
 
--- a/shortcuts/wiki/wiki_node_create.go
+++ b/shortcuts/wiki/wiki_node_create.go
@@ -461,20 +461,6 @@ func parseWikiNodeRecord(node map[string]interface{}) (*wikiNodeRecord, error) {
 	}, nil
 }
 
-// wikiNodeURL returns the user-facing link for a wiki node. The create/copy
-// OpenAPI responses carry a real `url` (undocumented in the server-docs schema
-// but present in practice); prefer it so the CLI surfaces the canonical link.
-// Fall back to BuildResourceURL synthesis only when the response omits it.
-func wikiNodeURL(brand core.LarkBrand, node *wikiNodeRecord) string {
-	if node == nil {
-		return ""
-	}
-	if u := strings.TrimSpace(node.URL); u != "" {
-		return u
-	}
-	return common.BuildResourceURL(brand, "wiki", node.NodeToken)
-}
-
 func parseWikiSpaceRecord(space map[string]interface{}) (*wikiSpaceRecord, error) {
 	if space == nil {
 		return nil, output.Errorf(output.ExitAPI, "api_error", "wiki space response missing space")
--- a/shortcuts/wiki/wiki_node_create.go
+++ b/shortcuts/wiki/wiki_node_create.go
@@ -118,6 +118,7 @@ type wikiNodeRecord struct {
 	OriginNodeToken string
 	Title           string
 	HasChild        bool
+	URL             string
 }
 
 // wikiSpaceRecord contains the response fields used when resolving spaces.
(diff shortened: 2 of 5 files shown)


shortcuts/wiki/wiki_node_create.go
```diff
@@ -456,9 +457,24 @@ func parseWikiNodeRecord(node map[string]interface{}) (*wikiNodeRecord, error) {
 		OriginNodeToken: common.GetString(node, "origin_node_token"),
 		Title:           common.GetString(node, "title"),
 		HasChild:        common.GetBool(node, "has_child"),
+		URL:             common.GetString(node, "url"),
 	}, nil
 }
 
+// wikiNodeURL returns the user-facing link for a wiki node. The create/copy
+// OpenAPI responses carry a real `url` (undocumented in the server-docs schema
+// but present in practice); prefer it so the CLI surfaces the canonical link.
```
fangshuyu-768: Nit: `wikiNodeURL` is consumed by both `+node-create` and `+node-copy`, but lives in `wiki_node_create.go`. Consider moving it to a shared helper file (e.g. `wiki_helpers.go`) so the placement reflects its broader usage. Not blocking — just a code-organization thought.

shortcuts/wiki/wiki_node_copy.go
```diff
@@ -89,6 +89,9 @@ var WikiNodeCopy = common.Shortcut{
 		fmt.Fprintf(runtime.IO().ErrOut, "Copied to node %s in space %s\n",
 			common.MaskToken(node.NodeToken), common.MaskToken(node.SpaceID))
 		out := wikiNodeCopyOutput(node)
+		if u := wikiNodeURL(runtime.Config.Brand, node); u != "" {
```
fangshuyu-768, line 92: Question: `wikiNodeURL` accepts `*wikiNodeRecord`, but here `node` comes from the copy flow. Could you confirm that `node` is indeed a `*wikiNodeRecord`? If the copy operation uses a different struct type for `node`, this would be a type mismatch. (If it compiles, we're good — just want to double-check the intent.)

shortcuts/wiki/wiki_node_create_test.go
```diff
@@ -628,3 +629,47 @@ func TestAugmentWikiNodeCreateOutputReturnsEmptyMapForNilInput(t *testing.T) {
...
+		},
+		{
+			name: "falls back to synthesized url when response omits it",
+			node: &wikiNodeRecord{NodeToken: "wik_token"},
+			want: "https://www.feishu.cn/wiki/wik_token",
+		},
+		{
+			name: "blank response url is treated as absent",
+			node: &wikiNodeRecord{NodeToken: "wik_token", URL: "   "},
+			want: "https://www.feishu.cn/wiki/wik_token",
```
fangshuyu-768, line 654: Good test coverage — the five subtests for `TestWikiNodeURL` cover all the important edge cases (nil, missing, blank, valid URL, no-token-no-url). 👍

fangshuyu-768 (commented):
## Review Summary

This is a clean, well-scoped fix. The `wikiNodeURL` helper correctly prioritizes the real API response URL over the synthesized fallback, and the test coverage is solid.

A few suggestions below — none are blocking.

**Verdict: LGTM** (after confirming the type compatibility question on `wiki_node_copy.go`).

shortcuts/wiki/wiki_node_create.go
liujinkun2025: Done in 13ca095 — moved `wikiNodeURL` into a new shared `shortcuts/wiki/wiki_helpers.go`. Pure move, no behavior change; gofmt/go vet/`go test ./shortcuts/wiki/...` all green.

shortcuts/wiki/wiki_node_copy.go
liujinkun2025: Confirmed — it's `*wikiNodeRecord`. In the copy flow, `node` is the return value of `parseWikiNodeRecord(common.GetMap(data, "node"))` (`wiki_node_copy.go:84`). `parseWikiNodeRecord` is the single shared parser (declared in `wiki_node_create.go`, signature `func(map[string]interface{}) (*wikiNodeRecord, error)`) used by both `+node-create` and `+node-copy` — there's no separate struct for copy. It compiles, and the updated copy e2e test (`TestWikiNodeCopyCopiesNodeToTargetSpace`) exercises this path and asserts the surfaced `url`, so the type and intent are both verified.

fangshuyu-768 (approved):

Issue #914, opened 2026-05-15
Clarify doctor/auth status when bot identity works but user OAuth is missing


Mycheers:
## Summary

`lark-cli doctor` and `lark-cli auth status --verify` can give confusingly different signals about whether auth is usable.

In one environment, `doctor` suggested token/auth problems, while `auth status --verify` still showed a usable bot identity. The important distinction was:

- bot tenant identity was still usable
- user OAuth identity was missing/expired/needed refresh

For agent workflows, this ambiguity causes the agent to conclude "auth is broken" when only one identity path is broken.

## Environment

- `lark-cli` version observed: `1.0.30`
- OS: WSL Ubuntu
- Brand: Feishu
- Workflow: mixed `--as user` reads and `--as bot` sends

## Observed confusion

Typical commands used during diagnosis:

```bash
export LARK_CLI_NO_PROXY=1
lark-cli doctor
lark-cli config strict-mode
lark-cli config show
lark-cli auth status --verify
lark-cli auth scopes
```

The key problem is that a user can see a token warning in `doctor`, but still have bot identity available. Conversely, user scopes may appear in config/scope output while `auth status --verify` reports that the user is not currently logged in or needs re-authorization.

Common downstream symptoms:

```text
need_user_authorization
```

```text
tokenStatus=needs_refresh
```

```text
refresh failed
```

```text
strict mode is "bot", only bot-identity commands are available
```

## Proposal

Please make `doctor` separate bot and user identity readiness explicitly, for example:

```text
Bot identity: ready / not configured / token check failed
User identity: ready / missing login / refresh failed / missing required scopes
Strict mode: off|bot|user
Active profile: <path>
Active app ID: <app_id>
Suggested next action: <login/bind/strict-mode command>
```

For JSON output, something like this would be agent-friendly:

```json
{
  "profile": "...",
  "app_id": "cli_xxx",
  "strict_mode": "off",
  "bot": { "status": "ready" },
  "user": { "status": "needs_login", "reason": "need_user_authorization" },
  "next_actions": [
    "lark-cli auth login --scope \"...\""
  ]
}
```

## Why this matters

Many automations use a split identity model:

- `lark-cli --as user` for reading chat history, docs, sheets, and user-visible resources
- `lark-cli --as bot` for stable message delivery

If `doctor` collapses these into a single auth status, users and agents debug the wrong identity path.

## Related direction

[...]

PR #961, opened 2026-05-19
fix(identitydiag): harden verify path and tighten status semantics

liangshuo-1:
Follow-ups to #957.

## Summary

- **Timeouts on verify**: bound `fetchBotInfo` and the user `VerifyUserToken` call with a 10s timeout (mirrors the existing doctor endpoint probe). Previously a hanging server could wedge `auth status --verify` or `doctor` indefinitely since both call sites passed an unbounded `context.Background()` / `cmd.Context()`.
- **Consistent not-configured semantics**: when the user-identity path is blocked by missing app config, return `StatusNotConfigured` instead of `StatusMissing` — matching the bot side. `StatusMissing` is now reserved for "app configured but user not logged in".
- **Richer bot HTTP errors**: on `/bot/v3/info` 4xx, try to parse the `{code, msg}` envelope. Error goes from `HTTP 401` to `HTTP 401: [99991663] app ticket invalid`, which is what agent workflows from #914 actually need.
- **Code polish**:
  - `cmd/auth/status.go`: introduce `identity{User,Bot,None}` constants, kill the bare string literals
  - `addStatusNote` uses the now-exported `identitydiag.StatusMessage()` so the note reads `User identity is not configured; ...` instead of `User identity is not_configured; ...`
  - Collapse the three duplicated `id.Status = StatusVerifyFailed; id.Available = false; id.Verified = boolPtr(false); ...` blocks in the user verify path into a single local helper

## Tests

`go test ./internal/identitydiag/ ./cmd/auth/ ./cmd/doctor/`

New unit cases in `internal/identitydiag/diagnostics_test.go`:

- `VerifyBotIdentity_HTTPErrorSurfacesEnvelope` — HTTP 401 with `{code, msg}` body asserts both HTTP code and envelope code make it into the message
- `VerifyBotIdentity_BusinessErrorCode` — HTTP 200 with `code != 0` still reports `StatusVerifyFailed`
- `VerifyUserIdentity_ServerRejects` — server-rejected user token reports `verify_failed` + unavailable + verified=false
- `UserIdentityExpired` — expired refresh token → `StatusMissing` with re-login hint
- `BotIdentityStrictUserOnly` — `SupportedIdentities` user-only flag → bot reports `not_configured` even when app secret is present
- `UserIdentityMissingAppConfig` — empty `AppID` → `not_configured` (regression-locks the new semantics)
- `StatusMessage` — table-driven coverage of the now-exported helper

## Summary by CodeRabbit

* **New Features**
  * Added timeout protection for identity verification to avoid long hangs.

* **Bug Fixes**
  * Verification error messages now include HTTP status and envelope error details for clearer diagnostics.
  * Identity status wording is more consistent and user-friendly (e.g., "not configured"); identity labels and status formatting standardized.

* **Tests**
  * Expanded diagnostic tests to cover many identity verification outcomes and message mappings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/961?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/auth/status.go
+++ b/cmd/auth/status.go
@@ -69,14 +69,20 @@ func authStatusRun(opts *StatusOptions) error {
 	return nil
 }
 
+const (
+	identityUser = "user"
+	identityBot  = "bot"
+	identityNone = "none"
+)
+
 func effectiveIdentity(d identitydiag.Result) string {
 	switch {
 	case d.User.Available:
-		return "user"
+		return identityUser
 	case d.Bot.Available:
-		return "bot"
+		return identityBot
 	default:
-		return "none"
+		return identityNone
 	}
 }
 
@@ -105,14 +111,14 @@ func addLegacyUserFields(result map[string]interface{}, user identitydiag.Identi
 
 func addEffectiveVerification(result map[string]interface{}, d identitydiag.Result) {
 	switch result["identity"] {
-	case "user":
+	case identityUser:
 		if d.User.Verified != nil {
 			result["verified"] = *d.User.Verified
 			if !*d.User.Verified {
 				result["verifyError"] = d.User.Message
 			}
 		}
-	case "bot":
+	case identityBot:
 		if d.Bot.Verified != nil {
 			result["verified"] = *d.Bot.Verified
 			if !*d.Bot.Verified {
@@ -125,7 +131,7 @@ func addEffectiveVerification(result map[string]interface{}, d identitydiag.Resu
 func addStatusNote(result map[string]interface{}, d identitydiag.Result) {
 	switch {
 	case !d.User.Available && d.Bot.Available:
-		result["note"] = "User identity is " + d.User.Status + "; bot identity is ready for bot/tenant API calls. Run `lark-cli auth login` to enable user identity."
+		result["note"] = "User identity is " + identitydiag.StatusMessage(d.User.Status) + "; bot identity is ready for bot/tenant API calls. Run `lark-cli auth login` to enable user identity."
 	case d.User.Status == identitydiag.StatusNeedsRefresh:
 		result["note"] = "User identity needs refresh and will be refreshed automatically on the next user API call."
 	case !d.User.Available && !d.Bot.Available:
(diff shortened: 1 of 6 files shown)


MaxHuang22 (dismissed):

liangshuo-1:
Pushed `bd50862` — fixes a P0 bug discovered while testing this branch locally.

`/open-apis/bot/v3/info` returns `{code, msg, bot: {...}}` but the envelope decoder in #957 read from `data`. Every `auth status --verify` and `doctor` run reported `Bot identity: verify failed: open_id is empty` — for **every** Lark account, since the field never existed at the path the code looked at.

Pre-existing test mocks emitted `{"data": {...}}` matching the buggy decoder, so unit tests stayed green while production silently broke. The fix is a one-character schema correction on the tag plus aligning all four mock fixtures.

Verified locally with both a normal account and a bot-only profile — `bot_identity` now `pass`/`verified=true`. Without this, the doctor `ok=false` regression in bot-only setups (the exact case #914 was about) would have continued.

liangshuo-1:
Pushed `cbc8fb4` — extending scope to fix the same schema bug in `shortcuts/common/runner.go`.

While analyzing the impact of the previous commit, I found `RuntimeContext.fetchBotInfo` has the **exact same** wrong-tag bug (`json:"data"` reading from a `"bot"` payload). Its tests pass for the same reason — mocks emit `"data"` matching the buggy decoder.

This one is currently dormant (`grep "\.BotInfo()"` shows zero production callers; only the test file and `TestNewRuntimeContextWithBotInfo` helper use it), so it hasn't caused user-visible regressions. But fixing it alongside avoids handing the next person who tries to use bot identity from a shortcut a self-confirming green test suite with broken production behavior.

Minimum-diff approach again: only the JSON tag changes, Go field name `Data` is kept. Four mock fixtures in `runner_botinfo_test.go` aligned to the real API shape.

### Branch summary

| Commit | Scope |
|---|---|
| `7b6349d` | Original follow-ups (timeouts, status consistency, 4xx body parsing, DRY, constants, test coverage) |
| `bd50862` | Schema fix in `internal/identitydiag` (discovered via local doctor testing) |
| `cbc8fb4` | Same schema fix in `shortcuts/common` (discovered via impact analysis) |

End-to-end verified: `lark-cli doctor` on a normal account and on a bot-only profile both report `ok: true` with `bot_identity: pass` and `verified: true` — the actual outcome #957 set out to deliver for #914.

MaxHuang22 (approved):

PR #964, opened 2026-05-19
feat(docs): support auto-chunking for oversized plain markdown text

jklasdf651:
Description
This PR significantly enhances the robust processing of markdown chunking, specifically addressing edge-case bugs and introducing safety guarantees for oversized documents.

Key Changes
1. Table Detection Optimization: Switched to alignment-row analysis as the definitive marker, supporting both standard markdown tables with/without leading/trailing pipes.
2. HTML Scan Loop: Upgraded from a single-token check to a continuous loop scan, preventing potential boundaries-crossing omissions.
3. Idempotency Guarantee: Introduced short-document pass-through logic (≤10k chars) to eliminate any processing risks for standard payloads.
4. Error Message Optimization: Refined error text to provide user-friendly, actionable suggestions for manually splitting complex/oversized structural elements.

Test Plan
Comprehensive unit tests have been added to `markdown_chunk_test.go` covering:
- Idempotency: Short content with code blocks, tables, and blockquotes passes through unchanged.
- Rejection Boundaries: Document sizes (>10k chars) combined with complex structural elements are correctly intercepted.
- Edge-case Markers: Validated alignment rows alone, consecutive pipe variations, less-than comparison operators vs. HTML tags to avoid false positives/negatives.

## Release Notes

* **New Features**
  * Documents with large Markdown content are now automatically chunked during creation and updates
  * Safety validation detects and rejects Markdown containing unsafe structural patterns (code blocks, blockquotes, HTML tags)
  * Proper UTF-8 support ensures emoji and international characters are handled correctly during content chunking

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/964?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/doc/markdown_chunk.go
+++ b/shortcuts/doc/markdown_chunk.go
@@ -1,0 +1,189 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package doc
+
+import (
+	"errors"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/larksuite/cli/internal/output"
+)
+
+const SafeParagraphLimit = 10000
+
+var ErrUnsafeMarkdown = errors.New("oversized markdown contains complex structural elements (fenced code blocks, tables, blockquotes, or HTML) that cannot be safely split. Please manually split the content below 10,000 characters before uploading")
+
+func isTableAlignmentRow(line string) bool {
+	line = strings.TrimSpace(line)
+	if line == "" || !strings.Contains(line, "-") {
+		return false
+	}
+	for _, r := range line {
+		if r != '|' && r != '-' && r != ':' && r != ' ' && r != '\t' {
+			return false
+		}
+	}
+	return true
+}
+
+func containsUnsafeMarkdown(md string) bool {
+	lines := strings.Split(md, "\n")
+
+	for i, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" {
+			continue
+		}
+
+		if strings.HasPrefix(trimmed, "```") || strings.HasPrefix(trimmed, "~~~") {
+			return true
+		}
+
+		if strings.HasPrefix(trimmed, ">") {
+			return true
+		}
+
+		temp := trimmed
+		for {
+			idx := strings.IndexByte(temp, '<')
+			if idx == -1 || idx+1 >= len(temp) {
+				break
+			}
+			next := temp[idx+1]
+			if (next >= 'a' && next <= 'z') || (next >= 'A' && next <= 'Z') || next == '/' || next == '!' {
+				return true
+			}
+			temp = temp[idx+1:]
+		}
+
+		if isTableAlignmentRow(trimmed) {
+			return true
+		}
+		if strings.Contains(trimmed, "|") && i+1 < len(lines) {
+			if isTableAlignmentRow(strings.TrimSpace(lines[i+1])) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func splitPlainParagraphs(md string) []string {
+	var paragraphs []string
+	var current strings.Builder
+	hasContent := false
+
+	for _, line := range strings.Split(md, "\n") {
+		if strings.TrimSpace(line) == "" {
+			if hasContent {
+				paragraphs = append(paragraphs, current.String())
+				current.Reset()
+				hasContent = false
+			}
+			continue
+		}
+		if hasContent {
+			current.WriteByte('\n')
+		}
+		current.WriteString(line)
+		hasContent = true
+	}
+	if hasContent {
+		paragraphs = append(paragraphs, current.String())
+	}
+
+	return paragraphs
+}
+
+func splitOversizedParagraph(para string) []string {
+	if utf8.RuneCountInString(para) <= SafeParagraphLimit {
+		return []string{para}
+	}
+
+	var chunks []string
+	pos := 0
+
+	for pos < len(para) {
+		remaining := para[pos:]
+		if utf8.RuneCountInString(remaining) <= SafeParagraphLimit {
+			chunks = append(chunks, remaining)
+			break
+		}
+
+		limitByte := runeOffset(remaining, SafeParagraphLimit)
+		splitAt := findSplitPoint(remaining, limitByte)
+
+		chunks = append(chunks, remaining[:splitAt])
+		pos += splitAt
+	}
+
+	return chunks
+}
+
+func runeOffset(s string, n int) int {
+	offset := 0
+	for i := 0; i < n && offset < len(s); i++ {
+		_, size := utf8.DecodeRuneInString(s[offset:])
+		offset += size
+	}
+	return offset
+}
+
+func findSplitPoint(s string, limitByte int) int {
+	minByte := limitByte * 3 / 4
+
+	for i := limitByte - 1; i >= minByte; i-- {
+		if s[i] == '\n' {
+			return i + 1
+		}
+	}
+
+	for i := limitByte - 1; i >= minByte; i-- {
+		if s[i] == ' ' {
+			return i + 1
+		}
+	}
+
+	return limitByte
+}
+
+func chunkMarkdownForUpload(md string) (string, error) {
+	if md == "" {
+		return "", nil
+	}
+
+	if utf8.RuneCountInString(md) <= SafeParagraphLimit {
+		return md, nil
+	}
+
+	if containsUnsafeMarkdown(md) {
+		return "", ErrUnsafeMarkdown
+	}
+
+	paragraphs := splitPlainParagraphs(md)
+	var chunks []string
+
+	for _, para := range paragraphs {
+		chunks = append(chunks, splitOversizedParagraph(para)...)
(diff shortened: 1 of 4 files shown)


shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+	"unicode/utf8"
+
+	"github.com/larksuite/cli/internal/output"
+)
+
+const SafeParagraphLimit = 10000
+
+var ErrUnsafeMarkdown = errors.New("oversized markdown contains complex structural elements (fenced code blocks, tables, blockquotes, or HTML) that cannot be safely split. Please manually split the content below 10,000 characters before uploading")
+
+func isTableAlignmentRow(line string) bool {
```
fangshuyu-768, line 18: The name `SafeParagraphLimit` doesn't convey that the unit is **runes** (Unicode code points), not bytes or grapheme clusters. Consider renaming to `SafeParagraphRuneLimit` or adding a comment like `// SafeParagraphLimit is the maximum number of Unicode code points (runes) per chunk.` to make the unit explicit.

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+	line = strings.TrimSpace(line)
+	if line == "" || !strings.Contains(line, "-") {
```
fangshuyu-768, line 20: The error message says "below 10,000 **characters**", but `SafeParagraphLimit` is actually 10,000 **runes** (Unicode code points). For CJK text or emoji, a single grapheme cluster may consist of multiple runes, so "characters" can be ambiguous. Suggest changing to "10,000 Unicode code points (runes)" or "10,000 runes" for precision.

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+		return false
+	}
+	for _, r := range line {
+		if r != '|' && r != '-' && r != ':' && r != ' ' && r != '\t' {
+			return false
+		}
+	}
```
fangshuyu-768, line 27: **False positive on setext-style headings.**

`isTableAlignmentRow("--- ---")` returns `true` because the line only contains `-`, ` `, and no disallowed characters. However, in Markdown, a line of `---` under text is a **setext h2 heading**, not a table alignment row:

```markdown
Section Title
---
```

This means any oversized document using setext headings will be incorrectly flagged as containing a table and rejected with `ErrUnsafeMarkdown`.

Suggestion: require at least one `|` character in the line for it to qualify as a table alignment row, or check the preceding line for pipe characters. This would also fix the test case `{"no pipe at all", "--- ---", true}` which should likely be `false`.

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return limitByte
+}
+
+func chunkMarkdownForUpload(md string) (string, error) {
+	if md == "" {
+		return "", nil
+	}
```
fangshuyu-768, line 155: **`minByte` is not guaranteed to be on a rune boundary.**

`minByte = limitByte * 3 / 4` is a pure byte-level arithmetic calculation. If the string contains multi-byte UTF-8 characters (CJK, emoji, etc.), `minByte` may land in the middle of a rune.

Currently this is *safe in practice* because the loop only checks for ASCII bytes (`\n` and `' '`), so it will never slice at `minByte` directly. However, the invariant is implicit and fragile — if someone later adds a non-ASCII split character check, it would silently produce invalid UTF-8.

Suggestion: either add a comment explicitly documenting this invariant, or align `minByte` to the next rune boundary:

```go
minByte := limitByte * 3 / 4
// Align to rune boundary
for minByte < len(s) && !utf8.RuneStart(s[minByte]) {
    minByte++
}
```

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if utf8.RuneCountInString(md) <= SafeParagraphLimit {
+		return md, nil
+	}
+
+	if containsUnsafeMarkdown(md) {
+		return "", ErrUnsafeMarkdown
+	}
+
+	paragraphs := splitPlainParagraphs(md)
```
fangshuyu-768, line 165: When no newline or space is found in the search range, `findSplitPoint` falls back to `limitByte`. This is correct because `limitByte` comes from `runeOffset` and is guaranteed to be on a rune boundary.

However, the intermediate positions checked in the loops above (`i` from `limitByte-1` down to `minByte`) are **not** guaranteed to be on rune boundaries. When `findSplitPoint` returns `i+1` (after finding `\n` or ` `), that position *happens* to be safe because `\n` and ` ` are single-byte ASCII characters, so `i+1` is always a valid rune start.

This works today, but the correctness depends on an unstated assumption. Consider adding a comment or a debug assertion:

```go
// Safe because \n and ' ' are single-byte ASCII runes,
// so i+1 is always a valid rune boundary.
return i + 1
```

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+	var chunks []string
+
+	for _, para := range paragraphs {
+		chunks = append(chunks, splitOversizedParagraph(para)...)
+	}
+
+	return strings.Join(chunks, "\n\n"), nil
+}
+
+func applyChunkingToBody(body map[string]interface{}, contentKey, docFormat string) error {
```
fangshuyu-768, line 175: **Re-joining with `"\n\n"` changes the original whitespace structure.**

If the original markdown had 3+ blank lines between paragraphs, or if `splitOversizedParagraph` splits at a newline causing a trailing `\n` in a chunk, the re-joined output will have different whitespace than the input. This means `chunkMarkdownForUpload` is not truly idempotent — calling it twice on the same input could produce different results.

For rendering purposes this is usually fine, but it's worth documenting this behavior. Alternatively, you could track the original separator between paragraphs and restore it on re-join.

shortcuts/doc/markdown_chunk.go
```diff
@@ -0,0 +1,189 @@
...
+	if docFormat != "markdown" {
+		return nil
+	}
+	content, _ := body[contentKey].(string)
+	if content == "" {
+		return nil
+	}
+	chunked, err := chunkMarkdownForUpload(content)
+	if err != nil {
+		return output.Errorf(output.ExitAPI, "chunk_error", "%v", err)
```
fangshuyu-768, line 185: **Silent type assertion ignores non-string content.**

```go
content, _ := body[contentKey].(string)
```

If `body[contentKey]` exists but is not a `string` (e.g., `[]byte`, `json.RawMessage`), the type assertion fails silently and `content` becomes `""`. The function then returns `nil` with no error, effectively skipping chunking for that content.

This could lead to hard-to-debug issues where oversized content silently bypasses chunking. Consider either:
1. Returning an error when the key exists but isn't a string, or
2. Logging a warning, or
3. Adding a comment explaining why silent fallback is intentional.

shortcuts/doc/markdown_chunk_test.go
```diff
@@ -0,0 +1,508 @@
...
+			if utf8.RuneCountInString(chunk) > SafeParagraphLimit {
+				t.Errorf("chunk %d exceeds limit: %d runes", i, utf8.RuneCountInString(chunk))
+			}
+		}
+	})
+}
+
+func TestChunkMarkdownChineseUTF8(t *testing.T) {
+	t.Run("Chinese text is split correctly", func(t *testing.T) {
+		longChinese := strings.Repeat("你好世界", 3000)
```
fangshuyu-768, line 367: **Missing round-trip content integrity check.**

The `TestChunkMarkdownChineseUTF8` and `TestChunkMarkdownEmojiSafety` tests only verify that the output is valid UTF-8, but they don't verify that the **total content is preserved** after chunking and re-joining. A bug that drops characters would still pass these tests.

Consider adding an assertion that the concatenation of all chunks (minus the added `\n\n` separators) equals the original input:

```go
joined := strings.ReplaceAll(result, "\n\n", "")
original := strings.ReplaceAll(longChinese, "\n\n", "")
if joined != original {
    t.Error("content was lost during chunking")
}
```

Or more precisely, verify that the total rune count is preserved.

fangshuyu-768 (changes requested):
## Review Summary

Overall the design is solid — the pass-through for short content, rejection of oversized unsafe content, and UTF-8-aware splitting are well thought out. Tests are comprehensive. Below are issues I found, ordered by severity.

### Critical
1. `findSplitPoint` — `minByte` is not guaranteed to be on a rune boundary. While currently safe (only ASCII delimiters are searched), this is fragile and should be documented or fixed.
2. `isTableAlignmentRow` — false positive on setext-style headings (e.g. `---` on its own line is an h2, not a table).

### Medium
3. `ErrUnsafeMarkdown` says "10,000 characters" but the limit is 10,000 runes.
4. `SafeParagraphLimit` name doesn't convey the unit is runes.
5. `chunkMarkdownForUpload` re-joins with `"\n\n"`, altering original whitespace — not truly idempotent.
6. `applyChunkingToBody` silently ignores non-string content values.

### Minor
7. UTF-8 tests check validity but not round-trip content integrity.

PR #1018, opened 2026-05-21
fix(sheets): +batch-update sub-op CLI-side validation + cond-format / filter schema alignment

zhengzhijiej-tech:
## Summary

Two adjacent fixes around `+batch-update` and related write shortcuts, both surfaced by BOE smoke testing of multi-op batches.

- **8d0fefd `fix(sheets): push +batch-update sub-op validation down into xxxInput builders`** — sub-ops that omit `--sheet-id` (or any other required flag) used to slip past CLI validation and surface as opaque server errors like `sheet undefined not found` after a network round-trip. Every batchable `xxxInput` builder now returns `(map, error)` and owns its own sheet-selector + required-field checks via the new `requireSheetSelector` helper. Standalone `Validate` shrinks to `validateViaInput(xxxInput)`, so identical bad input now fires identical CLI errors from both standalone and batch paths (wrapped as `operations[N] (+shortcut): ...` in the batch case). `sheetMoveBatchInput`'s explicit-`source-index` constraint is preserved.

- **4e44e66 `fix(sheets): align +cond-format / +filter with server schema (#4 + #5)`**
  - `+cond-format`: `condFormatEnhance` was writing `properties.rule.type` (nested), but the server schema (`manage_conditional_format_object`) wants flat `properties.rule_type`. The CLI enum (`cellValue / formula / duplicate / ...`) was also a CLI-invented vocabulary none of whose values the server accepts. **Breaking**: enum aligned with server (`cellIs / duplicateValues / containsText / timePeriod / containsBlanks / notContainsBlanks / dataBar / colorScale / rank / aboveAverage / expression / iconSet`); scripts using old names now hit cobra `invalid value ..., allowed: ...`. No alias layer.
  - `+filter-{update,delete}`: server contract is `filter_id === sheet_id` and both ops MUST carry `filter_id` — the CLI translator never set it, so every call failed server-side. Translator now auto-injects `filter_id` from `sheet_id`. Because `filter_id` must equal `sheet_id` (not `sheet_name`), update/delete reject `--sheet-name`-only with a friendly error pointing at `+workbook-info`.
  - Doc artifacts that got regenerated alongside (sync from sheet-skill-spec):
    - `+dim-{hide,unhide,group,ungroup}` `--start/--end` desc changed from `(0-based, inclusive)` to `(0-based)` / `(exclusive)` to match the half-open range semantics the code has always implemented.
    - `+sheet-move` reference gains a batch-internal warning about explicit `--sheet-id` + `--source-index`.
    - `+pivot-create --properties` example drops the stale `data_range` field (data source goes through `--source`).

## Test plan

- [x] `go vet ./shortcuts/sheets/...`
- [x] `go build ./...`
- [x] `go test ./shortcuts/sheets/...` — all green, including
  - `TestBatchOp_BodyMatchesStandalone` (~40 cases): batch sub-op bodies stay byte-identical to standalone bodies.
  - `TestBatchOp_ErrorEquivalence` (7 cases): XOR / logical-constraint errors fire identically standalone vs batch.
  - `TestBatchOp_RejectsBadSubOpInput` (10 cases): cobra-required flags + filter sheet-name-only now rejected CLI-side in batch.
  - `TestObjectCRUDShortcuts_DryRun`: cond-format body uses flat `properties.rule_type`; filter update/delete auto-inject `filter_id = sheet_id`.
- [x] BOE smoke (spreadsheet `ICFwstkUGheyfptGWS2bB7RgcDf`):
  - Missing `--sheet-id` in a batch sub-op now returns `operations[0] (+dim-insert): specify at least one of --sheet-id or --sheet-name` before any network call.
  - `+cond-format-create --rule-type cellIs` writes `properties.rule_type: "cellIs"` flat; old enum `cellValue` rejected by cobra with the new enum listed.
  - `+filter-update --sheet-id X` auto-populates `filter_id: X`; `--sheet-name` only gets the friendly error.

## Notes / follow-ups

- **Breaking change** for `+cond-format-{create,update} --rule-type`: any script passing the old vocabulary needs to migrate to server enum (`cellValue → cellIs`, `duplicate → duplicateValues`, etc.). Per offline discussion no alias layer is added — earlier alignment is better than carrying two vocabularies.
- `+sparkline-update` still requires user to hand-fill `properties.sparklines[].sparkline_id`; auto list-then-update is a separate change of its own (architecture limit, not in this PR).
- This PR sits on top of the upstream `0ea7c14` / `1e05e7b` flagView abstraction; no overlap with those commits.

### DIFF
--- a/shortcuts/sheets/batch_op_contract_test.go
+++ b/shortcuts/sheets/batch_op_contract_test.go
@@ -6,6 +6,7 @@ package sheets
 import (
 	"encoding/json"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/larksuite/cli/shortcuts/common"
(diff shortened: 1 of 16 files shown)


xiongyuanwen-byted (approved):
LGTM

PR #1059, opened 2026-05-24
fix(keychain): prefer file-based master key to fix background/automat…

### LINKED ISSUE #1038: auth status --verify fails in macOS automation context: keychain Get failed: keychain not initialized
## Summary

`lark-cli --profile <profile> auth status --verify` works in an interactive desktop session, but fails consistently in an automation/background context on macOS with:

```text
keychain Get failed: keychain not initialized
```

This blocks any later Feishu doc/wiki operations before network/API access is even attempted.

## Version

```text
lark-cli version 1.0.38
```

## Environment

- macOS
- `lark-cli` installed from npm
- profile already configured and valid in interactive use
- same machine, same user account
- failure happens specifically in a recurring automation / background execution context

## Reproduction

1. Configure a normal user profile, for example:
   ```bash
   lark-cli --profile oomol auth status --verify
   ```
2. Run the same command in a foreground interactive terminal session.
3. Run the same command from an automation/background context on the same machine and same user account.

## Expected behavior

If the profile and token are valid, `auth status --verify` should either:

- succeed in the background context, or
- return a more precise error that clearly explains the required macOS/session precondition.

## Actual behavior

In the background/automation context, the command fails repeatedly with:

```json
{
  "ok": false,
  "error": {
    "type": "config",
    "message": "keychain Get failed: keychain not initialized"
  }
}
```

We retried the exact same command 3 times in the same automation run and got the same error all 3 times.

## Important observations

- The same profile succeeded earlier on the same day in an interactive context.
- `LARK_PROFILE` was not misconfigured.
- This does **not** fail at Feishu doc/wiki read/write time.
- This does **not** fail at webhook/message send time.
- The failure occurs before any Feishu API call, while `lark-cli` is reading local credentials / keychain state.

## Why this seems like a real bug / compatibility gap

From the changelog it looks like the project has already improved sandbox / keychain handling for agent environments. However, on `1.0.38`, a real macOS automation context can still reliably reproduce this failure.

This is especially problematic for users who rely on `lark-cli` inside local recurring automations, agent runners, or other non-foreground execution contexts.

## Suggested direction

Any of the following would help:

- document the exact macOS session requirements for `auth status --verify`
- provide a supported non-keychain credential mode for automation contexts
- improve detection and recovery for background/session-limited execution
- clarify whether this scenario is unsupported by design

Thanks.

Zhang-986:
## 背景

Closes #1038

macOS 上在后台/自动化场景跑 `auth status --verify` 会报 "keychain not initialized"，但同一个 profile 在交互式终端里完全正常。

原因是 `platformSet` 优先用 system keychain 的 master key 来加密凭据，而 system keychain 在没有 GUI session 的情况下根本访问不了。用户在交互式终端里 `config init` 存的密钥，到了自动化脚本里就读不出来了。

## 改了什么

**1. `platformSet` 里调换了 key 优先级**

原来：读已有 file key → system keychain（允许创建）→ 创建 file key
现在：读已有 file key → 创建 file key → system keychain（允许创建）

这样新存的凭据一律用 file key 加密，后台和交互式都能读。

**2. `platformGet` 里加了自动迁移**

当 `platformGet` 通过 system keychain 成功解密后，会自动用 file key 重新加密。老用户不需要重新配置，下次读取就自动迁移了。

## What changed

**1. Swapped key priority in `platformSet`**

Before: existing file key → system keychain (allow create) → create file key
After: existing file key → create file key → system keychain (allow create)

New credentials are now always encrypted with the file-based key, which works in both interactive and background contexts.

**2. Added automatic migration in `platformGet`**

When a secret is successfully decrypted using the system keychain key, it gets re-encrypted with the file-based key. Existing users are migrated on their next read — no manual reconfiguration needed.

## 测试

- [x] `go test ./internal/keychain/... -race` — 5 个测试全过
- [x] `go vet` / `gofmt` — 没问题
- [x] `TestPlatformSetCreatesFileMasterKeyFirst` — 验证新凭据用 file key，不碰 system keychain
- [x] `TestMigrateToFileMasterKey` — 验证迁移：system keychain 解密 → file key 重加密 → 断掉 system keychain 后仍能读取
- [x] 在 macOS 自动化环境里手动验证

## Testing

- [x] `go test ./internal/keychain/... -race` — all 5 tests pass
- [x] `go vet` / `gofmt` — clean
- [x] `TestPlatformSetCreatesFileMasterKeyFirst` — new credentials use file key, not system keychain
- [x] `TestMigrateToFileMasterKey` — migration works: decrypt with system key → re-encrypt with file key → read works without system keychain
- [x] Manual verification in a macOS automation context

## Summary by CodeRabbit

* **New Features**
  * Secrets stored in macOS system keychain are automatically migrated to local file-based storage on first access.

* **Bug Fixes**
  * Improved master key creation logic for local file-based storage when system keychain is unavailable.

* **Tests**
  * Added tests for file-based master key creation and automatic secret migration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/1059?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/keychain/keychain_darwin.go
+++ b/internal/keychain/keychain_darwin.go
@@ -269,16 +269,41 @@ func platformGet(service, account string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	migrateToFileMasterKey(service, account, plaintext)
+
 	return plaintext, nil
+}
+
+// migrateToFileMasterKey re-encrypts a secret with the file-based master key
+// so that future reads work in background/automation contexts without system keychain access.
+func migrateToFileMasterKey(service, account, plaintext string) {
+	fileKey, err := getFileMasterKey(service, true)
+	if err != nil {
+		return
+	}
+	encrypted, err := encryptData(plaintext, fileKey)
+	if err != nil {
+		return
+	}
+	dir := StorageDir(service)
+	targetPath := filepath.Join(dir, safeFileName(account))
+	tmpPath := targetPath + "." + uuid.New().String() + ".tmp"
+	if vfs.WriteFile(tmpPath, encrypted, 0600) != nil {
+		return
+	}
+	if vfs.Rename(tmpPath, targetPath) != nil {
+		_ = vfs.Remove(tmpPath)
+	}
 }
 
 // platformSet stores a value in the macOS keychain.
 func platformSet(service, account, data string) error {
 	key, err := getFileMasterKey(service, false)
 	if err != nil {
-		key, err = getMasterKey(service, true)
+		key, err = getFileMasterKey(service, true)
 		if err != nil {
-			key, err = getFileMasterKey(service, true)
+			key, err = getMasterKey(service, true)
 			if err != nil {
 				return err
 			}
--- a/internal/keychain/keychain_darwin_test.go
+++ b/internal/keychain/keychain_darwin_test.go
@@ -55,6 +55,126 @@ func TestPlatformSetFallsBackToFileMasterKey(t *testing.T) {
 	}
 }
 
+// TestPlatformSetCreatesFileMasterKeyFirst verifies that platformSet creates a
+// file-based master key before attempting system keychain access, ensuring
+// credentials stored in interactive sessions are readable in background contexts.
+func TestPlatformSetCreatesFileMasterKeyFirst(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+
+	keychainKey := make([]byte, masterKeyBytes)
+	for i := range keychainKey {
+		keychainKey[i] = byte(i + 33)
+	}
+
+	keyringGetCalled := false
+	origGet := keyringGet
+	origSet := keyringSet
+	keyringGet = func(service, user string) (string, error) {
+		keyringGetCalled = true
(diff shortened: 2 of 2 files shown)


albertnusouo:
@Zhang-986 
Thanks for taking this on — #1038 is a real pain point. For context, we ended up shipping an alternative for the same bug via [#1085](https://github.com/larksuite/cli/pull/1085) (merged): a `lark-cli config keychain-downgrade`subcommand. The two approaches differ on the implicit/explicit axis:

This PR auto-migrates inside `platformGet` / `platformSet` — transparent, zero user action.
#1085 adds an explicit subcommand the user runs once interactively; the migration is then durable.
We landed on the explicit design for two reasons:

Discoverability and reversibility — the user opts in knowingly, the OS Keychain entry is preserved as a cold backup, and the action appears in shell history.
Surfaceable security trade-off — auto-migration on read silently demotes the storage from Keychain (per-app ACL) to a 0600 file (same-user readable). An explicit command lets us document and disclose that delta in --help and in the runtime hint.
The analysis in this PR is sound — closing in favor of #1085 is one option, but feel free to leave it open if there's a remaining gap we missed.

PR #1083, opened 2026-05-25
fix(sheets): +dropdown-get accepts --sheet-id/--sheet-name + bare --range

zhengzhijiej-tech:
## Background

`+dropdown-get` was the only `get_cell_ranges` wrapper without `--sheet-id` / `--sheet-name` selector flags — it required the sheet prefix inline in `--range` (e.g. `Sheet1!A2:A100`) and **unconditionally** sent the prefix as `sheet_name`.

Callers pasting the sheet-id form straight from a spreadsheet URL hit a misleading server error:

```
$ lark-cli sheets +dropdown-get --range "7f8fba!C2:C6"
tool "get_cell_ranges" failed: [900015047]
  sheet not found, sheetId: , sheetName: 7f8fba
```

…where `7f8fba` is a valid sheet id (works on every other `+dropdown-*` / `+cells-*` / `+csv-*` shortcut via `--sheet-id`), but this wrapper forced it through the `sheet_name` field, producing the self-contradictory "sheetId empty / sheetName=<an id>" error.

## Change

Align `+dropdown-get` with its siblings `+cells-get` / `+csv-get`:

- Add `--sheet-id` / `--sheet-name` (XOR, public)
- Drop the `--range` sheet-prefix requirement; `--range` is now a bare A1 reference
- `Validate` runs through `resolveSheetSelector` like the siblings; `dropdownGetInput` takes `(sheetID, sheetName)` and forwards them as the canonical selector
- Tests migrated to the new shape; `TestDropdownGet_RequiresSheetSelector` replaces the old prefix-required guard; `TestReadData_RequiresRange` extended to cover `+dropdown-get`

Flag schema (`shortcuts/sheets/data/flag-defs.json`) and skill reference (`skills/lark-sheets/references/lark-sheets-read-data.md`) regenerated from the upstream Lark Base (Shortcut-flags table).

## Test

- `go build ./...` clean
- `go test ./shortcuts/sheets/...` passes (new `+dropdown-get with --sheet-id` / `with --sheet-name` DryRun cases + selector-required Validate case)
- `go vet ./shortcuts/sheets/` clean

## Migration (breaking for the rare direct caller)

```bash
# before
lark-cli sheets +dropdown-get --url ... --range "Sheet1!A2:A100"

# after
lark-cli sheets +dropdown-get --url ... --sheet-name "Sheet1" --range "A2:A100"
lark-cli sheets +dropdown-get --url ... --sheet-id   "7f8fba"  --range "A2:A100"
```

### DIFF
--- a/shortcuts/sheets/data/flag-defs.json
+++ b/shortcuts/sheets/data/flag-defs.json
@@ -1180,7 +1180,7 @@
         "kind": "own",
         "type": "string",
         "required": "required",
-        "desc": "A1 range, e.g. `Sheet1!A1:F10`"
+        "desc": "A1 range, e.g. `A1:F10` (no sheet prefix — use `--sheet-id` / `--sheet-name` to select the sheet)"
       },
       {
         "name": "include",
@@ -1304,7 +1304,7 @@
         "kind": "own",
         "type": "string",
         "required": "required",
-        "desc": "A1 range, e.g. `Sheet1!A1:F30`"
+        "desc": "A1 range, e.g. `A1:F30` (no sheet prefix — use `--sheet-id` / `--sheet-name` to select the sheet)"
       },
       {
         "name": "value-render-option",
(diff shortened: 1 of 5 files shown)


xiongyuanwen-byted (approved):
LGTM

PR #1098, opened 2026-05-26
fix(sheets): +dropdown source-range warns when highlight + cells > 2000

zhengzhijiej-tech:
## Summary

- **Warn at Validate time** when `+dropdown-set` / `+dropdown-update` target a `--source-range` covering more than 2000 cells while highlight is on. Matches the server-side rule in `ListFromRangeValidation.checkOptionsValid()` (`LIST_WITH_COLOR_MAX_COUNT = 2000` — highlight + large source is unsupported, server flags the dropdown as `isOptionError=true`). CLI previously had no signal for this; users only learned by seeing the dropdown render as option-error in the workbook.
- **Align `--options` / `--source-range` flag descs** with server reality. The previous "≤500 items, each ≤100 chars, no commas" claim on `--options` was fabricated — server-side `ListOfItemValidation` enforces no count, no per-item length, and accepts commas (escape-encoded on the wire). Truth source updated in the sheet-skill-spec base; resynced via `npm run sync:tool-shortcut-map` + `generate:cli` + `sync:consumers`.
- **Bundles an earlier upstream `tools-schema.json` sync** (commit `a0d6472`) that hadn't landed yet; included here so `npm run check:tool-schemas` stays green after the spec sync.

## Why

Aligns three artifacts that disagreed: CLI help text claimed limits that didn't exist server-side; server's only real limit (highlight + 2000 cells) had no surfacing in CLI; PPE regression probes asserted the fabricated CLI claims, so they failed every run.

The fix layer is CLI because the limits live in CLI's flag descs and the soft cap can be checked at Validate time without an extra RPC. No server change needed — byted-sheet behavior is already correct.

## Test plan

- [ ] `go build ./...` and `go test ./shortcuts/sheets/...` pass
- [ ] `lark-cli sheets +dropdown-set --help` shows updated `--options` and `--source-range` descs after rebuild
- [ ] PPE: `+dropdown-set --source-range Sheet1!A1:B2000 --options ...` succeeds without warning; same with B2001 emits `warning: --source-range covers 4002 cells; server marks the dropdown as option-error when highlight is on...` to stderr but still returns `ok:true`
- [ ] PPE: `+dropdown-set ... --source-range <large> --highlight=false` suppresses the warning
- [ ] PPE: `+dropdown-set ... --options '[<501 items>]'` succeeds (server has no item cap)

## Companion changes

Spec source-of-truth MR (internal): `ee/sheet-skill-spec!11 docs(spec): sync dropdown flag descs from base — strip fabricated limits`

## Commits

- `a0d6472` feat: 同步 tools-schema.json 改动 (pre-existing upstream drift)
- `bb7ccae` fix(sheets): warn when +dropdown source-range exceeds 2000 cells with highlight on
- `930c9c7` docs(sheets): sync lark-sheets skill from spec — dropdown flag descs reflect server reality


vinnyxiong (approved):
LGTM

PR #1123, opened 2026-05-27
feat: add iconpark lookup for lark slides

ethan-zhx:
## Summary
Add IconPark lookup support for the lark-slides skill, including reference docs, index data, and a helper script for searching and selecting icons.

## Changes
- Add IconPark reference docs and generated icon index for lark-slides.
- Add `iconpark_tool.py` with lookup helpers and tests.
- Update lark-slides skill/reference docs to mention IconPark usage.

## Test Plan
- [ ] Unit tests pass
## Related Issues 
- None

## Summary by CodeRabbit

* **New Features**
  * Added a command-line tool for searching and resolving semantic icons with keyword queries and validation support.

* **Documentation**
  * Enhanced documentation on semantic icon usage in slides, including lookup workflows and fallback options.
  * Added icon usage reference guide covering conventions, sizing, and color contrast requirements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/1123?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


fangshuyu-768 (commented):
Overall I did not find a blocking functional bug in this PR, but I do see a few non-blocking follow-ups worth tracking:

1. Testing coverage
The new test file exercises the helper APIs well, but it does not cover the CLI contract yet. It would be helpful to add a few end-to-end CLI tests for `search`, `resolve`, `list-categories`, `--help`, and invalid argument handling, so future refactors do not silently change exit codes or stderr behavior.

2. Ranking maintainability
`iconpark_tool.py` relies on a heuristic combination of `CURATED_ICON_BOOSTS`, token expansion, and the generated offline index. That looks reasonable for now, but ranking behavior can drift when the index is regenerated or when boost rules are adjusted. A small set of deterministic top-result contract tests for the highest-value queries would make this much safer to evolve.

3. Data refresh risk
Because this feature depends on a large generated `iconpark-index.json`, the long-term correctness is partly a data maintenance problem, not only a code problem. I would recommend documenting the regeneration/update workflow for that index somewhere near the script or reference docs, so future updates stay reproducible and reviewers can tell whether a ranking change came from code or from data refresh.

None of these look blocking to me for the current diff, but they are the main deeper risks I would keep an eye on.

skills/lark-slides/scripts/iconpark_tool_test.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2026 Lark Technologies Pte. Ltd.
+# SPDX-License-Identifier: MIT
+from __future__ import annotations
+
+import unittest
+
+import iconpark_tool
+
+
+class IconParkToolTest(unittest.TestCase):
```
fangshuyu-768, line 17: The helper-level coverage here is good, but this file still leaves the CLI contract untested. Since `iconpark_tool.py` is intended to be invoked directly from the skill workflow, I would strongly consider adding a few end-to-end CLI tests for `search`, `resolve`, `list-categories`, `--help`, and invalid argument handling. That would make future refactors much less likely to accidentally change exit codes or stderr behavior without being noticed.

skills/lark-slides/scripts/iconpark_tool.py
```diff
@@ -0,0 +1,362 @@
...
+        fail("iconpark index must contain an icons array")
+    return index_data
+
+
+def icon_search_text(entry: dict[str, Any]) -> str:
+    parts = [
+        entry.get("iconType", ""),
+        entry.get("category", ""),
+        entry.get("name", ""),
+        " ".join(entry.get("tags") or []),
```
fangshuyu-768, line 165: This ranking logic looks reasonable for the current use case, but it is also where long-term behavior drift is most likely to show up as the curated boosts and offline index evolve. It would be helpful to protect a small set of highest-value queries with deterministic top-result contract tests, so reviewers can distinguish an intentional ranking change from an accidental one after future data refreshes or scoring tweaks.

fangshuyu-768 (approved):

PR #1148, opened 2026-05-28
perf(im): parallelize thread_replies fetches with bounded concurrency

sammi-bytedance:
## Summary

`ExpandThreadReplies` previously issued a strictly-serial GET per unique `thread_id`. Each call is ~1s RTT, so chats with N thread roots paid an N×~1s stall before the main message output could render — the same latency-multiplier pattern that motivated #1146 (reactions). This PR parallelizes the per-thread fetches with the same bounded fan-out pattern and consolidates the follow-up sender-name resolution into a single batched contact API call. Scope is limited to `convert_lib/thread.go` + its test.

## Changes

- **Phase 1 — plan + concurrent fetch.** Walk messages in first-seen order to enumerate unique `thread_id`s and allocate a per-thread limit greedily from the remaining `totalLimit` budget (threads past the budget skip outright, preserving the existing semantic). Dispatch the planned `fetchThreadReplies` calls under bounded concurrency (cap = 4 in flight, same as the reactions PR). Each goroutine writes only to its own `results[i]` slot — no shared mutable state.
- **Phase 2 — sequential attach.** Stays single-threaded because `FormatMessageItem` may trigger a `merge_forward` sub-message GET that writes to the shared `nameCache`. Phase 2 is now split into three sub-steps:
  - 2a) `FormatMessageItem` each thread's raw replies.
  - 2b) **One** batched `ResolveSenderNames` across **all** prepared replies (was previously called once per thread, potentially firing a contact API request per thread that introduced a new sender).
  - 2c) `AttachSenderNames` + write `thread_replies` onto each plan's host outer message (the first message that referenced that `thread_id` — preserves existing duplicate handling).
- **Semantics preserved.** First outer message per `thread_id` is still the host. `totalLimit` budget honored (with a documented mild pessimism: budget is allocated based on the planned per-thread limit, not the actual returned count). `thread_replies_error: true` on fetch failure unchanged. `thread_has_more: true` propagated unchanged.

## Test Plan

- [x] All three pre-existing tests pass unchanged (`TestExpandThreadReplies`, `TestFetchThreadRepliesError`, `TestExpandThreadRepliesMarksFetchError`) — they exercise the single-thread fast path.
- [x] New `TestExpandThreadRepliesMultiThreadConcurrent`: 5 distinct thread roots → 5 planned fetches under the bounded fan-out. Each thread's reply is tagged with its own `thread_id` so a goroutine cross-contamination bug would manifest as mis-attached replies. Assertion verifies `callCount=5` and every host received the right reply.
- [x] `go test -race -count=1 ./shortcuts/im/convert_lib/` clean.
- [x] `gofmt -l .` / `go vet ./...` clean.
- [x] Live `lark-cli im +chat-messages-list --chat-id <oc_xxx> --page-size 50 --format json --as user --no-reactions` against a real group chat with 4 thread roots:

  | variant | `--no-reactions` median wall time |
  |---|---|
  | before (serial fetches) | 14.3s |
  | after (parallel + batched sender resolve) | **12.8s** |

  Modest delta on this specific chat (~1.5s) because only 4 thread roots → serial path was 4 × ~1s of which parallel saves ~3 of 4. Improvement scales linearly with thread-root count: a chat with 10 thread roots would save proportionally more.

## Related Issues

- Orthogonal to #1146 (reactions parallelization). The two optimizations compose: with both merged, a busy chat's `--no-reactions` wall time drops further than either alone.
- Out of scope here (separate follow-ups when needed): parallelizing `merge_forward` sub-message expansion, and fixing the `fetchMergeForwardSubMessages` error-masking bug (`code != 0` with `data: null` currently surfaces as a generic `"empty data"` error, hiding the real `code` / `msg` / `log_id`).

## Summary by CodeRabbit

* **Performance**
  * Optimized thread reply fetching with enhanced concurrency handling.

* **Bug Fixes**
  * Improved error handling for thread reply fetch failures.

* **Tests**
  * Added test coverage for concurrent multi-thread reply expansion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/1148?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/im/convert_lib/thread.go
+++ b/shortcuts/im/convert_lib/thread.go
@@ -6,6 +6,7 @@ package convertlib
 import (
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/larksuite/cli/shortcuts/common"
 	larkcore "github.com/larksuite/oapi-sdk-go/v3/core"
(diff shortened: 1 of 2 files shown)


sammi-bytedance:
Folding this into #1146 so both perf changes ship in one PR. Closing in favor of that.

PR #1163, opened 2026-05-28
feat: support trusted remote auth proxy

zhexuany:
## Summary
Adds a trusted remote HTTPS auth proxy path for environments where lark-cli cannot receive app secrets directly.

## Changes
- Add `lark-cli config auth-proxy trust|untrust|list` to store trusted remote proxy hosts in local config instead of env.
- Allow `LARKSUITE_CLI_AUTH_PROXY=https://host[:port]` only when the host is locally trusted, while keeping HTTP constrained to same-host sidecar mode.
- Split remote proxy session from request signing: `LARKSUITE_CLI_PROXY_SESSION` is sent to the proxy, while `LARKSUITE_CLI_PROXY_KEY` signs requests and is not transmitted.
- Harden proxy endpoint and `X-Lark-Proxy-Target` parsing, and share target validation with the demo servers.

## Test Plan
- [x] `go test ./sidecar -count=1`
- [x] `go test -tags authsidecar ./extension/credential/sidecar ./extension/transport/sidecar -count=1`
- [x] `go test -tags authsidecar_demo ./sidecar/server-demo -count=1`
- [x] `go test -tags authsidecar_multi_tenant_demo ./sidecar/server-multi-tenant-demo -count=1`
- [x] `go test ./cmd/... ./internal/... ./shortcuts/... ./extension/... -count=1`
- [x] `go test -tags authsidecar ./cmd/... ./internal/... ./shortcuts/... ./extension/... ./sidecar -count=1`
- [x] `go vet ./...`
- [x] `gofmt -l .`
- [x] `go mod tidy` (no `go.mod` / `go.sum` diff)
- [x] `git diff --check`
- [ ] `make unit-test` blocked locally: Go `go1.26.0 darwin/arm64` crashes in `runtime.systemstack_switch` under `-race -gcflags="all=-N -l"`; observed as runtime SIGSEGV before package assertions run.

## Related Issues
- None

## Release Notes

* **New Features**
  * Added `config auth-proxy` command to manage trusted remote authentication proxy hosts with `trust`, `untrust`, and `list` subcommands.
  * Introduced support for remote HTTPS managed authentication proxies alongside existing local sidecar mode, with trust validation to ensure secure connections.

* **Tests**
  * Comprehensive test coverage for auth-proxy command functionality, configuration persistence, and remote proxy validation.


zhexuany:
@liangshuo-1 PTAL

We want to manage internal user login via a single gate.

liangshuo-1:
Thanks for working on this. My main concern is around the subjective / agent-safety risk.

Trusting a remote auth proxy is effectively approving a new egress path for credentials and business traffic. Once a host is trusted, subsequent requests, proxy session material, and business data can flow through it. Right now, `config auth-proxy trust` looks like a normal `write` config change and can also be run from an agent environment, so it may become “the agent approved a new remote proxy by itself” rather than “the user explicitly approved it.”

zhexuany:
Agreed. Treating auth-proxy trust as a normal config write is too weak because it grants a new egress path for credential/session material and business traffic.

  I’ll change `config auth-proxy trust` into a human-consent boundary:
  - mark only `trust` as `high-risk-write`
  - require explicit confirmation
  - refuse `trust` from detected agent workspaces, with no force override
  - keep `list` as read and `untrust` as write
  - add regression tests proving an agent workspace cannot create trust, even with confirmation flags

zhexuany:
Updated the PR to make remote auth-proxy trust an explicit user authorization boundary.

  `config auth-proxy trust` is now marked `high-risk-write`, requires `--yes`, and refuses to run from detected agent workspaces even when `--yes` is passed. The trust policy is stored in the user's base config, so agents can use hosts
  the user already approved but cannot approve a new remote proxy themselves. The confirmation error also states that future lark-cli requests, proxy session material, and business request bodies may flow through the trusted host.

PR #1165, opened 2026-05-28
fix(im): correct 64-bit MP4 box size handling to prevent panic on crafted media

lhfer:
## Summary

The three MP4 box walkers in `shortcuts/im/helpers.go` mishandle the 64-bit ("largesize") box-size form. They set `boxEnd` to the raw `largesize` instead of `offset + largesize` — even though the 32-bit branch directly below correctly uses `offset + size` — and convert the `uint64` to a signed int without any bounds check. This silently loses the duration of MP4s that carry a 64-bit box size at a non-zero offset, and **panics the CLI** (`slice bounds out of range`) on a crafted or corrupt MP4 whose `largesize` has the high bit set. The in-memory walkers run on URL-sourced IM media, whose bytes the caller does not control, so a malformed download crashes the process instead of degrading gracefully.

## Changes

- `findMP4Box`, `readMp4DurationBytes`, `readMp4Duration`: compute `boxEnd = offset + largesize`, mirroring the existing 32-bit `offset + size`.
- Reject `largesize` values smaller than the 16-byte 64-bit header or larger than the remaining input, so malformed media returns `0` / `-1` (the parsers' documented best-effort contract) instead of overflowing `uint64 → int(64)` into a negative `boxEnd` and panicking (CWE-190). The lower bound also guarantees the walk always makes forward progress.
- Add `shortcuts/im/mp4_box_test.go` covering the overflow (must not panic) and a 64-bit box at a non-zero offset (must walk correctly).

## Test Plan

- [x] `go test -race ./shortcuts/im/...` passes. The added regression tests panic / fail on the pre-fix code and pass after the fix.
- [x] `go build ./...`, `go vet ./...`, and `gofmt -l` are clean; no new dependencies (`go.mod` / `go.sum` unchanged).
- [ ] Manual `lark-cli <domain> <command>` verification — this is an internal binary-parser fix with no change to any shortcut's flags, params, or request structure, so per `AGENTS.md` no dry-run E2E is required; behavior is covered by the added unit tests.

## Related Issues

- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened MP4 duration parsing to validate 64-bit box sizes before use, preventing integer overflows, out-of-bounds reads, and crashes when encountering malformed or oversized boxes.

* **Tests**
  * Added tests covering 64-bit box size scenarios, malformed largesize values, and non-zero offsets to ensure duration extraction and box searching behave correctly without panics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/larksuite/cli/pull/1165?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/shortcuts/im/helpers.go
+++ b/shortcuts/im/helpers.go
@@ -543,7 +543,17 @@ func findMP4Box(data []byte, start, end int, boxType string) (int, int) {
 			if offset+16 > end {
 				return -1, -1
 			}
-			boxEnd = int(binary.BigEndian.Uint64(data[offset+8:]))
+			// 64-bit "largesize" is the whole box length including its 16-byte
+			// header, so the box ends at offset+largesize (mirroring the
+			// offset+size used for 32-bit boxes below). Reject sizes that do not
+			// fit the search window; this also rejects values that would
+			// overflow int and drive boxEnd negative (CWE-190), which would
+			// otherwise index data out of range and panic.
+			largesize := binary.BigEndian.Uint64(data[offset+8:])
+			if largesize < 16 || largesize > uint64(end-offset) {
+				return -1, -1
+			}
+			boxEnd = offset + int(largesize)
 			dataStart = offset + 16
 		default:
 			if size < 8 {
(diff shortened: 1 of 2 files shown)


YangJunzhou-01:
Thanks for the fix. I reviewed the change and it looks good to me.

The implementation correctly treats 64-bit MP4 `largesize` as the box length and computes `boxEnd` as `offset + largesize`, with bounds validation before converting to signed offsets. This addresses both the non-zero-offset duration parsing bug and the crafted/corrupt MP4 panic risk.

The added regression tests cover the overflow/no-panic case and 64-bit boxes at non-zero offsets.

Verified with:
- go test ./shortcuts/im -run 'TestMP4Box|TestParseMp4Duration|TestParseMediaDuration' -count=1
- go test -race ./shortcuts/im

LGTM, thanks for the careful fix.

YangJunzhou-01 (approved):

lhfer:
Thank you for the review!

PR #1220, opened 2026-06-02
feat(sheets): spec-driven shortcut refactor with backward-compatible package

xiongyuanwen-byted:
## Summary
Merges the spec-driven `sheets` shortcut refactor into `main`, together with a `shortcuts/sheets/backward/` compatibility layer that keeps the pre-refactor shortcut names and flags working. Latest `main` is folded in (transport refactor #1213, global `--format` injection #1156).

## Changes
- **Sheets refactor (92 non-merge commits, scope almost entirely `sheets`)**: flags and JSON schemas aligned with the server / upstream spec; fixes across pivot / chart / filter-view / dropdown / float-image / dimension ops; read-cap and A1-reference handling; skill docs regenerated from the upstream sheet-skill-spec (22 feat / 20 fix / 29 docs / 7 refactor / 3 revert).
- **Breaking**: `+dim-resize` split into `+rows-resize` and `+cols-resize`.
- **Backward-compat package** `shortcuts/sheets/backward/` (+9850 lines, registered via `shortcuts/register.go`): restores pre-refactor sheet shortcuts (cell data / style / images, dropdown, filter views, float images, row-column & spreadsheet management) with full unit-test coverage.
- **main sync**: resolves the transport-refactor conflict by porting `WarnIfProxied(..., interactive)` (non-interactive proxy-warning silencing) into the new `internal/transport` package.

## Test Plan
- [x] Unit tests pass — `go build ./...` clean; `go test` green for `shortcuts/sheets`, `shortcuts/sheets/backward`, `shortcuts/common`, `internal/transport`, `internal/cmdutil`
- [ ] Manual local verification of `lark-cli sheets <command>` flows (not run in this session)

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * CLI “did you mean” suggestions for unknown flags and subcommands; improved root-level error handling.
  * --print-schema to inspect sheets shortcut flag schemas.
  * Extensive new Sheets capabilities: workbook/sheet structure, read/write cells, range ops, batch-update, object CRUD/list, search/replace, export, image & chart workflows.

* **Bug Fixes & Improvements**
  * Structured, hinting flag errors; suppress usage noise on flag-parse failures.
  * Suppress proxy warnings in non-interactive runs; strip UTF‑8 BOM from inputs; schema-driven flag validation.

* **Deprecations**
  * Older sheets aliases marked deprecated with migration hints.


shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -0,0 +1,816 @@
...
+}
+
+func csvPutInput(runtime flagView, token, sheetID, sheetName string) (map[string]interface{}, error) {
+	if err := requireSheetSelector(sheetID, sheetName); err != nil {
+		return nil, err
+	}
+	if strings.TrimSpace(runtime.Str("csv")) == "" {
+		return nil, common.FlagErrorf("--csv is required")
+	}
+	anchor := strings.TrimSpace(runtime.Str("start-cell"))
```
fangshuyu-768, line 305: This looks inconsistent with the standalone CLI contract. In a batch sub-op, if neither `start-cell` nor `range` is provided, `runtime.Str("start-cell")` can still resolve to the default `A1` via `mapFlagView`, so the operation writes to `A1` instead of failing like the standalone command would. Could we require that either `start-cell` or `range` is explicitly set before accepting the sub-op?

shortcuts/sheets/batch_op_dispatch.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	if !fv.Changed("source-index") {
+		return nil, common.FlagErrorf("+sheet-move in +batch-update requires source_index (auto-derive needs a network lookup unavailable mid-batch)")
+	}
+	return map[string]interface{}{
+		"excel_id":     token,
+		"operation":    "move",
+		"sheet_id":     sheetID,
+		"source_index": fv.Int("source-index"),
+		"target_index": fv.Int("index"),
```
fangshuyu-768, line 234: This no longer enforces that `index` is explicitly provided, even though the standalone `+sheet-move` path rejects a missing `--index`. Because `mapFlagView` falls back to flag defaults, a batch sub-op can silently produce `target_index=0` here instead of failing validation. I think the batch translator should keep parity with the standalone contract and reject the sub-op unless `index` was actually set.

skills/lark-sheets/SKILL.md
```diff
@@ -1,343 +1,156 @@
...
+   - ⚠️ **不确定 sheet 名时禁止直接猜 `Sheet1`**：除非用户对话明确说出 sheet 名 / id，或上下文（之前的工具调用 / URL 锚点 `?sheet=xxx`）已经出现过具体值，否则**第一步先调 `+workbook-info --url "..."`**（或 `--spreadsheet-token`）拿 `sheets[].sheet_id` / `sheets[].title` 列表再选。中文环境下子表常叫"数据" / "Sheet"（无数字）/ "工作表 1" / 业务名，猜 `Sheet1` 大概率撞 `sheet not found`，比先查多耗一次失败调用 + 重试。
+   - ⚠️ **`--range` 里的 `Sheet1!` 前缀不能替代 sheet 定位**：即使写了 `--range 'Sheet1!A1:B2'`，仍**必须**额外传 `--sheet-id` 或 `--sheet-name`，否则照样报上面的错。
+   - ⚠️ **A1 reference 含 `!`**（`--source` / `--range` / `--ranges`）**：shell session 起手先 `set +H`** 关 bash history expansion，否则 `"Sheet1!A1"` 会被拦成 `event not found`；含特殊字符（`-` / 空格 / 非 ASCII）的 sheet 名还要内部 single-quote 包，如 `--source "'Sales-2025'!A1:D100"`。
+   - **例外**：徽章标为 `_公共：URL/token（无 sheet 定位）…_` 的 shortcut（如 `+workbook-info` / `+workbook-export` / `+batch-update` / `+dropdown-get|update|delete` / `+cells-batch-set-style` / `+cells-batch-clear` / `+sheet-create`）**不接受也不需要** sheet 定位，只给一组 spreadsheet 定位即可。`+pivot-create` 用 `--target-sheet-id` / `--target-sheet-name`（XOR，可都不传，落点细节见 `lark-sheets-pivot-table`）。
```
fangshuyu-768: The documentation here seems out of sync with the implementation. `+dropdown-get` is listed as an exception that does not require sheet selection, but the code now calls `resolveSheetSelector(...)` and still requires `--sheet-id` / `--sheet-name`. I think this should be updated, otherwise agents following the skill literally will generate an invalid call.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -0,0 +1,1000 @@
...
+		return "", 0, output.Errorf(output.ExitAPI, "tool_output", "get_workbook_structure returned non-object output")
+	}
+	sheets, _ := m["sheets"].([]interface{})
+	for _, raw := range sheets {
+		sm, ok := raw.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		id, _ := sm["sheet_id"].(string)
+		name, _ := sm["sheet_name"].(string)
```
fangshuyu-768, line 971: This lookup is matching `sheet_name`, but the workbook structure output used elsewhere in this PR appears to expose sheet titles as `title` instead. If the real tool response does not include `sheet_name`, every `--sheet-name` path that depends on `lookupSheetIndex()` (for example `+sheet-move`) will fail to resolve the target sheet. It would be safer to match the actual field returned by `get_workbook_structure`, or support both keys defensively.

shortcuts/sheets/backward/helpers.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func normalizeWriteRange(sheetID, input string, values interface{}) string {
+	rows, cols := matrixDimensions(values)
+	input = normalizeSheetRangeSeparators(input)
+	if input == "" {
+		return buildRectRange(sheetID, "A1", rows, cols)
+	}
+
+	input = normalizeSheetRange(sheetID, input)
```
fangshuyu-768, line 90: This normalization path looks risky for backward compatibility. The legacy `+read` / `+write` / `+append` commands first fall back to `sheet-id` when `range` is omitted, and then route that value through the range normalizer. If a sheet id happens to look A1-ish, it can be reinterpreted as a cell/range anchor instead of staying a pure sheet selector. I think the old "sheet-only" fallback should bypass range normalization entirely.

shortcuts/sheets/lark_sheet_write_cells.go
zhengzhijiej-tech: Addressed in da198cf0. csvPutInput now guards on runtime.Changed("start-cell") (not emptiness, since it defaults to A1) and requires --range when --start-cell wasn't explicitly set, mirroring the standalone MarkFlagsOneRequired contract — it now errors with `--start-cell or --range is required` instead of silently writing to A1.

shortcuts/sheets/batch_op_dispatch.go
zhengzhijiej-tech: Addressed in da198cf0. sheetMoveBatchInput now rejects the sub-op with `+sheet-move in +batch-update requires index` unless fv.Changed("index"), so it no longer silently falls back to target_index=0. Matches the standalone +sheet-move contract.

shortcuts/sheets/lark_sheet_workbook.go
zhengzhijiej-tech: Addressed in 27c63336. lookupSheetIndex now falls back to sm["title"] when "sheet_name" is absent, so --sheet-name resolves regardless of which field get_workbook_structure returns. Regression test added (TestExecute_SheetMove_LookupsIndexByTitle).

shortcuts/sheets/backward/helpers.go
zhengzhijiej-tech: Addressed in 27c63336. The sheet-only fallback in +read/+write/+append now bypasses range normalization: read/append pass the bare sheet id through verbatim, write builds the rect from the selector's A1. This stops A1-ish sheet ids (e.g. shtABC123) from being mangled into shtABC123!shtABC123:shtABC123. Regression tests added.

skills/lark-sheets/references/lark-sheets-workbook.md
```diff
@@ -0,0 +1,197 @@
...
+| `--sheet-id` | string | optional | 仅 csv 模式必填：指定要导出哪张 sheet 为 CSV。这是 `+workbook-export` 专有 flag，与公共四件套的 sheet 定位无关（本 shortcut 不接受公共 sheet 定位） |
+| `--output-path` | string | optional | 本地保存路径；省略时只触发导出不下载 |
+
+## Examples
+
+公共四件套：所有 shortcut 顶部排列 `--url` / `--spreadsheet-token` / `--sheet-id` / `--sheet-name`（XOR）。`+workbook-info` 只用前两者；`+sheet-*` 系列对单个工作表操作，需 `--sheet-id` 或 `--sheet-name`。
+
+### `+workbook-info`
+
+输出契约：返回 `sheets[]`，每个含 `sheet_id` / `sheet_name` / `row_count` / `column_count` / `index` / `is_hidden`，以及计数字段 `merged_cells_count` / `chart_count` / `pivot_table_count` / `float_image_count`（无 `frozen_*` 字段，冻结信息请用 `+sheet-info` 读取）。是操作飞书表格的第一步——任何后续 sheet 级动作都需要先拿这里的 sheet_id。
```
fangshuyu-768: This output contract looks stale now. The current implementation and regression tests already treat the sheet display name as `title` (with `sheet_name` only as a fallback for older/alternate payloads), so documenting `sheets[].sheet_name` as the canonical field here is likely to mislead skill consumers. I would update this section to reflect the actual returned field name, or explicitly document both fields and their precedence.

skills/lark-sheets/references/lark-sheets-workbook.md
```diff
@@ -0,0 +1,197 @@
...
+
+```bash
+lark-cli sheets +sheet-rename --url "..." --sheet-id "$SID" --title "汇总"
+```
+
+### `+sheet-move`
+
+standalone 路径在缺 `--source-index` / 只给 `--sheet-name` 时会自动发起一次 `+workbook-info` 读把它们解出来。
+
+> ⚠️ **在 `+batch-update` 内调用 `+sheet-move`**：必须同时显式传 `--sheet-id` 和 `--source-index`。batch 中途无法发起结构查询，所以 batch translator 强制要求两者都显式。
```
fangshuyu-768: This batch-specific note is now incomplete. The translator no longer only requires `sheet_id` and `source_index`; it also explicitly requires `index`, to prevent the batch sub-op from silently falling back to the default target position `0`. Without mentioning `--index` here, the documented batch shape will still lead agents to construct a request that fails validation.

skills/lark-sheets/references/lark-sheets-workbook.md
zhengzhijiej-tech: Addressed in 46712cf9. Updated the +workbook-info output contract: the sheet display name is now documented as `title` (with `sheet_name` as a legacy fallback, `title` taking precedence on read), matching the implementation and regression tests.

skills/lark-sheets/references/lark-sheets-workbook.md
zhengzhijiej-tech: Addressed in 46712cf9. The batch note now lists `--index` alongside `--sheet-id` and `--source-index`, noting that an omitted `--index` would silently fall back to target position 0, so the batch translator requires all three explicitly.

zhengzhijiej-tech (dismissed):

fangshuyu-768 (dismissed):

caojie0621 (dismissed):

zhengzhijiej-tech (approved):

caojie0621 (approved):

fangshuyu-768 (approved):

PR #1250, opened 2026-06-03
feat(mail): return typed error envelopes across the mail domain

evandance:
## Summary

`shortcuts/mail` returned many command-facing errors through legacy `output.Err*` / `output.Errorf` / final `fmt.Errorf` paths and legacy API-call envelopes, so consumers had to recover classification from free-form message text. This migrates the mail domain to typed `errs.*` envelopes, giving callers stable `type`, `subtype`, `param` / `params`, `hint`, `retryable`, and `log_id` metadata across validation, API, network, file-I/O, and partial-failure flows. The shared drive-media upload helpers gain additive typed counterparts so mail's attachment uploads emit typed envelopes as well.

## Changes

- Migrate command-facing errors in `shortcuts/mail` from legacy output constructors, common legacy wrappers, and final bare `fmt.Errorf` returns to typed `errs.*` builders
- Convert mail API boundaries from legacy `runtime.CallAPI` / JSON helpers to typed calls or `runtime.ClassifyAPIResponse`, and update mail-specific enrichers to read and preserve `errs.ProblemOf` metadata
- Add `shortcuts/mail/mail_errors.go` with mail-local typed helpers for validation, file I/O, invalid-response, problem-message decoration, and hint enrichment
- Add `UploadDriveMediaAllTyped` / `UploadDriveMediaMultipartTyped` to `shortcuts/common` and deprecate the legacy pair (legacy implementations unchanged); mail's large-attachment and template-attachment uploads consume the typed variants so upload failures carry typed subtype / code / log_id
- Register mail API-code metadata for stable retry/quota hints and keep command-specific guidance in local enrichment paths
- Emit batch draft-send partial failures through `runtime.OutPartialFailure` so successful and failed draft sends stay structured in stdout while the command exits with a typed multi-status signal; account-level fatal classification reads typed category/subtype/code (unregistered not_found codes stay per-draft recoverable)
- Folder/label flags now resolve both IDs and names; `--message-ids` accepts unpadded base64url IDs; attachment and patch-file paths are validated with `SafeInputPath`/`ValidatePath` before reading; unknown-flag errors emit typed validation envelopes with candidate suggestions in `params`
- Update mail tests and fakes to assert typed errors, structured params, typed API metadata preservation, and partial-failure envelopes; add typed upload coverage in `shortcuts/common`
- Enforce typed-only mail errors through golangci path exceptions and the errscontract legacy-envelope / legacy-API-call guards

## Test Plan

- [x] `gofmt -l .`
- [x] `go vet ./...`
- [x] `go test -race ./shortcuts/mail/... ./shortcuts/ ./shortcuts/common/ ./internal/errclass/ ./internal/output/ ./errs/`
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run --new-from-rev=origin/main ./shortcuts/mail/... ./shortcuts/common/... ./internal/errclass/...` — 0 issues
- [x] `go run -C lint . ..`
- [x] `go build ./...`

## Related Issues

N/A

### DIFF
--- a/shortcuts/mail/mail_draft_send.go
+++ b/shortcuts/mail/mail_draft_send.go
@@ -5,11 +5,10 @@ package mail
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
-	"github.com/larksuite/cli/internal/output"
+	"github.com/larksuite/cli/errs"
 	"github.com/larksuite/cli/shortcuts/common"
 )
 
@@ -66,9 +65,9 @@ type batchSendOutput struct {
 // drafts are user-owned resources and bot has no coherent semantics here.
 //
 // Output schema is the batchSendOutput type above. Partial failures (any
-// failed[]) return exit 1 with envelope.error.type="partial_failure" so that
-// agents can distinguish "all sent" from "some sent" without parsing the
-// success_count field.
+// failed[]) emit an ok:false multi-status envelope so that agents can
+// distinguish "all sent" from "some sent" without parsing the success_count
+// field.
 var MailDraftSend = common.Shortcut{
 	Service: "mail",
 	Command: "+draft-send",
@@ -101,14 +100,14 @@ var MailDraftSend = common.Shortcut{
 //  2. Validate the draft-id slice (non-empty, under MaxBatchSendDrafts cap,
 //     no empty elements).
 //  3. Loop over each draft ID, calling POST .../drafts/:id/send directly via
-//     runtime.CallAPI. Per-draft outcomes:
+//     runtime.CallAPITyped. Per-draft outcomes:
 //     - fatal err (isFatalSendErr) → return immediately (bypasses --stop-on-error).
 //     - recoverable err → append to failed[]; honor --stop-on-error.
-//     - success + automation_send_disable signal → return immediately with
-//     ExitAPI/"automation_send_disabled".
+//     - success + automation_send_disable signal → return immediately with a
+//     failed-precondition error.
 //     - success → append to sent[].
 //  4. Emit batchSendOutput via runtime.Out.
-//  5. If any draft failed, return ExitAPI/"partial_failure" so exit code = 1.
+//  5. If any draft failed, emit ok:false via runtime.OutPartialFailure.
 func executeDraftSend(ctx context.Context, rt *common.RuntimeContext) error {
 	mailboxID := resolveComposeMailboxID(rt)
 	draftIDs, err := normalizedDraftSendIDs(rt)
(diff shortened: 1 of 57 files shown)


shortcuts/mail/mail_draft_send.go
```diff
@@ -133,7 +132,7 @@ func executeDraftSend(ctx context.Context, rt *common.RuntimeContext) error {
 				hadProgress := out.hasProgress()
 				out.Failed = append(out.Failed, failedDraft{DraftID: id, Error: err.Error()})
 				if hadProgress {
-					emitDraftSendOutput(rt, &out)
+					_ = emitDraftSendPartialFailure(rt, &out)
```
chanthuang: **Issue:** When a fatal send error occurs after earlier progress, this branch writes an `ok:false` partial-failure ledger to stdout and then returns the fatal typed error. The root dispatcher will still render that returned typed error as a second failure envelope on stderr. A single CLI invocation therefore exposes two machine-readable failure results for the same operation, which can make agents disagree on whether to recover from the batch ledger or from the fatal envelope. The same pattern exists in the automation-disabled branch below.

**Severity:** 🚫 Must Fix

**Suggested Fix:** Emit exactly one failure contract for this path. The safest shape is to treat "fatal after progress" as an aborted batch result: add structured fields to the batch ledger such as `aborted: true`, `abort_reason`, and optionally `abort_error` / `fatal: true` on the failed draft, then return `emitDraftSendPartialFailure(...)` so `handleRootError` only sets the exit code and writes no additional stderr envelope. If no draft has made progress yet, returning the fatal typed error directly is still fine because stdout is empty. Also update the fatal-after-success and automation-disabled-after-success tests to assert that the full command emits only one JSON failure result.

evandance: Fixed in 39241145, following your suggested shape: when the batch aborts after some drafts already went out, the stdout ledger is now the only failure result — it carries `aborted: true` and `abort_error` (the typed cause), and the command returns the partial-failure signal so nothing extra goes to stderr. Aborts with no progress still return the typed error directly, and the automation-disabled branch works the same way.

One simplification: I used `abort_error` (the structured object you marked optional) and dropped `abort_reason`, since the object's `message` already covers it and also carries code/subtype/hint for recovery.

Tests now assert a single failure result for both abort paths, plus one asserting `--stop-on-error` does not set `aborted`.

chanthuang (approved):
LGTM. The aborted +draft-send path now emits the batch ledger as the single failure result and returns the envelope-less partial-failure signal, so the previous double-envelope concern is resolved.\n\nVerified locally: ok  	github.com/larksuite/cli/shortcuts/mail	(cached) and ok  	github.com/larksuite/cli/shortcuts/mail	1.669s
ok  	github.com/larksuite/cli/shortcuts/common	0.595s.\n\n

chanthuang (approved):

PR #1259, opened 2026-06-04
feat(drive): add drive preview and cover shortcuts and document quota details

zhoujunteng-max:
## Summary

This PR adds `drive +preview` and `drive +cover` shortcuts for Drive file preview artifact and cover downloads, and updates the `lark-drive` skill metadata to expose quota detail capability and the new shortcuts to AI agents.

## Changes

- add `drive +preview` to list and download available preview artifacts
- add `drive +cover` to list and download built-in cover presets
- normalize preview result wrapping and cover parameter mapping based on backend validation rules
- add unit tests, dry-run E2E tests, and live workflow E2E tests for preview/cover flows
- update `skills/lark-drive/SKILL.md` and add preview/cover references
- document `quota_details.get` in `skills/lark-drive/SKILL.md`

## Test Plan

- [x] Unit tests pass
- [x] Manual local verification confirms the `lark-cli drive +preview` and `lark-cli drive +cover` commands work as expected

Additional verification:
- `make unit-test`
- `go test ./shortcuts/drive -run 'TestDrivePreview|TestDriveCover|TestShortcutsIncludesExpectedCommands'`
- `go test ./tests/cli_e2e/drive -run 'TestDrive(Preview|Cover)DryRun'`
- `go test ./tests/cli_e2e/drive -run 'TestDrive_PreviewAndCoverWorkflow'`

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Added Drive +preview and +cover shortcuts to list/download preview artifacts and cover presets (type/spec selection, optional version, dry-run support, output path, and collision policies).

* **Bug Fixes / Behavior**
  * Stronger validation with clearer errors (404→failed-precondition for unavailable specs/types, hints to re-run list-only). Improved preview candidate normalization, extension inference, masked progress output, and safe output collision handling (error/overwrite/rename).

* **Documentation**
  * New user guides and reference pages with examples, parameters, behaviors, and permission notes.

* **Tests**
  * Extensive unit and e2e tests for list-only, download, dry-run, and workflow scenarios.


wittam-01 (dismissed):

shortcuts/drive/drive_cover.go
```diff
@@ -0,0 +1,122 @@
...
+// wrapDriveCoverDownloadError reclassifies preview_download HTTP 404 responses
+// on the +cover path as a failed precondition on --spec, because the Drive
+// shortcut contract documents 404 as "this file has no artifact for that cover
+// preset" rather than a transient transport failure.
+func wrapDriveCoverDownloadError(err error, requestedSpec string) error {
+	if err == nil {
+		return nil
+	}
+	problem, ok := errs.ProblemOf(err)
+	if !ok || problem.Code != http.StatusNotFound {
```
fangshuyu-768, line 109: This looks too broad for a blanket 404 -> `--spec` failed-precondition rewrite. `preview_download` can also return 404 for a bad `file-token`, a missing/invalid `version`, or a file that is no longer resolvable for the caller. Since this path does not inspect the backend error payload before rewriting the subtype/message, those cases would now be misreported as "the requested spec is not available for this file", which can send users in the wrong direction. Can we narrow this reclassification to the specific backend error shape that really means "no artifact for this cover preset"?

shortcuts/drive/drive_preview_common.go
```diff
@@ -0,0 +1,813 @@
...
+		}
+	}
+
+	var firstAliasMatch drivePreviewCandidate
+	hasAliasMatch := false
+	for _, candidate := range candidates {
+		if !slices.Contains(previewAliasesForCandidate(candidate), requested) {
+			continue
+		}
+		if candidate.Downloadable {
```
fangshuyu-768, line 321: The user-facing contract here says `--type image` is a supported selector, but the actual selection logic is response-order dependent. `image` fans out to multiple concrete preview types (`png`, `jpg`, `png_lin`, `jpg_lin`, `split_png`), and `selectDrivePreviewCandidate()` returns the first downloadable alias match it sees. That means the same `drive +preview --type image` invocation can silently change output format/extension depending on backend candidate ordering. For agent and script callers this is pretty hard to reason about. Would it be safer to require an exact type from `--list-only`, or define an explicit ranking among image candidates instead of trusting response order?

zhoujunteng-max: In practice, the API does not return duplicate entries for the same concrete preview type. In other words, a specific concrete type maps to at most one candidate in preview_results.

shortcuts/drive/drive_cover.go
zhoujunteng-max: That makes sense. I kept the current 404 classification logic, but refined the user-facing guidance. The updated hint now explicitly calls out both possibilities: either no artifact exists for the requested spec, or the provided file token / version is invalid.

fangshuyu-768 (approved):

PR #1299, opened 2026-06-05
docs(skills): prefer raw Lark resource URLs

### LINKED ISSUE #1298: Improve skill guidance for writing Lark resource URLs
## Problem
When agents write Feishu/Lark document, wiki, sheet, or Base links into document content, they often wrap the URLs as Markdown links such as `[title](url)`. For Feishu/Lark Drive-native resources, raw URLs are usually a better default because the editor can preserve native card, preview, or mention conversion opportunities.

The current skills did not state that rule clearly, so agents had no reason to distinguish:
- Drive-native resource URLs such as Docx, Sheets, Base/bitable, Wiki, Drive files/folders, Slides, and Minutes.
- External URLs where ordinary URL or explicit display-text links may be fine.
- Structured API fields that require `rich_text` or link objects.

## Proposed Change
Add a shared skill rule that raw URLs are the default for Feishu/Lark Drive-native resources in rich editor/content surfaces, with explicit exceptions for custom display text, external URLs, non-native share/download/callback URLs, and structured link/rich_text APIs.

Then reference that shared rule from the Docx, Sheets, and Base write paths where agents construct Markdown, XML, rich_text cells, or Base text/description values.

## Acceptance
- The full rule has one shared owner instead of being copied into every skill.
- Docx Markdown/XML guidance tells agents not to default to `[title](url)` for Feishu/Lark resource URLs.
- Sheets and Base guidance preserve legitimate rich link behavior for custom display text while defaulting Drive-native resource URLs to raw strings.
- Wiki coverage is explicit through shared guidance or a local pointer if a Wiki rich-content write path exists.
- Local skill/doc checks pass.

bagaking:
## Summary
This PR improves skill guidance for writing Feishu/Lark resource links into rich editor surfaces. It adds a shared rule that cloud-space native resource URLs should default to raw URL content, then points Docx, Sheets, and Base write paths to that rule so agents do not unnecessarily wrap these links as Markdown or custom-display hyperlinks.

Wiki resource URLs are covered through the shared `lark-shared` rule, which the Wiki skill already requires agents to read before use.

## Why
Agents often fall back to Markdown link syntax in Markdown-capable contexts. For Feishu/Lark native resources, writing `[title](url)` or custom hyperlink markup can prevent the editor from converting the raw resource URL into its native card, preview, or mention-like representation.

The intended contract is now explicit:

- Native Feishu/Lark resource URL, no custom label requested: write raw URL content.
- User asks for custom display text: use Markdown/link syntax when the target surface supports it.
- External URL without display text: it may remain a raw URL string.
- External URL that needs display text or a preview card: use Markdown, XML preview, or structured link forms as required by the target API.
- API fields that explicitly require structured `link` / `rich_text` objects: keep using that structured representation.

## Changes
- Add a shared `lark-shared` rule for Feishu/Lark cloud-space native resource links.
- Keep raw URL content as the default for Docx, Sheets, Base/bitable, Wiki, Drive files/folders, Slides, Minutes, and similar native resource URLs.
- Tighten the exceptions so external URLs only force link markup when display text or preview behavior is requested.
- Reference the shared rule from Docx Markdown/XML authoring guidance, Sheets cell-writing guidance, and Base text/URL/description guidance.

## Prompt-resolution evaluation
A read-only link-format evaluation compared the baseline skill docs at `8c3cba1` with this PR at `973f026`. It used 12 agent-visible contract surfaces, 16 representative prompt-resolution cases, and two model-family runs.

| Metric | Baseline `8c3cba1` | This PR `973f026` |
|---|---:|---:|
| Agent-visible contract coverage | 0/12 (0.0%) | 12/12 (100.0%) |
| Codex / GPT-5.5 prompt success | 16/16 (100.0%) | 16/16 (100.0%) |
| Codex / GPT-5.5 native no-label success | 8/8 (100.0%) | 8/8 (100.0%) |
| Claude / Claude Opus 4.8 prompt success | 11/16 (68.8%) | 16/16 (100.0%) |
| Claude / Claude Opus 4.8 native no-label success | 4/8 (50.0%) | 8/8 (100.0%) |

Interpretation: stronger agents could already infer many cases from the baseline, but the contract was not visible in the docs. The new rule makes the behavior explicit and materially improves the ambiguous baseline cases in the Claude / Claude Opus 4.8 run, especially native Feishu/Lark resource URLs without custom labels.

## Validation
- [x] Unit tests pass: `GOMAXPROCS=4 GOFLAGS='-p=1' make unit-test`
- [x] Manual local verification confirms the skill guidance flow and prompt-resolution evaluation output.
- `node scripts/skill-format-check/index.js`
- `bash scripts/check-skill-wire-vocab.sh`
- `gofmt -l .`
- `go mod tidy && git diff --exit-code -- go.mod go.sum`
- `GOMAXPROCS=4 GOFLAGS='-p=1' go vet ./...`
- `GOMAXPROCS=4 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run --new-from-rev=upstream/main --concurrency=1`

## Related Issues
Closes #1298


skills/lark-base/references/lark-base-cell-value.md
```diff
@@ -18,7 +18,9 @@
 
 ### 2.1 text / phone / url
 
-用字符串。URL 字段也传 URL 字符串；普通文本里可以保留 Markdown 风格链接文本，平台会按字段类型处理。
+用字符串。URL 字段也传 URL 字符串；外部链接可按 URL 字段的普通字符串行为写入。
+
+飞书/Lark Drive 原生资源链接写入 text / url 字段时，遵循 [`lark-shared`](../../lark-shared/SKILL.md) 的「飞书资源链接写入」规则：默认直接写原始 URL，只有共享规则中的例外场景才保留 Markdown 风格链接文本或自定义显示文本。
```
fangshuyu-768: This wording still collapses `text` and `url` into one rule, but they do not have the same expressive power. A `text` field can legitimately carry Markdown link text when the caller explicitly wants custom display text; a `url` field still fundamentally wants a raw URL string. As written, “only in exception cases keep Markdown-style link text or custom display text” reads like it also applies to `url` fields, which can steer agents into writing `[label](https://...)` into a URL field instead of a plain URL. I think it would be safer to split the guidance here: keep the shared-rule reference for `text`, but say explicitly that `url` fields should remain raw URL strings even in the custom-label cases.

skills/lark-doc/references/lark-doc-xml.md
```diff
@@ -32,6 +32,8 @@ p, h1-h9, ul, ol, li, table, thead, tbody, tr, th, td, blockquote, pre, code, hr
 | `<button>` | 操作按钮 | `background-color`、`src`，必须包含 `action=OpenLink\|DuplicatePage\|FollowPage` |
 | `<time>` | 提醒 | 必包含 `expire-time`、`notify-time`（毫秒时间戳）、`should-notify=true\|false` |
 
+> 飞书/Lark 云空间原生资源链接写入遵循 [`lark-shared`](../../lark-shared/SKILL.md) 的共享规则。Docx、Sheets、Base/bitable、Wiki、云空间文件等资源 URL 默认直接写入原始 URL，保留编辑器转换为原生卡片/预览的机会。仅在共享规则中的例外场景（用户明确要求自定义显示文本、外部 URL 需要预览、非云空间原生资源，或需要结构化 link 元素）时使用 `<a>` / `<a type="url-preview">`。
```
fangshuyu-768: I think this over-rotates the XML guidance toward generic hyperlinks. In Docx XML, “don’t default to raw URL” does not necessarily mean “use `<a>` / `<a type="url-preview">`” — this file already defines stronger native resource forms like `<cite type="doc">`, `<sheet>`, and `<bitable>`. For many Lark-native resources, those richer XML constructs are a better fit than a generic hyperlink, and this paragraph currently does not call that out. Would it make sense to add an explicit note that the shared raw-URL rule is only the default for plain text surfaces, and that when the caller is intentionally constructing Docx XML, native resource tags such as `<cite type="doc">`, `<sheet>`, or `<bitable>` should still take precedence over generic `<a>` links when they better express the intent?

skills/lark-doc/SKILL.md
```diff
@@ -30,6 +30,7 @@ lark-cli docs +update --api-version v2 --doc "文档URL或token" --command appen
 > **格式选择规则（全局）：**
 > - **创建 / 导入场景**（`docs +create`，或 `docs +update --command append/overwrite` 的整段写入）：XML 和 Markdown 都可以。用户提供 `.md` 本地文件、或明确说"导入 Markdown"时，直接用 Markdown；否则默认 XML（可用 callout、grid、checkbox 等富 block）。
 > - **精准编辑场景**（`docs +update` 的 `str_replace` / `block_insert_after` / `block_replace` / `block_delete` / `block_move_after` 等局部精修指令）：优先使用 XML（`--doc-format xml`，即默认值）。XML 能稳定表达 block 结构和样式，局部精修更可控；不要因为 Markdown 更简单就自行切换。
+> - **飞书/Lark 云空间原生资源链接**：遵循 [`lark-shared`](../lark-shared/SKILL.md) 的「飞书资源链接写入」规则。写入 Docx、Sheets、Base/bitable、Wiki、云空间文件等资源 URL 时，默认直接写原始 URL，保留编辑器转成原生卡片/预览的机会；只有共享规则中的例外场景（用户明确要求自定义显示文本、外部/非飞书 URL 需要展示文本或预览、非云空间原生资源、API/语法要求结构化 link 元素）才使用 Markdown/HTML/XML 超链接。
```
fangshuyu-768: I think this is starting to drift away from the “one shared owner” goal from #1298. `lark-shared` already defines the full rule and its exception list, but this paragraph re-expands most of that contract again in the Doc skill, and the same pattern is repeated in `lark-doc-md.md` and `lark-doc-xml.md`. That makes future maintenance harder because a change to the shared rule now has to be synchronized across multiple Doc-specific copies. Would it be safer to keep only a short local pointer here (for example: “Docx link writing follows the shared rule; in XML, native tags such as `<cite>/<sheet>/<bitable>` still take precedence when appropriate”) and leave the full exception list owned by `lark-shared`?

skills/lark-base/references/lark-base-field-create.md
```diff
@@ -50,7 +50,7 @@ POST /open-apis/base/v3/bases/:base_token/tables/:table_id/fields
 
 - `--json` 必须是 **JSON 对象**，顶层直接传字段定义，不要再套一层。
 - 顶层最少包含：`name`、`type`。
-- 所有字段类型都支持可选 `description`；支持纯文本，也支持 Markdown 链接，如 `协作约定可参考[团队字段约定](https://example.com/field-spec)`。
+- 所有字段类型都支持可选 `description`；支持纯文本，也支持 Markdown 链接，如 `协作约定可参考[团队字段约定](https://example.com/field-spec)`。写入飞书/Lark Drive 原生资源链接时，遵循 [`lark-shared`](../../lark-shared/SKILL.md) 的「飞书资源链接写入」规则，默认写原始 URL；只有用户明确要求自定义显示文本，或外部链接需要展示文本时才用 Markdown 链接。
```
fangshuyu-768: This has the same layering issue on the Base side: the shared rule already lives in `lark-shared`, but here we are re-stating the local default and exceptions again in a command-specific reference, and the same wording is repeated in several other Base write-path docs in this PR. I agree with adding a local reminder, but at this point it is more than a pointer and starts to create multiple owners for the same contract. Would it be possible to keep the full rule only in `lark-shared`, and reduce Base-side copies to the shortest possible scene-specific reminder (for example: “description/text fields follow the shared rule; URL fields still want raw URL strings”)?

bagaking:
@fangshuyu-768 Thanks again for the review. I’ve addressed all feedback in the latest commit, including removing the XML-specific guidance and keeping the full link-writing contract owned by `lark-shared`. The remaining CI workflows are awaiting maintainer approval.

Could you please take another look and approve the workflows when convenient? Please let me know if anything else is needed before merging.

fangshuyu-768:
Closing this PR because it substantially overlaps with #1446, which covers the same guidance: preserving raw Lark/Feishu resource URLs instead of wrapping them as Markdown links across the relevant Skills. We will keep #1446 as the canonical PR because it has the clearest and most complete scope. If this PR contains any wording, examples, or tests that are not yet covered there, please carry those improvements over to #1446. Thank you for the contribution.

PR #1355, opened 2026-06-09
feat(sheets): typed table I/O & error contract, workbook import/export, skill refresh

xiongyuanwen-byted:
## Summary
Lands the accumulated `feat/lark-sheets-develop` work into `main` (337 commits relative to local main, 57 truly branch-only). The thrust is **typed, type-faithful batch I/O for Lark Sheets** plus a domain-wide migration to the typed error contract, with broad `lark-sheets` skill/doc refresh from the upstream spec.

### New shortcuts & capabilities (feat)
- **Typed table I/O**: `+table-put` / `+table-get` with column type & number-format preservation (`lark_sheet_table_io.go`).
- **Typed write protocol**: pandas-DataFrame-shaped `--sheets` JSON accepted by `+table-put` / `+workbook-create` (pandas DataFrame ↔ payload helper at `skills/lark-sheets/scripts/sheets_df.py`).
- **One-step typed write with styling**: `+table-put --styles` (cell_styles / cell_merges / row_sizes / col_sizes).
- **`+workbook-create` rework**: typed `--sheets` + `--styles` (deep-merge cell/border styles, manual border validation); old `--headers` / `--header-style` removed (subsumed).
- **Workbook import/export**: `+workbook-import` wraps the drive import core; `+workbook-export` refactored to reuse the drive export core (inherits resume/cancel/sanitize/status labels).
- **Gridline shortcuts**: `+sheet-show-gridline` / `+sheet-hide-gridline`.
- **Wiki URLs**: `--url` / `--spreadsheet-token` accept `/wiki/<node_token>` (resolved at Execute, Validate/DryRun stays network-free).
- **Chart**: `aggregateType` gains `counta` (non-empty count incl. text).
- **Error hygiene & ergonomics**: `--token` alias for `--spreadsheet-token`; write-op error messages carry actionable hints (`--print-schema`, stdin fallback); stdin single-input constraint surfaced in flag help.
- **Style field aliases**: `styles` accepts `halign` / `valign` / `horizontal_align` / `vertical_align` as aliases for the canonical alignment fields.

### Notable removal
- **Arrow IPC `--dataframe` / `--dataframe-out` channel + `apache/arrow/go/v17` dep**: an earlier revision of this branch added an Arrow IPC binary input/output channel to `+table-put` / `+table-get` / `+workbook-create`, plus a CI `go-licenses --ignore github.com/apache/arrow/go/v17` exemption. It was removed before merge (commit `9cd27855`): the JSON `--sheets` path already covers everything pandas users need, and the column-major / zero-copy benefits collapse the moment the CLI re-encodes into the row-oriented sheets OpenAPI JSON body. Pandas users keep the round-trip via `skills/lark-sheets/scripts/sheets_df.py` over the typed JSON path. Pre-removal state preserved on branch `feat/sheets-arrow-stash`.

### Robustness & typed error contract (refactor / fix)
- **Sheets domain migrated to the typed error contract**: `common.FlagErrorf` → `common.ValidationErrorf`; `+table-put` partial-success ported off legacy `output.ExitError` to `runtime.OutPartialFailure`; file-I/O and tool-call errors classified.
- **Unified partial-state contract for `+workbook-create`**: when the spreadsheet POST succeeds but the follow-up initial fill fails, the result now emits the same `OutPartialFailure` envelope shape as `+table-put`'s partial write (ok:false stdout envelope with `spreadsheet_token` + `reason` + `hint` + structured `cause`), instead of the previous `failed_precondition` stderr envelope — so agents see one consistent "side effect landed but follow-up didn't" shape across the domain.
- `+table-get` now defaults to the full used range (probed over the physical grid) instead of the A1 current_region — fixes silent truncation past internal blank rows; append mode's probe gets the same fix.
- Conditional-format attrs validated by `rule_type` (cross-field), with a schema-drift guard test — closes a class of dirty payloads that crashed the frontend on snapshot deserialization.
- `+pivot-create`: `--target-position` and `--range` are mutually exclusive (both map to `properties.range`).
- `set_cell_range` `mention_type` constrained to the proto MENTION_FILE_TYPE enum.
- `+cells-get` / `+csv-get` `--max-chars` default raised 200000 → 500000.

### Skill / docs (synced from sheet-skill-spec)
- `set+H` quoting & guidance generalized; cheat-sheet adds hallucinated commands; `+workbook-import` routing.
- `+csv-put` formula-with-comma example & RFC-4180 escaping; structured-write preference surfaced at the decision point.
- `--max-chars` exposed as a visible flag with "spill to file" guidance for large reads.
- Cell-image vs float-image routing clarified by record-binding.
- `read-data` cross-checks `+sheet-info` for hidden rows/cols and `+workbook-info` for physical row_count.
- Drops over-restrictive caveats ("Feishu sheets only" / "not for local Excel") now that import/export close the loop.
- `lark-sheets` skill version bumped to 3.0.0.

## Test Plan
- [x] Unit tests pass — `go test -race ./cmd/... ./internal/... ./shortcuts/... ./extension/...`
- [x] `go build ./...`, `golangci-lint run --new-from-rev=origin/main` (0 issues), `gofmt` clean
- [x] Dry-run E2E for `+table-put --styles` (`tests/cli_e2e/sheets/sheets_table_put_dryrun_test.go`)
- [x] Live E2E for wiki-URL resolution (`+workbook-info`, `+cells-get`, `+csv-get`, `+sheet-create`, `+sheet-rename`, `+csv-put`)
- [x] Dry-run + live E2E for `+sheet-{show,hide}-gridline`, `+workbook-import`, `+workbook-export`
- [ ] Full e2e suite (`tests/cli_e2e`) — not run locally (requires live credentials)

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * Typed table read/write: +table-get and +table-put.
  * Sheet gridline toggles: +sheet-show-gridline and +sheet-hide-gridline.
  * Workbook import shortcut and typed --sheets support for workbook creation.

* **Enhancements**
  * Typed --sheets one-step workbook create+write; workbook export/import have clearer dry-run/execute/resume behaviors.
  * CSV input now treats =-prefixed cells as formulas; csv-get row-prefix handling clarified; empty output-dir normalizes to cwd.

* **Removals**
  * Removed +csv-get --rows-json (use +table-get).

* **Documentation**
  * Added financial-modeling standards and expanded sheets references.


fangshuyu-768 (commented):
Reviewed the latest head again and left inline notes for the remaining behavior/data-safety issues I found.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1615 @@
...
+			// "before row N+1" as out-of-range). Inserting before the last
+			// existing row pushes that row down and effectively appends
+			// `count` blank rows — the data writes that follow will overwrite
+			// the existing rows from row 1, so the placement of the inserted
+			// blanks doesn't matter as long as the total dimension grows.
+			input := map[string]interface{}{
+				"excel_id":  token,
+				"sheet_id":  sheetID,
+				"operation": "insert",
+				"position":  strconv.Itoa(curRows),
```
fangshuyu-768: This growth path can move existing boundary data before the write happens. For example, when the sheet currently has data on the last physical row and an append/write needs row `curRows + 1`, inserting at `curRows` pushes that existing last row down into the area that the following `set_cell_range` writes to. The same issue exists for columns below. That means a capacity grow can silently overwrite user data, and `allow_overwrite=false` cannot protect it because the structural insert has already mutated the sheet before the write collision check. Can we either use a true grid-resize operation that does not shift cells, or fail/handle the case where the boundary row/column contains data? This should also have a regression test where existing data sits exactly on the default last row/column.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -628,362 +678,909 @@ var WorkbookCreate = common.Shortcut{
...
+func workbookCreateStyleDimensions(styles *workbookCreateStylePayload, baseCol, baseRow int) (rows, cols int) {
+	if styles == nil {
+		return 0, 0
 	}
-	target := outPath
-	if info, statErr := runtime.FileIO().Stat(outPath); statErr == nil && info.IsDir() {
-		name := strings.TrimSpace(preferredName)
-		if name == "" {
-			name = client.ResolveFilename(apiResp)
+	for _, op := range styles.CellStyles {
```
fangshuyu-768: `workbookCreateStyleDimensions` only accounts for `CellStyles`, so style-only operations such as `cell_merges`, `row_sizes`, or `col_sizes` do not create any write payload extent. In `+workbook-create`, `Execute` only calls `writeTypedSheets` when `payload != nil`, so a command like `+workbook-create --styles '{"styles":[{"name":"Sheet1","cell_merges":[{"range":"A1:B1"}]}]}'` creates the workbook but silently drops the requested merge. Please include merges and row/column resize ranges when deriving the required extent, or apply visual operations even when there are no value/cell-style cells to write.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1615 @@
...
+	columnNames := make([]interface{}, ncols)
+	colTypes := make([]string, ncols)
+	dtypes := make(map[string]interface{}, ncols)
+	formats := map[string]interface{}{}
+	for c := 0; c < ncols; c++ {
+		typ, format := inferColumnType(dataRows, c)
+		colTypes[c] = typ
+		name := tableGetColumnName(headerRow, c, noHeader)
+		columnNames[c] = name
+		dtypes[name] = typeToDtype(typ)
```
fangshuyu-768, line 1280: Duplicate header cells make the `+table-get` output impossible to round-trip through `+table-put`. This loop emits duplicate entries in `columns` and also collapses `dtypes[name]` to a single map key, while the writer later rejects duplicate column names. A real sheet with headers like `amount, amount` will therefore export successfully but fail on import. Consider deduplicating generated column names on read, or returning a clear validation error that asks callers to use `--no-header` when headers are not unique.

shortcuts/sheets/lark_sheet_dataframe.go
```diff
@@ -0,0 +1,554 @@
...
+// one-shot (one command per process). Tests reset by setting it back to nil.
+var dataframeStdinCache []byte
+
+// readDataframeBytes resolves --dataframe to raw binary. A literal `@` prefix
+// is tolerated for symmetry with --sheets (`@/tmp/x.arrow` and `/tmp/x.arrow`
+// both work). `-` reads stdin verbatim — cached on first call so Validate /
+// Execute / DryRun all see the same bytes. Bytes are returned untouched: no
+// TrimSpace, no BOM strip — both would corrupt an Arrow IPC stream.
+func readDataframeBytes(rctx *common.RuntimeContext, raw string) ([]byte, error) {
+	if raw == "-" {
```
fangshuyu-768: Because `--dataframe` intentionally bypasses the common text `Input` resolver, `--dataframe -` is not counted by the existing single-stdin guard. As a result, commands such as `+table-put --dataframe - --styles -` or `+workbook-create --dataframe - --styles -` are accepted until one flag consumes stdin and the other sees an empty stream or malformed payload. Please add explicit validation for `--dataframe -` combined with any other stdin-backed flag, or model this as a binary input source in the common resolver so the existing guard still applies.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1615 @@
...
+	raw := strings.TrimSpace(runtime.Str("sheets"))
+	if raw == "" {
+		return nil, common.ValidationErrorf("--sheets is required")
+	}
+	dec := json.NewDecoder(strings.NewReader(raw))
+	dec.UseNumber()
+	var wire struct {
+		Sheets []tableSheetIn `json:"sheets"`
+	}
+	if err := dec.Decode(&wire); err != nil {
```
fangshuyu-768, line 258: This `json.Decoder` call accepts trailing non-whitespace after the first JSON value because we never check for EOF after `Decode`. That makes `--sheets '{...} trailing'` look valid, unlike the other JSON flags that use `json.Unmarshal` and reject extra bytes. The same pattern appears in `+workbook-create --values`. Please either switch these paths to `json.Unmarshal` while preserving `UseNumber`, or perform a second decode/EOF check so generated malformed payloads are surfaced.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -628,362 +678,909 @@ var WorkbookCreate = common.Shortcut{
...
+		// /wiki/ URL carries a node_token that needs the get_node step only
+		// Execute may take, so re-resolve the token here.
+		if p.Token, err = resolveSpreadsheetTokenExec(runtime); err != nil {
+			return err
+		}
+		applyWorkbookOutputPath(&p, runtime.FileIO(), runtime.Str("output-path"))
+		return drive.RunExport(ctx, runtime, p)
+	},
+	Tips: []string{
+		"Polls for a bounded window; if the export is still running it returns a resume reference instead of blocking. Pass --output-path to download the file once ready (omit it to only create the export task and get the file token back).",
```
fangshuyu-768, line 1594: From a CLI user's perspective, this new `sheets +workbook-export` overlaps heavily with the existing `drive +export --doc-type sheet`, but the two entry points expose different flags and defaults. `drive +export` uses `--token`, `--doc-type sheet`, `--sub-id`, `--output-dir`, `--file-name`, and `--overwrite`, and it defaults to downloading into the current directory. This wrapper uses `--url` / `--spreadsheet-token`, `--sheet-id`, and a single `--output-path`, and omitting `--output-path` means "create/poll the export task but do not download". That difference is subtle, especially because both commands ultimately reuse the same Drive export core. I think this will confuse users who discover both commands: the same spreadsheet export can either download by default or only return a file token depending on which namespace they choose. Can we either align the default behavior with `drive +export` (download by default, with an explicit no-download/export-task-only mode), or make the difference very explicit in the command help/dry-run output? It would also help to call out that this is a sheet-specialized wrapper around Drive export and map `--sheet-id` to Drive's `--sub-id` terminology for CSV exports.

fangshuyu-768 (commented):
Added one more export-focused CLI UX review note.

shortcuts/sheets/lark_sheet_table_io.go
xiongyuanwen-byted: Fixed in d2517180. The backend `set_cell_range` now auto-grows the sub-sheet to fit the write, so I dropped the manual `ensureSheetCapacity` helper (and its `modify_sheet_structure` dim-insert call) entirely instead of trying to make it safe — this closes the data-safety hole you flagged at the root (no structural mutation runs before the write-collision check anymore).

Verified end-to-end against a real spreadsheet: `+table-put` writing 300×25 into a fresh Sheet1 (default 200×20) succeeds in a single `set_cell_range` and the sheet ends up 301×25.

shortcuts/sheets/lark_sheet_workbook.go
xiongyuanwen-byted: Fixed in 1aa3305f. Two changes:

1. `workbookCreateStyleDimensions` now folds `cell_merges` / `row_sizes` / `col_sizes` ranges into the write-extent computation, so a payload with non-cell_styles ops contributes to the rectangle.
2. Pure dimension-only payloads (e.g. only `row_sizes` / `col_sizes`) can never expand a cell rectangle, so I also added an Execute fast-path: when `payload == nil` but a styles item is present, look up the default sheet and call `applyWorkbookCreateVisualOps` directly. The dry-run mirrors this so the preview shows the visual ops.

End-to-end verified against a real spreadsheet — your exact repro (`+workbook-create --styles '{"styles":[{"name":"Sheet1","cell_merges":[{"range":"A1:B1"}]}]}'`) now produces a sheet with `merged_cells_count: 1`.

shortcuts/sheets/lark_sheet_table_io.go
xiongyuanwen-byted: Fixed in 086876d2. Went with your "clear validation error asking callers to use --no-header" option — `readSheetAsSpec` now detects duplicate column names at read time and returns a typed validation error with the hint:

```
sheet "Sheet1": duplicate header column name "amount" at columns 1 and 2;
this would break the +table-get → +table-put round-trip.
Rename the headers or pass --no-header to read by position (col1/col2/…).
```

`--no-header` mode is exempt because the fallback `col<N>` names are always unique by construction.

End-to-end verified against a real spreadsheet (`amount, amount` header → typed validation error; `--no-header` succeeds and returns positional names).

shortcuts/sheets/lark_sheet_dataframe.go
xiongyuanwen-byted: Fixed in d994c278. Added a `stdinConsumed` marker on `RuntimeContext` that the two paths share explicitly: `common.resolveInputFlags` sets it when an `Input` flag uses `-`, and `readDataframeBytes` both checks it (rejecting a second consumer up front with an actionable @file hint) and sets it (so a later `Input` flag `-` also gets rejected if `--dataframe -` came first). Covered with a unit test (`TestReadDataframeBytes_RejectsSecondStdinConsumer`).

I went with the explicit-check approach rather than modeling `--dataframe` as a binary input source in the common resolver — that would have meant teaching the resolver to skip TrimSpace/BOM-strip for the Arrow IPC bytes, which felt like more surface area than the bug warrants.

shortcuts/sheets/lark_sheet_table_io.go
xiongyuanwen-byted: Fixed in 086876d2. Added a `decoderExpectEOF(dec)` helper and call it after the first `Decode` on both `--sheets` (table-io) and `--values` (workbook-create). Kept `json.Decoder` rather than switching to `Unmarshal` so `UseNumber` semantics stay unchanged for large-int cells; the helper does a second `Decode(&trailing)` and treats anything other than `io.EOF` as trailing data.

End-to-end verified — `--sheets '{...} trailing'` now fails with:
```
--sheets: trailing data after JSON value: invalid character 'a' in literal true (expecting 'u')
```
(and the cause is preserved via `.WithCause(err)` so callers can `errors.As` it).

shortcuts/sheets/lark_sheet_workbook.go
xiongyuanwen-byted: Documented in a07b178b (and the upstream spec change). I went with the **minimum-impact** path:

- `--output-path` description rewritten to lead with "omit = no-download" and explicitly contrast with `drive +export`'s `--output-dir` / `--file-name` / `--overwrite` split + default-cwd-download behavior.
- A new dedicated `+workbook-export` section in `lark-sheets-workbook.md` reference with the four common forms (no-download / file / dir / csv with `--sheet-id`) and the same contrast call-out — so a user who hits this surface knows when to fall back to `drive +export`.

The deeper alignment (adding `--output-dir` / `--file-name` / `--overwrite` flag aliases or changing the default to download-to-cwd) needs a base-table flag add and **breaks the existing `TestWorkbookExport_ExecuteExportOnly` contract**, so I left it as a follow-up. Wanted to keep the no-behavior-change discipline for this PR.

End-to-end verified — omit-output-path returns `ok:true, downloaded:false, file_token`; passing `--output-path ./out.xlsx` writes 17892 bytes locally as expected.

shortcuts/sheets/lark_sheet_table_io_test.go
```diff
@@ -0,0 +1,991 @@
...
+		{"number expects numeric", `{"sheets":[{"name":"S","columns":[{"name":"n","type":"number"}],"rows":[["abc"]]}]}`, "number expects"},
+		{"invalid json", `{not json`, "invalid JSON"},
+	}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			_, err := parseTablePutPayload(stubFlagView{"sheets": tt.json})
+			if err == nil || !strings.Contains(err.Error(), tt.want) {
+				t.Errorf("want error containing %q, got %v", tt.want, err)
+			}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in b276e92f. Did a domain-wide sweep — every error-path assertion in sheets test files now goes through one of two helpers landed in `helpers_test.go`:

```go
requireProblem(t, err, wantCategory, wantSubtype, msgContains) *errs.Problem
requireValidation(t, err, msgContains) *errs.ValidationError  // shorthand for CategoryValidation + SubtypeInvalidArgument
```

~61 assertion sites across 18 sheets test files moved off the `strings.Contains(stdout+stderr+err.Error(), ...)` pattern. The substring is kept as a sanity check against `.Message` / `.Hint` / `.Param` on the returned typed envelope, but category / subtype drift now fails loudly instead of slipping past.

Also did the parallel sweep in `shortcuts/common/` (47a3c1c6) — same helpers, 8 more sites, covering the cases where the production code actually returns a typed envelope. The few sites that came back to plain `*errors.errorString` (cobra-native flag-parser errors, SDK-level `fetchBotInfo` wraps, `rejectPositionalArgs`) were left as substring with a one-line comment explaining why (re-typing them needs a custom `FlagErrorFunc` or wrapping cobra/SDK paths, both out of scope here).

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1148 @@
...
+func tablePutFullRange(s *tableSheetSpec, totalRows int) string {
+	_, col0, row0, err := sheetAnchor(s)
+	if err != nil {
+		return strings.TrimSpace(s.StartCell)
+	}
+	ncols := len(s.Columns)
+	return fmt.Sprintf("%s%d:%s%d",
+		columnIndexToLetter(col0), row0+1,
+		columnIndexToLetter(col0+ncols-1), row0+totalRows)
+}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Fixed in 086876d2. `tablePutFullRange` now returns `""` when `totalRows <= 0` or `len(s.Columns) == 0`, so the `header=false` + `rows=[]` dry-run preview reports an empty range instead of a degenerate `A1:C0`. Added `TestTablePutFullRange_EmptyMatrix` as a regression guard.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1148 @@
...
+	ncols := len(s.Columns)
+
+	// append mode starts below the sheet's existing data; start_cell's row is
+	// ignored (its column is still honored). overwrite mode anchors at row0.
+	baseRow := row0
+	writeHeader := headerOn(s)
+	if s.Mode == "append" {
+		lastRow, err := lastDataRow(ctx, runtime, token, sheetID)
+		if err != nil {
+			return nil, fmt.Errorf("resolving last data row for append: %w", err)
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Addressed across several commits:

- The legacy `output.Errorf` / `output.ExitError` usages in this file are gone entirely (the typed-error-contract migration commit `fc44564` + the `+table-put` partial-success migration `5c7100e` removed them).
- 9ef0d370 fixed seven remaining `ValidationErrorf("...%w", err)` call sites — `fmt.Sprintf` doesn't support `%w`, so they were rendering causes as literal `%!w(*fmt.wrapError=&{...})` and dropping the cause from the typed-error chain. Switched to `%v` for clean rendering and `.WithCause(err)` to preserve the chain.
- The `fmt.Errorf` calls that remain are genuine intermediate errors that the command layer rewraps into typed validation/internal errors before reaching the dispatcher; each carries a `//nolint:forbidigo` with a one-line reason consistent with the guideline.

End-to-end verified — a `buildTypedCell` type-mismatch error now renders as `--sheets[0] "Sheet1": row 0 column "d": date expects an ISO yyyy-mm-dd string, got a number` (clean message, typed envelope, cause preserved).

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -0,0 +1,1148 @@
...
+
+// tableGetTargets resolves which sheets +table-get reads: the one named by
+// --sheet-id / --sheet-name, or every sheet (in workbook order) when neither is
+// given.
+func tableGetTargets(ctx context.Context, runtime *common.RuntimeContext, token string) ([]tableGetSheet, error) {
+	id := strings.TrimSpace(runtime.Str("sheet-id"))
+	name := strings.TrimSpace(runtime.Str("sheet-name"))
+	if id != "" {
+		return []tableGetSheet{{id: id}}, nil
+	}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Fixed in 086876d2. Both fallback paths in `tableGetTargets` (structure read failure and selector-mismatch on `--sheet-id`) now use the id as the name so the output spec is never nameless:

```go
return []tableGetSheet{{id: id, name: id}}, nil
```

This unblocks the `+table-get → +table-put` round-trip when only `--sheet-id` was given and the structure miss. Covered by `TestTableGet_SheetIDFallbackBackfillsName` (structure has a different sheet → fallback triggers → asserts `name == testSheetID`).

xiongyuanwen-byted:
## Code review round-up

All 10 inline review comments now have point-by-point replies on the diff. Summarizing here for the two CodeRabbit "outside diff range" notes and the overall picture:

### Outside-diff CodeRabbit notes

- **`data/flag-schemas.json` — `+batch-update.operations.shortcut` enum missing gridline shortcuts** → fixed in e9c4b1e1 (CLI) + upstream sheet-skill-spec `7584e57`. Verified end-to-end: `+batch-update --operations '[{"shortcut":"+sheet-hide-gridline","input":{"sheet_id":"…"}}]'` now passes schema validation and the backend returns `succeeded: 1, failed: 0`.
- **`lark_sheet_workbook_test.go:431-450` — typed metadata assertions** → covered by the domain-wide sweep in b276e92f. The workbook test now uses `requireProblem(t, err, errs.CategoryConfirmation, errs.SubtypeConfirmationRequired, "")` for the high-risk confirmation gate, and `requireValidation` for the other negative-path cases.

### Sites intentionally left as substring (each carries a one-line comment in code)

- `lark_sheet_dataframe_test.go` ×3 — `decodeArrowToSheet` returns intermediate `fmt.Errorf` per its `//nolint:forbidigo` reason "command layer wraps it"; the unit tests assert the unwrapped intermediate directly.
- `lark_sheet_search_replace_test.go` ×1 — cobra's native `ValidateRequiredFlags` fires before our `Validate` hook and produces a plain `*errors.errorString`. Re-typing this needs a custom `FlagErrorFunc` (sheets does not install one today; only mail does), which is out of scope here.
- `lark_sheet_workbook_test.go` ×2 — same cobra-native flag-parser errors (mutually-exclusive group + missing required flag).
- `shortcuts/common/runner_botinfo_test.go` ×6 — `BotInfo` / `fetchBotInfo` wrap upstream SDK errors with raw `fmt.Errorf` (the SDK-level message `[99991]` / `403` / `invalid character` shows through). Touching that is a common-infra change unrelated to this PR.
- `shortcuts/common/runner_args_test.go` ×4 — `rejectPositionalArgs` returns raw `fmt.Errorf` to satisfy cobra's `PositionalArgs` contract.

Total: ~70 substring sites moved to typed envelope assertions across 25 test files; the ~16 left are all listed above with a stated reason and a code comment.

### Commits answering the inline comments

| Topic | Commit |
| --- | --- |
| Drop manual capacity grow / data-safety hole | `d2517180` |
| `+workbook-create` style-only ops no longer dropped | `1aa3305f` |
| `+table-get` duplicate header rejected with `--no-header` hint | `086876d2` |
| `--dataframe` stdin guard shared with common Input resolver | `d994c278` |
| `--sheets` / `--values` JSON trailing data rejected | `086876d2` |
| `--sheet-id` fallback uses id as name to keep round-trip intact | `086876d2` |
| `tablePutFullRange` empty-matrix dry-run | `086876d2` |
| `ValidationErrorf %w` → `%v` + `.WithCause(err)` (×7) | `9ef0d370` |
| `+batch-update` enum gridline + upstream spec sync | `e9c4b1e1` |
| Typed-Problem assertions sweep (sheets) | `b276e92f` |
| Typed-Problem assertions sweep (common) | `47a3c1c6` |
| `+workbook-export` doc + drive +export cross-reference | `a07b178b` |

Every behavior fix above was end-to-end verified against a real spreadsheet (PPE) — not just unit tests. The cases live in the per-comment replies.

zhengzhijiej-tech (dismissed):

fangshuyu-768 (approved):

liangshuo-1 (approved):

PR #1357, opened 2026-06-09
feat(sidecar): add client key self-registration API

hGrany:
## Summary

Adds `/_sidecar/keys/register` management-plane endpoint so clients can dynamically register their own per-client HMAC keys without server-side pre-provisioning.

### Problem

In containerized / sandboxed environments (e.g. Coder workspaces, CI runners), the sidecar's `keys/` directory is typically a **read-only host mount**. Operators had to manually create per-client `.key` files for every new sandbox before it could connect — this doesn't scale and blocks fully automated provisioning.

### Solution

Clients generate a random 64-char hex HMAC key locally (in their writable home directory), then register it with the sidecar via a single `POST /_sidecar/keys/register` call signed with the shared `proxy.key`. The server:

1. **Validates** the client ID (safe characters, no path traversal) and key format (64 hex chars, not colliding with proxy key)
2. **Persists** the key to `keys/{client_id}.key` on disk
3. **Hot-loads** it into the in-memory key map — no server restart needed
4. Returns `"registered"` on first call, `"already_registered"` on idempotent re-registration, or `409 Conflict` if the client already has a different key

### Changes

| File | Change |
|------|--------|
| `auth_bridge.go` | Add `keyRegistrar` interface + `handleRegisterKey` handler + `/_sidecar/keys/register` route |
| `handler.go` | Add `validateClientID`, `registerClientKey` (persist + hot-load), `hotLoadClientKey` |
| `main.go` | Replace `validate.SafeInputPath` with `charcheck`-based `validateSidecarPath` that allows absolute server paths; reorder init to resolve circular dependency (handler ↔ authBridge) |
| `handler_test.go` | 4 new tests: registration + idempotency, conflict detection, input validation (empty/reserved/traversal/bad hex/proxy collision), full HTTP round-trip |
| `README.md` | Document self-registration flow, add `client-init.sh` example script, update architecture diagram and end-to-end workflow sections |

### Design decisions

- **Management-plane HMAC** for registration: reuses the existing `proxy.key` signing scheme — no new auth mechanism needed
- **Disk persistence + hot-load**: keys survive sidecar restarts; existing file-scan (`loadClientKeys`) continues to work for pre-provisioned keys
- **Strict validation**: regex-based client ID check, hex format enforcement, proxy key collision guard, path traversal prevention
- **Idempotent**: safe to call on every container startup (init scripts typically run on each boot)

### Testing

- All 4 new unit tests pass
- All existing sidecar demo tests pass (no regressions)
- **Production validated**: deployed and running in a Coder workspace environment with 20+ concurrent client containers for 1+ week

## Test plan

- [x] `go test -tags authsidecar_multi_tenant_demo ./sidecar/server-multi-tenant-demo/ -v` — all tests pass
- [x] `go build -tags authsidecar_multi_tenant_demo ./sidecar/server-multi-tenant-demo/` — builds cleanly
- [x] Production smoke test with real containers registering keys dynamically

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Clients can self-register per-client HMAC keys via a management endpoint; idempotent registration returns clear statuses (200 success/already-registered, 409 conflict).

* **Documentation**
  * README updated with client init scripts, end-to-end workflow, management endpoint details, and clarified multi-tenant dual-key design.

* **Tests**
  * Added tests for registration success, idempotency, conflict handling, and invalid-input cases.

* **Bug Fixes / Validation**
  * Improved input/path validation and runtime key-loading consistency.

### DIFF
--- a/sidecar/server-multi-tenant-demo/README.md
+++ b/sidecar/server-multi-tenant-demo/README.md
@@ -285,7 +285,7 @@ client key works across all instances because each independently verifies it.
 ```bash
 # Register with all organization instances during init
 for PORT in 16384 16385; do
-  call_sidecar POST /_sidecar/keys/register "$REGISTER_BODY" "$SIDECAR_HOST" "$PORT"
+  SIDECAR_PORT="$PORT" call_sidecar POST /_sidecar/keys/register "$REGISTER_BODY"
 done
 
 # Switch to Organization A at runtime
--- a/sidecar/server-multi-tenant-demo/README.md
+++ b/sidecar/server-multi-tenant-demo/README.md
@@ -8,25 +8,38 @@
 Organizations often manage **multiple Lark/Feishu apps** (e.g. one per
 department, one per product line), each with its own `app_id` and `app_secret`.
 These credentials must never be exposed to end-user environments (CI runners,
-developer sandboxes, containerized workspaces). At the same time, when multiple
-users share the same sidecar infrastructure, their Feishu identities must be
-strictly isolated — user A must never accidentally operate as user B.
+developer sandboxes, containers). At the same time, when multiple users share
+the same sidecar infrastructure, their Feishu identities must be strictly
+isolated — user A must never accidentally operate as user B.
+
+Additionally, in containerized or sandboxed deployments the sidecar's `keys/`
+directory is frequently a **read-only host mount**. Pre-creating a `.key` file
+for every new client before it starts is a manual bottleneck that prevents
+fully automated provisioning and elastic scaling.
 
 The single-tenant [server-demo](../server-demo/) solves the credential-hiding
 problem for **one app with one user**. This multi-tenant demo extends it to
 support:
 
-1. **Multiple apps** — run one sidecar instance per app; each instance holds
-   its own `app_id` / `app_secret` and listens on a separate port. Clients
-   choose which app to use by pointing `LARKSUITE_CLI_AUTH_PROXY` to the
-   corresponding port.
+1. **Multiple organizations / apps** — run one sidecar instance per app; each
+   instance holds its own `app_id` / `app_secret` and listens on a separate
+   port. Clients choose which app (organization) to use by pointing
+   `LARKSUITE_CLI_AUTH_PROXY` to the corresponding port.
 2. **Per-client identity isolation** — each client environment gets a unique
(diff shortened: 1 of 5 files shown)


sang-neo03:
@hGrany 
This is a real deployment need, especially in containerized, sandboxed, and CI
  environments where pre-provisioning keys for every client adds operational
  complexity.

  That said, this directory is currently a demo. It is better kept focused on
  the basic sidecar mechanism, such as HMAC verification, per-client isolation,
  OAuth login bridging, and request forwarding. Adding /_sidecar/keys/register
  introduces a new management endpoint, key lifecycle handling, disk
  persistence, hot loading, concurrent creation handling, and conflict
  semantics. These are closer to production deployment concerns and would
  noticeably increase the complexity of the demo.

  I would prefer to keep the demo minimal and limited to the core example logic.
  Key distribution, automated registration, and integration with secret  management systems should be handled by      users according to their own  deployment environment.

  So my suggestion is to close this PR and keep this capability out of the   current demo.

hGrany:
Thanks for the review @sang-neo03 — I understand the concern about keeping the demo minimal and focused.

That said, I'd like to point out that this demo directory is, in practice, the **only implementation** of the multi-tenant sidecar. There's no separate "production" version — everyone deploying multi-tenant setups builds from `server-multi-tenant-demo/`. So the bar for "demo completeness" is effectively the bar for "can I actually use this."

I'd also appreciate your thoughts on #1367, which addresses a more fundamental issue: **the demo silently falls back to `offline_access` when no scope cache exists**, resulting in `scope_count=1`. Users get a "login successful" response but effectively zero useful permissions — document creation, drive access, search, etc. all return `99991679: Permission denied`. This is a silent failure that's very hard to diagnose.

Both PRs came from running the demo in production (Coder workspaces, 90+ containers). The demo's current state requires significant setup knowledge and manual pre-configuration to work correctly. These patches aim to make it more self-contained — closer to **"clone, configure, run"** rather than "clone, read source code, understand scope caching internals, manually create cache files, then run."

If you'd prefer to keep #1357 out, I can accept that — key distribution is indeed deployment-specific. But #1367 is fixing a **correctness issue** in the demo itself: the `offline_access` fallback masks a misconfiguration that should be surfaced as an explicit error.

sang-neo03:
@hGrany Okay, I will review #1367.

PR #1364, opened 2026-06-09
fix(im): continue contact batch resolution on transient API failures

nguyenngothuong:
## Problem

When resolving sender display names in IM messages, a single transient API failure causes **all remaining batches** to be skipped, leaving hundreds of sender names empty.

## Root Cause

`batchResolveByBasicContact()` and `batchResolveUsers()` in `shortcuts/im/convert_lib/helpers.go` use `break` on error â€” stopping the entire loop instead of just skipping the failed batch.

Meanwhile, `batchQueryChatContexts()` in the same codebase already uses a best-effort `continue` pattern.

## Fix

Change `break` â†’ `continue` in both functions so a failed batch only loses its own entries.

| Function | Batch size | Before | After |
|----------|-----------|--------|-------|
| `batchResolveByBasicContact` | 10 | break on error | continue |
| `batchResolveUsers` | 50 | break on error | continue |

## Testing

- Added `TestBatchResolveByBasicContactContinuesOnFailure` â€” first batch fails, second succeeds, verifies partial results
- Added `TestBatchResolveUsersContinuesOnFailure` â€” same pattern for bot identity path
- All existing tests pass (62 tests)
- `go vet` clean
- `gofmt` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Batch name resolution now tolerates transient API failures (continuing to produce partial results), emits warnings on repeated errors, and stops further retries after a small number of consecutive failures; successful responses reset the failure count so processing can resume.

* **Tests**
  * Added tests covering transient-failure handling, circuit-breaker behavior, and reset-on-success scenarios.

### DIFF
--- a/shortcuts/im/convert_lib/helpers.go
+++ b/shortcuts/im/convert_lib/helpers.go
@@ -148,6 +148,13 @@ func ResolveSenderNames(runtime *common.RuntimeContext, messages []map[string]in
 	return nameMap
 }
 
+// maxConsecutiveBatchFailures bounds how many batches in a row may fail before
+// name resolution gives up. A single transient error should not discard the
+// remaining batches, but a run of failures usually signals a systemic problem
+// (expired token, missing scope, downstream outage) where continuing would only
+// fire more doomed requests and add latency, so we break out early.
+const maxConsecutiveBatchFailures = 3
+
 // batchResolveByBasicContact resolves user names via POST /contact/v3/users/basic_batch.
 // This API has lighter permission requirements and works with user identity
 // even when the target user is not in the app's visible range.
@@ -155,6 +162,7 @@ func ResolveSenderNames(runtime *common.RuntimeContext, messages []map[string]in
 // The basic_batch endpoint caps user_ids at 10 per request.
 func batchResolveByBasicContact(runtime *common.RuntimeContext, missingIDs []string, nameMap map[string]string) {
 	const batchSize = 10
+	consecutiveFailures := 0
 	for i := 0; i < len(missingIDs); i += batchSize {
 		end := i + batchSize
 		if end > len(missingIDs) {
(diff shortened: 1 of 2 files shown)


YangJunzhou-01:
Thanks for the fix! The intent makes sense — we shouldn't discard all remaining batches due to a single transient failure.
However, I have a concern: if the failure is systemic (e.g., expired token, insufficient permissions, service outage), then all subsequent batches will inevitably fail as well. In that case, continue causes us to fire N-1 additional doomed requests, adding unnecessary latency and pressure on the downstream service — whereas the original break at least failed fast.
Suggestion: consider adding a consecutive-failure threshold (circuit breaker) to distinguish transient blips from systemic failures:
```go
consecutiveFailures := 0
for _, batch := range batches {
    data, err := runtime.DoAPIJSONTyped(...)
    if err != nil {
        consecutiveFailures++
        if consecutiveFailures >= 3 {
            break // likely systemic, stop early
        }
        continue
    }
    consecutiveFailures = 0
    // ... process data
}
```


﻿
This way we get the best of both worlds: tolerating occasional transient errors while still failing fast when the underlying cause is persistent.
Also, it would be helpful to log a warning on continue (including the error and batch index) for observability.

YangJunzhou-01:
Thanks for the contribution, @nguyenngothuong. The core idea here is sound — switching the transient-failure `break` to `continue` and adding a consecutive-failure circuit breaker (`maxConsecutiveBatchFailures = 3`) preserves the results from other batches during a single hiccup, while still avoiding pointless requests under a systemic failure (expired token, missing scope, downstream outage). We're happy with that part of the logic.

There's one thing we'd like you to adjust before this can be merged:

**Please remove the `warning:` logging written to `runtime.IO().ErrOut`.**

Specifically, these lines (the symmetric pairs in both `batchResolveByBasicContact` and `batchResolveUsers`):

```go
fmt.Fprintf(runtime.IO().ErrOut, "warning: contact_..._resolve_failed: batch %d-%d: %v\n", i, end-1, err)
...
fmt.Fprintf(runtime.IO().ErrOut, "warning: contact_..._resolve_aborted: %d consecutive failures, skipping remaining batches\n", consecutiveFailures)
```

Reasoning:

- **Name resolution is a soft dependency.** When it fails, some user names simply fall back to their IDs — the correctness of the main flow is unaffected. Emitting warnings to `ErrOut` for this "degraded but non-fatal" case interferes with the command's normal output, especially when the output is consumed via a pipe or when repeated batch failures spam these messages and pollute output that should stay clean.
- Since this is explicitly a soft dependency, the right behavior is to degrade silently (keeping the `continue` and circuit-breaker control flow) rather than surfacing internal retry details to the user.

**Please keep the `continue` + circuit-breaker logic and only remove the 4 `fmt.Fprintf(...ErrOut...)` log statements.** If any test assertions depend on that `ErrOut` output, please clean those up as well; the breaker and reset-on-success behavior tests can stay.

We'll re-review once that's done. Thanks!

PR #1366, opened 2026-06-09
fix(im): return partial mget results in messages search instead of all-or-nothing

nguyenngothuong:
## Problem

In `+messages-search`, when the batch mget step fails for any single chunk, **all** message details are discarded and replaced with a bare `message_ids` list. AI agents lose all message content, sender names, and timestamps.

## Root Cause

`batchMGetMessages()` in `shortcuts/im/im_messages_search.go` returns `nil, err` on the first batch failure, and the caller falls back to ID-only output.

With 55 search results split into 2 mget batches (50 + 5), a transient failure on batch 1 loses all 55 messages even though batch 2 would succeed.

## Fix

Change `batchMGetMessages` to best-effort:
- `continue` on batch failure instead of `return nil, err`
- Return partial results (successfully fetched batches)
- When all batches fail â†’ empty slice â†’ existing fallback still applies

Also simplified the function signature from `([]interface{}, error)` to `[]interface{` since errors are no longer propagated.

## Testing

- Added `TestImMessagesSearchMgetPartialResults` â€” 55 IDs in 2 batches, first fails, second succeeds, verifies 5 partial results returned
- All existing tests pass (3/3)
- `go vet` clean
- `gofmt` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IM message search resilience to handle transient API failures gracefully. The search now returns available partial results instead of failing entirely when some message details cannot be retrieved.

### DIFF
--- a/shortcuts/im/im_messages_search.go
+++ b/shortcuts/im/im_messages_search.go
@@ -121,9 +121,8 @@ var ImMessagesSearch = common.Shortcut{
 		}
 
 		// ── Step 2: Batch fetch message details (mget) ──
-		msgItems, err := batchMGetMessages(runtime, messageIds)
-		if err != nil {
-			// Fallback when mget fails: return ID list only
+		msgItems := batchMGetMessages(runtime, messageIds)
+		if len(msgItems) == 0 {
 			outData := map[string]interface{}{
 				"message_ids": messageIds,
 				"total":       len(messageIds),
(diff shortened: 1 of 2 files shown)


YangJunzhou-01:
Thanks for raising this. After discussion, we've decided not to merge this change.

The reason is that this search tool is consumed by Agents that reason over its results. With this change, when some mget batches fail, the failed messages silently disappear — no IDs, no count, no warning — so the Agent sees only a subset while believing it has the full set. That's worse than the current all-or-nothing fallback, which returns the complete ID list and preserves a clear signal that details are incomplete. Silently returning partial data introduces a hard-to-debug failure mode where the Agent reasons over incomplete evidence without knowing it.

We'd prefer to keep the existing all-or-nothing fallback for now, as it maintains the completeness signal for downstream consumers.

Closing this, but feel free to reopen if you have additional context. Really appreciate the effort.

PR #1407, opened 2026-06-11
feat(sheets): resolve @-mention open_id/union_id to user_id

zhengzhijiej-tech:
## Problem

A sheet snapshot stores a user @-mention token as the tenant **user_id** (rich_text segment token, e.g. `5d7g1c33`). But an agent only ever holds an **open_id** — `contact +search-user` never returns a user_id — so @-mentioning a person via `+cells-set` could no longer be written: the model had no way to obtain the required user_id. The legacy open-platform values API used to translate open_id/union_id/email ↔ user_id at its boundary; the newer tool surface (`set_cell_range` / `get_cell_ranges`) passes the token through verbatim, so that translation was lost.

## Fix

Restore the translation at the CLI boundary:

- **Write** (`+cells-set`, and `set_cell_range` sub-ops dispatched via `+batch-update`): `open_id` / `union_id` mention tokens are resolved to `user_id` via `GET /contact/v3/users/batch` before the cells matrix is sent. Resolution is strict — an unresolvable id fails the write so a broken mention is never stored.
- **Read** (`+cells-get`): the stored `user_id` in a mention token is translated back to `open_id`, so callers get the id type they actually speak (consistent with `+search-user`).
- Document mentions (`@file`, `mention_type=1`) and already-resolved `user_id` tokens pass through untouched — fully backward compatible.

## Scope

- New `shortcuts/sheets/mention_resolve.go` (translation core: shape gating, batched contact lookups).
- Wired into `+cells-set` / `+cells-get` / `+batch-update` Execute paths.
- Added `contact:contact.base:readonly` + `contact:user.employee_id:readonly` scopes to the affected shortcuts.

## Verification

- New unit/integration tests in `mention_resolve_test.go` cover both directions, the batch path, strict rejection, user_id passthrough, and document-mention non-conversion.
- `go test ./shortcuts/...` and `go vet` pass.
- Verified end-to-end against a live spreadsheet: writing `ou_…` stores the resolved user_id and renders the person; reading it back returns `ou_…`; an invalid open_id is rejected.

### DIFF
--- a/shortcuts/sheets/lark_sheet_batch_update.go
+++ b/shortcuts/sheets/lark_sheet_batch_update.go
@@ -44,7 +44,9 @@ var BatchUpdate = common.Shortcut{
 	Command:     "+batch-update",
 	Description: "Execute a batch of write shortcuts as a single atomic request (rolls back on failure by default).",
 	Risk:        "high-risk-write",
-	Scopes:      []string{"sheets:spreadsheet:write_only"},
+	// contact.* scopes power @-mention resolution for set_cell_range sub-ops
+	// (open_id / union_id mention tokens → tenant user_id).
+	Scopes:      []string{"sheets:spreadsheet:write_only", "contact:contact.base:readonly", "contact:user.employee_id:readonly"},
 	AuthTypes:   []string{"user", "bot"},
 	HasFormat:   true,
 	Flags:       flagsFor("+batch-update"),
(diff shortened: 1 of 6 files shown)


zhengzhijiej-tech:
Closing: this CLI-layer approach cannot produce a functional @-mention. A working sheet mention requires the internal numeric UserID plus a MentionID (used for notification), neither of which is exposed by the open-platform contact API — contact v3 returns only open_id/union_id/user_id, and a mention token written as the open-platform user_id renders a dead card that does not notify. The resolution must happen server-side in the sheet tool layer, where the user-resolution used by the legacy values API is available. Moving the fix there.

Issue #1404, opened 2026-06-11
Refresh HTTP 400 can clear user auth despite valid refreshExpiresAt


ArthurNie:
## Summary

A refresh attempt can return HTTP 400 from `/open-apis/authen/v2/oauth/token` and leave the local user identity as `missing`, even though `auth status` had previously reported a still-valid `refreshExpiresAt` and `offline_access=true`.

This is different from #738. In this case, a real API call did trigger the refresh path, but refresh failed and the stored user auth became unrecoverable without a new device-flow login.

## Environment

- OS: macOS arm64
- lark-cli: 1.0.48
- Brand: feishu
- Install path: npm global package under `/opt/homebrew/lib/node_modules/@larksuite/cli`
- User auth included `offline_access`

## Sanitized timeline

- `2026-06-11T11:13:49+08:00`: device-flow login succeeded.
- Immediately after login, `auth status --verify` showed user auth ready, token valid, `offline_access=true`, access token expiring around `2026-06-11T13:13:49+08:00`, and refresh token expiring around `2026-06-18T11:13:49+08:00`.
- `2026-06-11T13:14+08:00` to `15:14+08:00`: hourly auth checks saw `tokenStatus=needs_refresh`; the refresh token was still in the `72-168h` remaining bucket.
- `2026-06-11T16:09:56+08:00`: a real user API call (`calendar events ...`) triggered `/open-apis/authen/v2/oauth/token`, which returned HTTP 400.
- After that, `lark-cli auth status --verify` reported the user identity as `missing`, and the only available recovery path was a new device-flow login.

## Sanitized auth log evidence

```text
[lark-cli] auth-response: time=2026-06-11T11:13:49.654245+08:00 path=/open-apis/authen/v2/oauth/token status=200 x-tt-logid=202606111113495AE92A6CCA8006FD8521 cmdline=... auth login ...
[lark-cli] auth-response: time=2026-06-11T16:09:56.417607+08:00 path=/open-apis/authen/v2/oauth/token status=400 x-tt-logid=20260611160956E4DCC67C175F8E260603 cmdline=... calendar events ...
```

## Expected behavior

When refresh fails while local metadata still says the refresh token has days remaining, `lark-cli` should ideally do at least one of the following:

1. Preserve enough local auth state for retry or diagnosis instead of immediately collapsing to `missing`.
2. Return a structured error that distinguishes refresh-token invalidation, server-side revocation, scope/app mismatch, and transient token endpoint failure.
[...]

PR #1416, opened 2026-06-11
auth: refresh user tokens earlier and clarify refresh failures

ArthurNie:
## Summary

Refresh user access tokens earlier and correct identity diagnostics after a successful refresh.

This PR has been rebased and narrowed onto current `main`. The earlier concurrency, refresh-error disposition, token-preservation, and generation-safety changes are now superseded by merged #2135, so the old `uat_client.go` changes were dropped rather than layered onto the hardened refresh flow.

## Changes

- Refresh access tokens within 60 minutes of expiry instead of 5 minutes.
- Document the assumption that normal access-token TTL exceeds the refresh-ahead window.
- After `auth status --verify` refreshes and successfully verifies the user token, report the user identity as `ready` and available.
- Add focused boundary and refresh-plus-verification regression coverage.
- Update existing one-hour "already valid" fixtures to 90 minutes so their intent remains valid under the wider window.

## Test plan

- `go test ./internal/auth ./internal/identitydiag -count=1`
- `go build ./...`
- `git diff --check`

## Related issue

Related to #1404. The destructive/concurrent refresh path is now handled by #2135; this PR retains only the independent proactive-refresh and diagnostic-readiness improvements.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved token refresh timing so tokens are refreshed earlier, reducing the risk of expiration during use.
  - Ensured successfully verified user identities consistently appear as ready and available.
  - Improved diagnostics for expiring tokens by refreshing them before verification.

- **Tests**
  - Added coverage for refresh timing and identity verification scenarios.

### DIFF
--- a/internal/auth/token_store.go
+++ b/internal/auth/token_store.go
@@ -25,7 +25,8 @@ type StoredUAToken struct {
 	GrantedAt        int64  `json:"grantedAt"` // Unix ms
 }
 
-const refreshAheadMs = 5 * 60 * 1000 // 5 minutes
+// Refresh-ahead assumes normal access token TTL is longer than this window.
+const refreshAheadMs = 60 * 60 * 1000 // 60 minutes
 
 var errStoredTokenCorrupt = errors.New("stored token data is corrupt")
 
--- a/internal/auth/token_store_status_test.go
+++ b/internal/auth/token_store_status_test.go
@@ -1,0 +1,25 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package auth
+
+import (
+	"testing"
+	"time"
+)
+
+func TestTokenStatusUsesRefreshAheadWindow(t *testing.T) {
+	now := time.Now().UnixMilli()
+	token := &StoredUAToken{
+		ExpiresAt:        now + (61 * time.Minute).Milliseconds(),
+		RefreshExpiresAt: now + (24 * time.Hour).Milliseconds(),
+	}
+	if got := TokenStatus(token); got != "valid" {
+		t.Fatalf("TokenStatus outside refresh window = %q, want valid", got)
+	}
+
+	token.ExpiresAt = now + (59 * time.Minute).Milliseconds()
+	if got := TokenStatus(token); got != "needs_refresh" {
+		t.Fatalf("TokenStatus inside refresh window = %q, want needs_refresh", got)
+	}
+}
--- a/internal/auth/uat_client_refresh_test.go
+++ b/internal/auth/uat_client_refresh_test.go
@@ -320,7 +320,7 @@ func TestRefreshDoesNotOverwriteNewerGeneration(t *testing.T) {
 			newGeneration := *stored
 			newGeneration.AccessToken = "access-login-new"
 			newGeneration.RefreshToken = "refresh-login-new"
-			newGeneration.ExpiresAt = time.Now().Add(time.Hour).UnixMilli()
+			newGeneration.ExpiresAt = time.Now().Add(90 * time.Minute).UnixMilli()
 			newGeneration.RefreshExpiresAt = time.Now().Add(24 * time.Hour).UnixMilli()
 			var calls atomic.Int32
 			client := scriptedRefreshClient(t, []refreshHTTPTestStep{{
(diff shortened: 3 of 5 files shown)


albertnusouo:
**Overall:** Solid, well-tested change addressing a real pain point — the 5-min window is easy for scheduled/intermittent CLI
  usage to miss. Code compiles, `meta`/`metaOK` are correctly in scope, and the new test meaningfully guards the constant
  (59/61 straddle the boundary and would fail under the old value). A few things worth considering before merge.

  ### Main concern — widening the window enlarges the blast radius of clearing a still-valid token

  In `GetValidAccessToken`, the `needs_refresh` path clears the token and returns `NeedAuthorizationError` on any non-retryable
  refresh failure. With the old 5-min window, the in-hand access token had ≤5 min left, so discarding it was cheap. At 60 min,
  a transient refresh-side hiccup (or an errclass misclassification) can now discard an access token with up to ~59 min of
  remaining validity and force a re-login. This trades availability for earlier refresh, and is strictly more fragile when the
  refresh endpoint is flaky.

  **Suggestion:** on a non-retryable refresh failure, check `now < stored.ExpiresAt` first — if the access token is still
  actually valid, keep using it (warn only) instead of clearing. If immediate invalidation on revoked/reused codes is an
  intentional security policy, please state that explicitly, as it's a deliberate tradeoff.

  ### Implicit assumption — access token TTL must exceed the refresh-ahead window

  This relies on TTL > 60 min (currently `7200s` default). If the server ever issues tokens with TTL ≤ 60 min, `ExpiresAt -
  60min` is always in the past and *every* CLI invocation triggers a refresh. Worth a one-line comment at the constant
  documenting this dependency.

  ### Minor

  - Refresh call frequency roughly doubles (one extra rotation per token lifetime) — negligible at current scale, but a noted
  cost.
  - **Nit:** the PR body mentions running `go test ./internal/errclass`, but no errclass code/tests are touched here
  (regression-only, no issue).

liangshuo-1 (dismissed):

ArthurNie:
Rebased and narrowed this PR onto current `main`.

The refresh concurrency/error-disposition work is now superseded by merged #2135, so I removed the old competing `uat_client.go` implementation. The PR now retains only:

- the 60-minute proactive refresh window, and
- corrected `ready` diagnostics after refresh plus successful server verification.

Validation passed: `go test ./internal/auth ./internal/identitydiag -count=1`, `go build ./...`, and `git diff --check`.

Could a maintainer please review the narrowed scope? The remaining CLA status appears to be repository-side/pending despite the contributor having previously signed it.

ArthurNie:
Rebased and narrowed this PR onto current `main`.

New exact head: `a22002f93c0e2e84bc76e9ede4e9a1703a2d7f9f` (base `755daa4de3ea12785c43a15244ffb8f012122c13`).

The concurrency, refresh-error disposition, token preservation, and generation-safety work is now superseded by merged #2135, so I dropped the old `uat_client.go` changes rather than layering competing behavior onto the hardened refresh flow.

This PR now retains only the independent pieces:
- refresh access tokens within 60 minutes of expiry, with 59/61-minute boundary coverage;
- report user identity as `ready` after a refresh and successful server verification, with an end-to-end diagnostic regression test.

Existing one-hour “valid token” fixtures were adjusted to 90 minutes where necessary so their original intent remains valid under the wider refresh-ahead policy.

Validation on the exact head:
- `go test ./internal/auth ./internal/identitydiag -count=1`;
- `go build ./...`;
- `git diff --check`;
- conflict-marker scan.

Could a maintainer review the narrowed scope? The remaining non-code blocker appears to be the CLA check.

ArthurNie:
Addressed the current review and CLA identity blockers on a refreshed main.

New exact head: `3e57b38e794d22de90dd81d04a1094fd06346c00`  
Base: `da371dc242b6c04823f78f53a537dfd0ab6e6377`

Changes since the prior head:
- the refresh/verify regression now also asserts that the stored access token is `access-new`, directly proving persistence of the refreshed credential;
- the single PR commit is now authored and committed by the GitHub-linked identity `ArthurNie <nlarthur@gmail.com>` rather than the unlinked local Merlin identity.

Final-head validation:
- `go test ./internal/auth ./internal/identitydiag -count=1`: passed;
- `go build ./...`: passed;
- `git diff --check`: passed;
- conflict-marker scan: passed;
- scope remains the same five auth/identity files.

Could CodeRabbit re-review the persisted-token assertion, and could the CLA check be rerun against the corrected commit identity?

Issue #758, opened 2026-05-06
docs +update --command block_move_after / block_copy_insert_after with --src-block-ids fails with 'src_block_ids empty'


baiyu-game:
### Environment

- `lark-cli version 1.0.23`
- Brand: `lark` (international)
- Identity: `user`
- OS: Windows 11
- Node.js: v24.13.1
- Skills: `npx skills add larksuite/cli` (24 skills installed)

### Issue

Both `block_move_after` and `block_copy_insert_after` fail with the same API error when `--src-block-ids` is provided, even though `--dry-run` shows the field is correctly serialized in the request body.

### Reproduction

```bash
# Setup: a doc with two paragraph blocks (anchor and source)
lark-cli docs +create --api-version v2 --doc-format markdown --content $'# Test\n\nAnchor paragraph.\n\nSource paragraph.'
# → fetch with --detail with-ids to get block IDs
# → say anchor=ANCHOR_ID, source=SRC_ID

# Try to move source after anchor:
lark-cli docs +update --api-version v2 --doc <DOC_ID> \
  --command block_move_after \
  --block-id ANCHOR_ID \
  --src-block-ids SRC_ID
```

### Expected

The source block is moved after the anchor block.

### Actual

```json
{
  "ok": false,
  "identity": "user",
  "error": {
    "type": "api_error",
    "code": 2,
    "message": "command 'block_move_after' requires content or src_block_ids but both were empty. Provide at least one.",
    "detail": {
      "log_id": "..."
    }
  }
}
```

### Dry-run output (looks correct)

```bash
lark-cli docs +update --api-version v2 --doc <DOC> --command block_move_after \
  --block-id ANCHOR --src-block-ids SRC --dry-run
```

```json
{
  "api": [
    {
      "desc": "OpenAPI: update document",
      "method": "PUT",
      "url": "/open-apis/docs_ai/v1/documents/<DOC>",
      "body": {
        "block_id": "ANCHOR",
        "command": "block_move_after",
        "format": "xml",
        "revision_id": -1,
        "src_block_ids": "SRC"
      }
    }
  ]
}
```

### Suspected cause

The API likely expects `src_block_ids` as a **JSON array** (e.g., `["SRC"]`) but the CLI sends it as a **string** (`"SRC"`). The dry-run output shows the field as a plain string, which the server-side validation seems to reject.

Same issue affects `block_copy_insert_after` (verified independently).

### Impact

[...]

PR #1427, opened 2026-06-12
feat(docs): add opt-in --emulate fallback for block_move_after/block_copy_insert_after

leeguooooo:
## Summary

On some server deployments (reproduced on a JP-tenant Lark international workspace, see #758), `docs +update --command block_move_after` / `block_copy_insert_after` fail with `command 'block_move_after' requires content or src_block_ids but both were empty` even though the CLI sends `src_block_ids` correctly — the server side does not read the field. This PR adds an opt-in `--emulate` flag that reproduces move/copy semantics client-side with the primitives that do work, so agents are not blocked while the server bug persists. Default behavior is completely unchanged; without `--emulate` both commands still send the same single request as today.

## Changes

- Add `--emulate` (bool, default false) to `docs +update`, valid only with `block_move_after` / `block_copy_insert_after`. When set, the command runs a client-side orchestration: fetch source blocks (full detail) → rebuild them after `--block-id` via `block_insert_after` → re-fetch to verify the rebuilt blocks exist → `block_delete` the originals (move only). Insert always precedes delete and any failure stops immediately with the inserted/deleted state reported on stderr, so no path loses source content (`shortcuts/doc/docs_update_emulate.go`).
- Supported block types: `p`, `h1`–`h9`, `ul`, `ol`, `li` (re-wrapped in its original list type), `blockquote`, `callout` (rgb colors dropped, emoji kept), `pre`, `img` (rebuilt from `href`, effectively a re-upload). Block types carrying server-side tokens or state (`whiteboard`, `sheet`, `bitable`, `table`, `synced_reference`, …) are rejected with a typed `failed_precondition` error before any write.
- The emulation is not semantically identical to the server-side command and says so explicitly — in the flag help, in a stderr warning on every run, and in a `semantic_differences` field of the success envelope: rebuilt blocks get new block ids, comments and `#block_id` anchors on the originals are not migrated, and image file tokens change.
- `--emulate` rejects a non-default `--revision-id` (the orchestration issues multiple writes against the latest revision) and duplicate ids in `--src-block-ids`.
- `--dry-run --emulate` prints the multi-call plan (fetch → insert → verify-fetch → delete for move; three calls for copy).
- Unit tests for the XML block indexer / rebuild rules and httpmock-based orchestration tests (move, copy keeps originals, unsupported type rejected before any write, verification failure keeps originals); dry-run E2E cases for both commands.
- Note the fallback (and its semantic differences) in the `lark-doc-update.md` skill reference so agents discover it next to the move/copy documentation.

## Test Plan

- [x] `make unit-test` (full `-race` suite green)
- [x] `go vet ./...`
- [x] `gofmt -l .` (no output)
- [x] `go mod tidy` (no go.mod/go.sum changes)
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run --new-from-rev=origin/main` (0 issues)
- [x] `LARK_CLI_BIN=$PWD/lark-cli go test ./tests/cli_e2e/docs/ -run TestDocs_DryRunDefaultsToV2OpenAPI`
- [x] Live verification on a scratch docx in the affected JP tenant:
  - without `--emulate`: unchanged behavior — the server still does not perform the move (observed both the #758 hard error and, on a later run, `result: "failed"` with degrade warning "Instruction produced no document changes"; either way `src_block_ids` is not honored)
  - `--emulate` single paragraph move, multi-block move (`p` + `li`), `img` move to `-1` (re-uploaded, new src token), copy keeping the original
  - `table` source block rejected with `failed_precondition` before any write

## Related Issues

- #758 (server-side reproduction evidence in the issue comments; this PR is the client-side opt-in workaround until the server fix lands)

## Summary by CodeRabbit

* **New Features**
  * Added an `--emulate` flag for `block_move_after` and `block_copy_insert_after` as a client-side fallback when servers reject requests due to missing `src_block_ids`, including dry-run and execute support.
  * Emulation rebuilds eligible blocks, inserts after the anchor, verifies the insertion is visible/contiguous, and deletes originals only for move (conservatively aborts on verification failures).

* **Documentation**
  * Updated `docs +update` guidance with behavior notes and limitations (block-id changes, link/comment behavior, image handling, supported block types, stop-on-failure).

* **Tests**
  * Added extensive unit, dry-run, and end-to-end coverage, including validation, error paths, and concurrency scenarios.

### DIFF
--- a/shortcuts/doc/docs_update_emulate.go
+++ b/shortcuts/doc/docs_update_emulate.go
@@ -216,6 +216,8 @@ func indexTopLevelBlocks(xml string) []topLevelBlock {
 	return out
 }
 
+// idAttrSet returns the block id(s) carried by an attribute run as a set,
+// or an empty set when the run has no id attribute.
 func idAttrSet(attrs string) map[string]bool {
 	ids := map[string]bool{}
 	if m := emulateBlockIDRE.FindStringSubmatch(attrs); m != nil {
@@ -229,6 +231,8 @@ func idAttrSet(attrs string) map[string]bool {
 // xmlAttrValue call.
 var xmlAttrValueRECache sync.Map // name string -> *regexp.Regexp
 
+// xmlAttrValue returns the value of the named attribute within an attribute
+// run and whether it was present.
 func xmlAttrValue(attrs, name string) (string, bool) {
 	re, ok := xmlAttrValueRECache.Load(name)
 	if !ok {
@@ -242,11 +246,15 @@ func xmlAttrValue(attrs, name string) (string, bool) {
 	return m[1], true
 }
 
+// xmlAttrEscape escapes a raw string for safe inclusion in a double-quoted
+// XML attribute value.
 func xmlAttrEscape(raw string) string {
 	r := strings.NewReplacer("&", "&amp;", "<", "&lt;", ">", "&gt;", `"`, "&quot;")
 	return r.Replace(raw)
 }
 
+// xmlEntityDecode reverses xmlAttrEscape, decoding the XML entities that
+// DocxXML export emits back into their literal characters.
 func xmlEntityDecode(s string) string {
 	r := strings.NewReplacer("&lt;", "<", "&gt;", ">", "&quot;", `"`, "&#39;", "'", "&amp;", "&")
 	return r.Replace(s)
@@ -336,6 +344,8 @@ func buildEmulatedBlockContent(b emulatedBlock) (content, insertedTag string, er
 	return inner, b.tag, nil
 }
 
+// splitSrcBlockIDs splits a comma-separated --src-block-ids value into trimmed,
+// non-empty block ids, preserving order.
 func splitSrcBlockIDs(raw string) []string {
 	var ids []string
 	for _, id := range strings.Split(raw, ",") {
--- a/shortcuts/doc/docs_update_emulate.go
+++ b/shortcuts/doc/docs_update_emulate.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/larksuite/cli/errs"
 	"github.com/larksuite/cli/shortcuts/common"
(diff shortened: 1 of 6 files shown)


leeguooooo:
Thanks @coderabbitai — addressed all three in 43b811e:

**1. Anchor-scoped verification (the data-loss one) — fixed.** You're right that "any new block id appeared" would, in a shared doc, treat a concurrent collaborator's edit as proof of our own insert and then delete the originals. Verification is now anchor-scoped: it confirms the rebuilt blocks landed as a *contiguous run immediately after the anchor* (document start for a page-id anchor, the final slots for `-1`) with the expected count and top-level tag sequence. Anything that doesn't line up fails conservatively and keeps the originals (recoverable duplicate over data loss). "Insert → verify → delete, stop on any failure" is preserved. The top-level walker matches Lark's flat DocxXML export (sibling blocks, no wrapping `<page>`); I caught and fixed a fixture/`-1`-anchor mismatch here via live testing on a scratch doc.

**2. Tag the offending `--src-block-ids` — fixed.** `buildEmulatedBlockContent` now attaches `Param: --src-block-ids` to both the unsupported-type rejection and the missing-`href` img rejection, so the command layer can surface which source block failed.

**3. Typed-error assertions — fixed.** The two error-path tests now assert the typed envelope (category/subtype/param) via the package's shared `assertValidationContract` helper instead of `err.Error()` substrings. Added coverage for the concurrent-edit guard and the document-start/`-1` anchor cases.

Re-verified after the changes: full `-race` unit suite, `go vet`, `gofmt`, `golangci-lint --new-from-rev` (0 issues), dry-run E2E, and the live six-case matrix (text move, multi-block move, img move to `-1`, copy, unsupported-type rejection, no-`--emulate` passthrough unchanged) all green.

leeguooooo:
Thanks @coderabbitai — picked up both items from the 43b811e pass.

**1. Typed-error assertions in `TestDocsUpdateEmulateValidation` — fixed (c09ad57, refined in 28326ec).** The four validation cases now pin the typed envelope instead of `err.Error()` substrings, routed through the package's shared `assertValidationContract` helper (same one already used by the unsupported-type / missing-`href` cases in this file), so each case asserts `CategoryValidation` + `SubtypeInvalidArgument` + the offending `Param` (`--emulate`, `--revision-id`, `--src-block-ids`, `--src-block-ids`). The message-substring check is kept as a secondary assertion so the human-facing wording stays covered.

**2. `xmlAttrValue` regex recompilation — fixed (c09ad57).** It now memoizes the per-attribute-name regex via a package-level `sync.Map`, so re-inserting an image-heavy document no longer recompiles the same pattern on every lookup. Kept it on the cold emulation path only; no change to the default (non-`--emulate`) flow.

Also raised docstring coverage on both touched files from 42% to 100% (cc4e47c) to clear the coverage gate — comments only, no behavior change.

Re-verified: full `-race` unit suite (217 green), `go vet`, `gofmt -l` (no output) all clean.

PR #1428, opened 2026-06-12
test(doc): derive fetch test flag defaults from v2FetchFlags

vegerot:
Replace hardcoded flag defaults in the fetch test helpers with
fetchDefault() / fetchDefaultInt() helpers that read the declared
defaults from v2FetchFlags(). This prevents future drift between
production flag defaults and test setup, and panics loudly if a
flag name is misspelled rather than silently returning "".

The tests now correctly avoid hardcoding doc-format, but other
flag defaults (detail, revision-id, scope, etc.) were still
duplicated here. Deriving all defaults from v2FetchFlags() keeps
the whole test command definition aligned with production.

Co-authored-by: TraeCli (Doubao-Seed-Dogfooding) <trae@bytedance.com>

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/larksuite/cli/pull/1428).
* #1424
* __->__ #1428


fangshuyu-768:
Thanks for the cleanup here. The change looks useful as a small test-maintenance improvement: deriving the test defaults from `v2FetchFlags()` should help avoid drift between the test fixture and the production flag definitions.

This PR is currently showing merge conflicts, so it will need a rebase or conflict resolution before we can continue reviewing or merge it.

fangshuyu-768 (approved):

Issue #1392, opened 2026-06-11
Proposal: ship a single umbrella skill (router + on-demand domain guides) to cut the always-on agent context cost of 26 skills


qiankunli:
## Problem

`npx skills add larksuite/cli -g` registers **26 separate skills** with the agent harness. The way harnesses like Claude Code / Codex load skills, the `name` + `description` of *every* registered skill is injected into the model context at **every session start** — only the SKILL.md body and references are loaded on demand (progressive disclosure).

Measured on current `main`: the frontmatter `name` + `description` across the 26 `lark-*` skills total **~6,300 chars** (mostly CJK, so a sizable token cost). That cost is paid in **every agent session, including sessions that never touch Lark at all**. Harnesses also apply budgets/limits to skill metadata, so 26 entries crowd out the user's other skills.

A second, related problem: users who hand-copy skills (or never re-run the install command) silently drift out of sync with the CLI binary. Skill lifecycle (install **and** update) should be owned by the CLI so the skills always match the binary.

## Proposal: a single umbrella skill, generated from the existing 26

Ship one additional skill `lark` that acts as a **router**, with the 26 domain skills mirrored inside it as on-demand reference guides:

```
skills/lark/
  SKILL.md                          # router: ~370-char description + routing table
  references/subskills/
    lark-doc/GUIDE.md               # mirror of skills/lark-doc/SKILL.md
    lark-doc/references/...         # mirror of skills/lark-doc/references/
    lark-sheets/GUIDE.md
    ...                            # all 26 domains
```

- Always-resident context drops from ~6,300 chars (26 descriptions) to **~370 chars (1 description) — a ~94% reduction**. Domain guides cost zero context until the agent actually `Read`s them.
- The umbrella is **generated at build time** from the existing skills — no double maintenance:
  - each `skills/lark-<domain>/` is mirrored to `skills/lark/references/subskills/lark-<domain>/`, with `SKILL.md` renamed to `GUIDE.md` (so harnesses that scan recursively never re-register the nested copies as standalone skills);
  - the sibling directory structure is preserved, so the **~380 existing cross-skill relative links** (`../lark-shared/SKILL.md`, `../../lark-doc/references/lark-doc-xml.md`, …) keep their exact depth — the only rewrite needed is the `SKILL.md` → `GUIDE.md` filename swap;
  - the routing table in the umbrella SKILL.md embeds each subskill's frontmatter `description` **verbatim**, so the per-domain descriptions remain the single source of truth for triggering.
[...]

PR #1432, opened 2026-06-12
fix(skills): add single entry skill

### LINKED ISSUE #1385: 就不能做成一个技能吗？找其他技能的时候，列表都被lark污染了！
就不能做成一个技能，再通过主目录的形式连接其他的md文件吗，技能太多了，claude code都提醒技能描述超过限制了，一个IM工具占了90%以上的skill描述，就离谱至极

xu91102:
## Background

Fixes https://github.com/larksuite/cli/issues/1385.

The repository currently exposes every `skills/lark-*` domain skill by default. In skill pickers/search UIs this floods the global skill list with Lark-specific entries and long domain descriptions, especially for large domains such as IM.

## Changes

- Add a root `lark-suite` `SKILL.md` as the default single entry skill.
- Keep domain-specific guidance in the existing embedded `skills/lark-*` docs and route agents to them through `lark-cli skills read <domain-skill>` only when needed.
- Update README / root help to describe the default single entry skill and document `--full-depth` for users who still want direct nested domain skill discovery.
- Extend the skill format checker and workflow so the root `SKILL.md` is validated alongside `skills/`.

## Verification

- `npx -y skills add . --list --json`: passed — reports `Found 1 skill` and lists `lark-suite`.
- `npx -y skills add . --list --full-depth --json`: passed — nested domain skills such as `lark-im` and `lark-calendar` remain discoverable when explicitly requested.
- `npx -y skills use . --skill lark-suite`: passed — renders the new entry skill.
- `bash scripts/skill-format-check/test.sh`: passed.
- `node scripts/skill-format-check/index.js .`: passed.
- `node scripts/skill-format-check/index.js skills`: passed.
- `go test -count=1 -timeout=3m ./cmd/skill ./internal/skillscheck ./cmd/update`: passed.
- `gofmt -l cmd/root.go internal/skillscheck/sync.go cmd/update/update.go`: no output.
- `git diff --check origin/main...HEAD`: passed.

## Risk and rollback

Default `npx skills add larksuite/cli` now installs the single entry skill instead of every nested domain skill. This intentionally reduces global skill-list noise. Existing domain skills remain in the repository and can still be discovered with `--full-depth` or read via `lark-cli skills read`. Rollback is reverting this PR.

## Summary by CodeRabbit

* **Documentation**
  * Added `SKILL.md` documentation with skill metadata and command guidance.
  * Updated README and Chinese README to reflect simplified skill model with single default entry.
  * Enhanced CLI help text with updated installation and discovery instructions.

### DIFF
--- a/.github/workflows/skill-format-check.yml
+++ b/.github/workflows/skill-format-check.yml
@@ -4,12 +4,14 @@ on:
   push:
     branches: [main]
     paths:
+      - "SKILL.md"
       - "skills/**"
       - "scripts/skill-format-check/**"
       - ".github/workflows/skill-format-check.yml"
   pull_request:
     branches: [main]
     paths:
+      - "SKILL.md"
       - "skills/**"
       - "scripts/skill-format-check/**"
       - ".github/workflows/skill-format-check.yml"
@@ -29,4 +31,6 @@ jobs:
           node-version: '20'
 
       - name: Run Skill Format Check
-        run: node scripts/skill-format-check/index.js
+        run: |
+          node scripts/skill-format-check/index.js .
+          node scripts/skill-format-check/index.js skills
--- a/README.md
+++ b/README.md
@@ -6,14 +6,14 @@
 
 [中文版](./README.zh.md) | [English](./README.md)
 
-The official [Lark/Feishu](https://www.larksuite.com/) CLI tool, maintained by the [larksuite](https://github.com/larksuite) team — built for humans and AI Agents. Covers core business domains including Messenger, Docs, Base, Sheets, Slides, Calendar, Mail, Tasks, Meetings, Markdown, and more, with 200+ commands and 26 AI Agent [Skills](./skills/).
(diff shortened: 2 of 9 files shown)


MaxHuang22:
Thanks for this PR — it's definitely aimed at the right problem (#1385 / #1392). After evaluating it, though, it diverges from the direction we're driving on a few key points, so I'm closing it with an explanation:

1. **Default behavior.** This PR flips the default to a single entry (only `lark-suite` is installed by default; domain skills fall back to `--full-depth`). Our direction is to **keep the separate domain skills as the default, with the umbrella as an opt-in mode and the user's choice remembered** — not to flip the default for everyone.

2. **Primary source.** The single-entry behavior relies on a root `SKILL.md` short-circuiting the local directory scan, but the primary install source `open.feishu.cn` is index-driven, and that path isn't verified here — on the real install flow it most likely won't take effect.

3. **Existing users.** `skills add` only adds, never removes, so users who already installed the domain skills would end up with both side by side, with no reduction in context.

Your work on the root `SKILL.md` routing and the format-check extension is valuable, and we'll draw on it as we build out the opt-in umbrella mode. We're tracking progress in #1392 — please feel free to keep the conversation going there. Thanks again!

PR #1445, opened 2026-06-13
fix(core): add LARKSUITE_CLI_FORCE_LOCAL escape hatch for inherited Agent env

### LINKED ISSUE #1405: lark-cli enters Hermes context when HERMES_HOME is inherited, then bind ignores HERMES_HOME path
## Summary

After installing Hermes, unrelated CLI/automation processes that happen to inherit `HERMES_HOME` start failing when running `lark-cli`.

`lark-cli` detects a Hermes context and refuses to use the existing local lark-cli config, returning:

```json
{
  "ok": false,
  "error": {
    "type": "config",
    "subtype": "not_configured",
    "message": "hermes context detected but lark-cli is not bound to it",
    "hint": "read `lark-cli config bind --help`, then ask the user to confirm intent and identity preset..."
  }
}
```

However, when trying to bind Hermes explicitly, `lark-cli` does not appear to use the `HERMES_HOME` path that triggered Hermes detection. Instead, it looks for a different default path:

```json
{
  "ok": false,
  "error": {
    "type": "config",
    "subtype": "invalid_config",
    "message": "failed to read Hermes config: open C:\\Users\\example\\.hermes\\.env: The system cannot find the path specified.",
    "hint": "verify Hermes is installed and configured at C:\\Users\\example\\.hermes\\.env"
  }
}
```

In this environment, `HERMES_HOME` points somewhere else, for example:

```text
HERMES_HOME=C:\Users\example\AppData\Local\hermes
```

## Expected behavior

If `lark-cli` detects Hermes because `HERMES_HOME` is present, then one of the following should happen:

1. `config bind --source hermes` should read Hermes config from `HERMES_HOME`, or
2. `lark-cli` should not enter Hermes context unless Hermes config can actually be resolved, or
3. there should be a documented way to force normal local config mode, e.g. an env var or flag like `--ignore-agent-context`.

## Actual behavior

Any unrelated process that inherits `HERMES_HOME` is treated as Hermes context.

This breaks existing `lark-cli` usage in non-Hermes automation tasks. Commands such as:

```powershell
lark-cli auth status
lark-cli config show
lark-cli im +chat-messages-list --as user ...
```

fail before using the existing local config.

If `HERMES_HOME` is removed from the current shell, the same commands work again and use the normal local config:

```powershell
Remove-Item Env:HERMES_HOME -ErrorAction SilentlyContinue
lark-cli auth status
```

## Reproduction

1. Have a working local `lark-cli` config.
2. Install Hermes or otherwise set `HERMES_HOME` to a Hermes data directory.
3. In a non-Hermes shell or automation process that inherits `HERMES_HOME`, run:

```powershell
lark-cli auth status
```

4. Observe `hermes context detected but lark-cli is not bound to it`.
5. Try:

```powershell
lark-cli config bind --source hermes --identity user-default
```

6. Observe that bind looks for `C:\Users\example\.hermes\.env` instead of the path from `HERMES_HOME`.

## Impact

This causes unrelated automation jobs that use `lark-cli` to fail after Hermes is installed, even though those jobs are not Hermes jobs and already have a valid local lark-cli config.

## Version

```text
lark-cli version 1.0.51
```

nguyenngothuong:
## Summary

Fixes #1405.

Agent path variables such as `HERMES_HOME` / `OPENCLAW_HOME` are user-facing and get exported into shell profiles, so they are inherited by **every** child process — not just Agent subprocesses. Any unrelated automation that merely inherits `HERMES_HOME` is routed into the Hermes workspace and refuses to use its valid local `lark-cli` config, failing with:

```json
{ "ok": false, "error": { "type": "config", "subtype": "not_configured",
  "message": "hermes context detected but lark-cli is not bound to it" } }
```

This implements **option 3** from the issue: a documented way to force normal local config mode without unsetting the inherited variable.

## Change

`DetectWorkspaceFromEnv` now honors `LARKSUITE_CLI_FORCE_LOCAL == "1"` as a priority-0 escape hatch that returns `WorkspaceLocal`, overriding every Agent signal. The check uses the same strict `== "1"` convention as `OPENCLAW_CLI` / `HERMES_QUIET` / `LARK_CHANNEL`, and the env-var naming matches the existing `LARKSUITE_CLI_CONFIG_DIR`.

The agent "not bound" hint now mentions the escape hatch so operators who hit the wall can discover it.

```sh
# inherited HERMES_HOME → hijacked into hermes context (the bug)
HERMES_HOME=/some/path lark-cli config show
#   → "hermes context detected but lark-cli is not bound to it"

# escape hatch → back to the normal local config
HERMES_HOME=/some/path LARKSUITE_CLI_FORCE_LOCAL=1 lark-cli config show
#   → uses local ~/.lark-cli config as before
```

## Notes

- The "bind ignores `HERMES_HOME`" part of the issue appears already addressed on `main`: `resolveHermesEnvPath()` (cmd/config/binder.go) already respects `HERMES_HOME`. This PR targets the remaining root cause — false-positive workspace detection.
- Scope is intentionally minimal (the escape hatch). Happy to follow up on a narrower detection model for Hermes (e.g. a subprocess marker like OpenClaw's `OPENCLAW_CLI`, downgrading `HERMES_HOME` to a secondary signal) if maintainers prefer that direction.

## Tests

- New table cases in `workspace_test.go` covering override-of-HERMES_HOME, override-of-all-signals, strict `== "1"` parsing (`true`/`0` ignored), and the no-op case.
- Updated the e2e unbound-workspace hint assertion.
- `go build ./...`, `gofmt`, `go vet`, and `go test ./internal/core/...` + the e2e config suite pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added an escape hatch mechanism to force use of local configuration, allowing users to bypass inherited workspace detection settings when needed.

* **Documentation**
  * Enhanced configuration binding guidance with additional instructions for cases where inherited context is triggered and users prefer local configuration.

* **Tests**
  * Added comprehensive test cases validating the override mechanism behavior across multiple configuration scenarios.

### DIFF
--- a/internal/core/notconfigured.go
+++ b/internal/core/notconfigured.go
@@ -52,7 +52,7 @@ const (
 	// the binding workflow" guidance. Always points at --help (never a
 	// ready-to-run bind command) so the AI reads the confirmation
 	// discipline (identity preset, user opt-in) before acting.
-	agentBindHint = "read `lark-cli config bind --help`, then ask the user to confirm intent and identity preset (bot-only or user-default); only after both are confirmed, run `lark-cli config bind`"
+	agentBindHint = "read `lark-cli config bind --help`, then ask the user to confirm intent and identity preset (bot-only or user-default); only after both are confirmed, run `lark-cli config bind`. If this is NOT an Agent job and the Agent context was triggered only by an inherited env var (e.g. HERMES_HOME), set LARKSUITE_CLI_FORCE_LOCAL=1 to use the normal local config instead."
 )
 
 // NotConfiguredError returns the canonical "not configured" error, with a
--- a/internal/core/workspace.go
+++ b/internal/core/workspace.go
@@ -72,6 +72,14 @@ func (w Workspace) IsLocal() bool {
 
 // DetectWorkspaceFromEnv determines the workspace from process environment.
 //
+// Escape hatch: LARKSUITE_CLI_FORCE_LOCAL == "1" forces WorkspaceLocal,
+// overriding every Agent signal below. Agent path variables such as
+// HERMES_HOME / OPENCLAW_HOME are user-facing and get exported into shell
+// profiles, so unrelated automation that merely inherits them would otherwise
+// be misrouted into an Agent context it never opted into (see issue #1405).
+// Setting LARKSUITE_CLI_FORCE_LOCAL=1 in such jobs restores normal local
+// config without unsetting the inherited Agent variables.
+//
 // Detection is signal-based, not credential-based: we look for environment
 // variables that the host Agent itself sets when launching a subprocess.
 // Generic FEISHU_APP_ID / FEISHU_APP_SECRET are intentionally NOT used —
@@ -79,6 +87,8 @@ func (w Workspace) IsLocal() bool {
 // false-positive routing into a Hermes workspace.
 //
 // Priority:
+//  0. LARKSUITE_CLI_FORCE_LOCAL == "1" → WorkspaceLocal (escape hatch,
+//     overrides every signal below; see note above).
 //  1. Any OpenClaw signal → WorkspaceOpenClaw
 //     - OPENCLAW_CLI == "1":  subprocess marker (added 2026-03-09 via
 //     OpenClaw PR #41411). Most precise, but absent on older builds.
@@ -100,6 +110,9 @@ func (w Workspace) IsLocal() bool {
 //     mirrors the OPENCLAW_CLI / HERMES_QUIET style.
 //  4. Otherwise → WorkspaceLocal
 func DetectWorkspaceFromEnv(getenv func(string) string) Workspace {
+	if getenv("LARKSUITE_CLI_FORCE_LOCAL") == "1" {
+		return WorkspaceLocal
+	}
 	if getenv("OPENCLAW_CLI") == "1" ||
 		getenv("OPENCLAW_HOME") != "" ||
 		getenv("OPENCLAW_STATE_DIR") != "" ||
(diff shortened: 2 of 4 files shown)


evandance:
Thanks for the PR and the clear root-cause notes — the analysis of where the signal comes from is genuinely helpful.

The bind-path part needs no change. On `main` (and already in 1.0.51) `config bind` resolves the path from `HERMES_HOME` and only falls back to the default when it's unset. Within a single run, detection and path resolution read the same `HERMES_HOME`, so they can't point at different locations — the default-path output only appears when `HERMES_HOME` is empty. So there's nothing to fix on that half.

For the escape hatch, I don't think we should merge it. Here's the reasoning.

"Context detected but not bound" is not a bug — it's how the isolation is meant to work. `config bind` keeps each Agent's credentials separate so that an Agent context never quietly reuses the developer's personal local config.

The real question is why `HERMES_HOME` is visible to a job that isn't a Hermes job. Hermes sets that variable per-process for the processes it launches; a normal install does not put it in a global shell profile or system environment. (The Docker image sets it container-wide, but that container *is* a Hermes context.) So a global `HERMES_HOME` that unrelated jobs inherit is an environment problem, and it should be fixed where the variable is set — not worked around in the CLI.

A permanent `LARKSUITE_CLI_FORCE_LOCAL` flag also opens a hole in that isolation. It uses the same channel that caused the problem — environment variables — to switch the CLI back to the developer's identity inside an Agent context. Anything that can set an env var could flip it. We'd rather not trade that safety property for a workaround to a misconfigured environment.

Two smaller points:
- The code comment says the agents export `HERMES_HOME` / `OPENCLAW_HOME` into shell profiles. They don't, so that shouldn't be the stated reason.
- I'd also hold off on demoting `HERMES_HOME` to a secondary signal. Some Hermes processes set only `HERMES_HOME`, so demoting it would break detection for real Hermes subprocesses, and the CLI can't rely on a marker Hermes doesn't send.

I've left guidance on the issue for the reporter on how to scope the variable. Thanks again for the careful work.

PR #1501, opened 2026-06-17
docs: clarify doc block id lifecycle

fangshuyu-768:
## Summary
- add a Block ID lifecycle table for docs +update commands
- make lark-doc SKILL.md require reading the update reference when editing existing docs
- call out when old block IDs are invalidated, when new IDs must be fetched, and when move operations require refreshing position context

## Tests
- go test ./shortcuts/doc
- node scripts/skill-format-check/index.js


fangshuyu-768:
Closing in favor of #1508. The block ID lifecycle guidance changes from this PR have been folded into #1508.

PR #1524, opened 2026-06-20
fix(sheets): translate rich text segment styles

### LINKED ISSUE #1356: Bug: sheets +cells-set rich_text segment styles are not applied
### Summary

`lark-cli sheets +cells-set` accepts `rich_text[].style` according to the CLI schema, but text segment styles such as `font_line: "line-through"` and `font_color` do not appear to be applied when written through the new `set_cell_range` shortcut path.

The same rich text styling works when using the lower-level Sheets `values` API with `segmentStyle`, and it also works through the deprecated `sheets +write` compatibility command.

### Environment

- `lark-cli version 1.0.49`
- Auth identity tested: bot
- Target: a throwaway Lark spreadsheet

### Reproduction

Create or use any spreadsheet, then run:

```bash
lark-cli sheets +cells-set \
  --as bot \
  --spreadsheet-token "$SPREADSHEET_TOKEN" \
  --sheet-id "$SHEET_ID" \
  --range A1 \
  --cells '[[{"rich_text":[{"type":"text","text":"example: "},{"type":"text","text":"OLD","style":{"font_line":"line-through"}},{"type":"text","text":"NEW","style":{"font_color":"#00B050"}}]}]]' \
  --json
```

Then read back:

```bash
lark-cli sheets +cells-get \
  --as bot \
  --spreadsheet-token "$SPREADSHEET_TOKEN" \
  --sheet-id "$SHEET_ID" \
  --range A1 \
  --include value,style,comment,rich_text \
  --json
```

### Actual behavior

- The write succeeds.
- The cell text is present.
- The response does not include `rich_text` segment styling for the cell.
- Exporting the workbook to xlsx shows text runs, but the run formatting is empty; no strikethrough or color is applied.

Dry run shows `+cells-set` going through the Sheet AI write tool:

```text
POST /open-apis/sheet_ai/v2/spreadsheets/<token>/tools/invoke_write
tool_name: set_cell_range
```

with input similar to:

```json
{
  "cells": [[{
    "rich_text": [
      {"type": "text", "text": "example: "},
      {"type": "text", "text": "OLD", "style": {"font_line": "line-through"}},
      {"type": "text", "text": "NEW", "style": {"font_color": "#00B050"}}
    ]
  }]]
}
```

### Expected behavior

The rich text segment styles documented by `+cells-set --print-schema --flag-name cells` should be applied to the text segments:

- `font_line: "line-through"` should render as strikethrough.
- `font_color: "#00B050"` should render as green text.
- `+cells-get --include rich_text` should be able to read back the rich text styling.

### Workarounds verified

The lower-level Sheets values API works when using `segmentStyle`:

```bash
lark-cli api PUT /open-apis/sheets/v2/spreadsheets/$SPREADSHEET_TOKEN/values \
  --as bot \
  --data - \
  --json <<'JSON'
{
  "valueRange": {
    "range": "SHEET_ID!A2:A2",
    "values": [[[
      {"type":"text","text":"example: ","segmentStyle":{}},
      {"type":"text","text":"OLD","segmentStyle":{"strikeThrough":true}},
      {"type":"text","text":"NEW","segmentStyle":{"foreColor":"#00B050"}}
    ]]]
  }
}
JSON
```

The deprecated compatibility command also works because it still uses the lower-level `values` endpoint:

```bash
lark-cli sheets +write \
  --as bot \
  --spreadsheet-token "$SPREADSHEET_TOKEN" \
  --sheet-id "$SHEET_ID" \
  --range A3:A3 \
  --values '[[[{"type":"text","text":"example: ","segmentStyle":{}},{"type":"text","text":"OLD","segmentStyle":{"strikeThrough":true}},{"type":"text","text":"NEW","segmentStyle":{"foreColor":"#00B050"}}]]]' \
  --json
```

### Notes

This looks like a mismatch between the new `+cells-set` shortcut schema (`rich_text[].style`, same shape as `cell_styles`) and the underlying format required for actual Sheet rich text segment styling (`segmentStyle` with fields such as `strikeThrough` and `foreColor`).

`+batch-update` appears to have the same issue for rich text segment styles because it translates `+cells-set` operations to the same `set_cell_range` tool.

Zhang-986:
## Summary
- translate `+cells-set` rich_text segment `style` objects into Sheets `segmentStyle` fields before calling `set_cell_range`
- map font color, font size, bold, italic, underline, and strikethrough segment styles
- add dry-run regression coverage for line-through and font color segment styles

Fixes #1356.

## Tests
- `go test ./shortcuts/sheets -run TestCellsSet_RichTextStyleTranslatedToSegmentStyle -count=1` (fails before the fix, passes after)
- `go test ./shortcuts/sheets -count=1`
- `go test ./cmd ./shortcuts/sheets -count=1`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced rich text formatting in spreadsheets. Text styling attributes are now properly handled across different format versions, ensuring consistent visual presentation.

* **Tests**
  * Added comprehensive test coverage for rich text style translation functionality.

### DIFF
--- a/shortcuts/sheets/lark_sheet_write_cells.go
+++ b/shortcuts/sheets/lark_sheet_write_cells.go
@@ -88,6 +88,7 @@ func cellsSetInput(runtime flagView, token, sheetID, sheetName string) (map[stri
 	if err != nil {
 		return nil, err
 	}
+	normalizeRichTextSegmentStyles(cells)
 	input := map[string]interface{}{
 		"excel_id": token,
 		"range":    strings.TrimSpace(runtime.Str("range")),
@@ -104,6 +105,82 @@ func cellsSetInput(runtime flagView, token, sheetID, sheetName string) (map[stri
 		return nil, err
 	}
 	return input, nil
+}
+
+func normalizeRichTextSegmentStyles(cells []interface{}) {
+	for _, rowValue := range cells {
+		row, ok := rowValue.([]interface{})
+		if !ok {
+			continue
+		}
+		for _, cellValue := range row {
+			cell, ok := cellValue.(map[string]interface{})
+			if !ok {
+				continue
+			}
+			segments, ok := cell["rich_text"].([]interface{})
+			if !ok {
+				continue
+			}
+			for _, segmentValue := range segments {
+				segment, ok := segmentValue.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				style, ok := segment["style"].(map[string]interface{})
+				if !ok {
+					continue
+				}
+				segmentStyle := map[string]interface{}{}
+				if existing, ok := segment["segmentStyle"].(map[string]interface{}); ok {
+					for k, v := range existing {
+						segmentStyle[k] = v
+					}
+				}
+				mergeRichTextStyle(segmentStyle, style)
+				if len(segmentStyle) > 0 {
+					segment["segmentStyle"] = segmentStyle
+				}
+				delete(segment, "style")
+			}
+		}
+	}
+}
+
+func mergeRichTextStyle(segmentStyle map[string]interface{}, style map[string]interface{}) {
+	if v, ok := style["font_color"]; ok {
+		segmentStyle["foreColor"] = v
+	}
+	if v, ok := style["font_size"]; ok {
+		segmentStyle["fontSize"] = v
+	}
+	if v, ok := style["font_weight"].(string); ok {
+		switch v {
+		case "bold":
+			segmentStyle["bold"] = true
+		case "normal":
+			segmentStyle["bold"] = false
+		}
+	}
+	if v, ok := style["font_style"].(string); ok {
+		switch v {
+		case "italic":
+			segmentStyle["italic"] = true
+		case "normal":
+			segmentStyle["italic"] = false
+		}
+	}
+	if v, ok := style["font_line"].(string); ok {
+		switch v {
+		case "line-through":
+			segmentStyle["strikeThrough"] = true
+		case "underline":
+			segmentStyle["underline"] = true
+		case "none":
+			segmentStyle["strikeThrough"] = false
+			segmentStyle["underline"] = false
+		}
+	}
 }
 
 // CellsSetStyle stamps a single style block across every cell in --range.
--- a/shortcuts/sheets/lark_sheet_write_cells_test.go
+++ b/shortcuts/sheets/lark_sheet_write_cells_test.go
@@ -115,6 +115,42 @@ func TestWriteCellsShortcuts_DryRun(t *testing.T) {
 	}
 }
 
+func TestCellsSet_RichTextStyleTranslatedToSegmentStyle(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


xiongyuanwen-byted:
## ⚠️ This change introduces a regression — `set_cell_range` rejects `segmentStyle`

I verified the actual end-to-end behavior (not just dry-run), and this PR would **break** rich-text styling rather than fix it.

`normalizeRichTextSegmentStyles` rewrites each segment's `style` (cell_styles shape) into `segmentStyle` and `delete`s the original `style`. But the `set_cell_range` tool's schema **does not accept a `segmentStyle` property** — sending it fails server-side parameter validation:

**PR-shaped request (what this PR makes the CLI send), real call:**
```
sheets +cells-set --range D1 --cells '[[{"rich_text":[
  {"type":"text","text":"OLD","segmentStyle":{"strikeThrough":true}},
  {"type":"text","text":"NEW","segmentStyle":{"foreColor":"#00B050"}}]}]]'
```
```
tool "set_cell_range" failed: [900015206] parameter validation failed:
[cells[0][0].rich_text[1].segmentStyle] unexpected property "segmentStyle" is not defined in schema
```
→ the write is **rejected outright** (worse than #1356, where the write at least succeeded).

**Current-shaped request (`style`, cell_styles shape — what the CLI sends today), real call:**
```
sheets +cells-set --range E1 --cells '[[{"rich_text":[
  {"type":"text","text":"OLD","style":{"font_line":"line-through"}},
  {"type":"text","text":"NEW","style":{"font_color":"#00B050"}}]}]]'
```
→ write succeeds; `+cells-get --include rich_text` reads back `style.textDecoration: 2` (strikethrough) and `style.foreColor: "#00B050"`. ✅

### Why the PR's tests don't catch this
`TestCellsSet_RichTextStyleTranslatedToSegmentStyle` is a **dry-run** assertion on the request body — it never issues a real request, so the server-side `900015206` validation failure is invisible to it.

### Root cause: two different API layers
- `segmentStyle` (`strikeThrough` / `foreColor` / `bold` …) belongs to the **low-level values API** (`PUT /open-apis/sheets/v2/spreadsheets/<tok>/values`) — that's the workaround in #1356.
- `set_cell_range` is the **sheet_ai tool**; its `rich_text[].style` contract is the **cell_styles shape** (`font_line` / `font_color` / `font_weight` …), and the backend maps it to the underlying `segmentStyle` itself.
- This PR borrows the low-level field names into the tool request body, which the tool doesn't recognize.

### #1356 is already fixed on the backend
The `set_cell_range` tool already translates cell_styles-shaped segment styles to the underlying format (`font_line:"line-through"` → `textDecoration:2`, `font_color` → `foreColor`), verified end-to-end on `v1.0.58`. The CLI's existing pass-through is correct, so **no CLI change is needed** and #1356 has been closed as resolved.

**Recommendation: close this PR.** If a client-side translation is ever wanted (e.g. to defend a backend that hasn't shipped the mapping), it must target the low-level `values` API path — not `set_cell_range` — and be covered by a **real** write+readback test, not dry-run.

PR #1526, opened 2026-06-21
docs(lark-task): default search scope to current user when subject is unspecified

### LINKED ISSUE #1479: 关于 lark-task skill 的默认搜索范围的讨论
## 问题描述
飞书lark-task这个 skill 某些场景会出现误读数据，当 query = “看一下名字包含‘测试’的任务”，它默认查询用户有查看权限的任务列表，大部分是公司其他同事的，而不是用户自己分内的任务，所以我们希望能在skill中增加规则：“当用户query没有明确查询主体时，默认为用户自己的任务”也许会更好一些。

## 复现query（可以更换引号内的关键词多试几次）：
- 搜索一下标题包含“会议”的任务
- 搜索一下标题包含“测试”的任务
- 搜索一下标题包含“开发”的任务

yyyCode:
## Summary
When a user searches tasks by keyword without explicitly specifying whose tasks to search (e.g., "搜索标题包含'测试'的任务"), the agent now defaults to the current user's own tasks instead of returning all visible tasks across the organization. This prevents surfacing irrelevant tasks from other colleagues.

## Changes
- Add a "默认搜索范围" rule to `skills/lark-task/SKILL.md` that instructs agents to prefer `+get-my-tasks --query` or `+search` with user-scoped filters when no explicit search subject is specified.
- Update the Workflow section in `skills/lark-task/references/lark-task-search.md` to include a scope check step before executing the search command.

## Test Plan
- [ ] Unit tests pass
- [x] Manual local verification confirms the `lark-cli skills read lark-task` flow works as expected

Manual verification:
- `make build`
- `lark-cli skills read lark-task | rg "默认搜索范围" -n`
- `lark-cli skills read lark-task references/lark-task-search.md | rg "Scope check" -n`

## Related Issues
- Closes https://github.com/larksuite/cli/issues/1479

## Summary by CodeRabbit

* **Documentation**
  * Clarified task search guidance for cases where a search query is provided without a specified scope.
  * Search behavior now defaults to the current user’s own tasks unless a broader scope is explicitly requested.
  * Updated instructions to use broader search only when the request clearly asks for all tasks, team tasks, or other users’ tasks.


yyyCode:
recheck

yyyCode:
@xoyowade If you could find the time to review this pull request, I would be very grateful

tengchengwei:
@yyyCode 

Thanks a lot for bringing up this suggestion and sparking this discussion — we talked through this design tradeoff internally and wanted to share our thinking.

The +search shortcut is meant to mirror the client-side search capability. In the official Lark Task client (both mobile and web), searching by keyword does not implicitly restrict results to the current user's own tasks — it searches across all tasks visible to the user, which is what users naturally expect from a "search" action.

By baking this implicit --assignee <me> filter into the skill, we'd be silently narrowing the result set every time the subject is unspecified. This can easily cause false negatives : when a user says "search for tasks about 'release'", they may genuinely be looking for a teammate's task or a team-wide task, and the implicit filter would hide those results without any indication. This is a behavior mismatch with the client and would surprise users (and agents) who expect search to behave like search.

If we want agents to default to "my tasks only" in ambiguous cases, I think the right place to encode that preference is at the higher prompt/orchestration layer (e.g., in the agent's system prompt or in caller-side instructions), not inside the +search skill itself. For example, a prompt can explicitly say: "When the user asks to search tasks without specifying scope, prefer +get-my-tasks --query <kw> to search only tasks assigned to me, unless the user explicitly asks for broader results." This keeps the CLI shortcut deterministic and aligned with the client, while still allowing agents to apply scoping behavior when desired — but as an explicit choice, not a hidden default.

yyyCode:
@tengchengwei 
Thanks for the thoughtful discussion — totally agreed that +search should stay deterministic and mirror the client behavior.

Updated the implementation accordingly:

SKILL.md — the default-scope rule is now an explicit routing choice: when scope is ambiguous, agents should prefer +get-my-tasks --query (inherently scoped to "my tasks") rather than adding an implicit --assignee to +search.
lark-task-search.md — reverted the scope-check step completely. +search stays clean and aligned with the client.
This way the "default to my tasks" preference is encoded at the routing layer as an intentional command selection, not as a hidden filter inside +search.

tengchengwei:
@yyyCode 

Thanks for the thoughtful discussion — one point I’d add: keyword-based task searches should prioritize `+search`. Its tokenization, full-text matching and relevance ranking are far more robust, and search quality is the core experience for keyword lookup scenarios. Falling back to `+get-my-tasks --query` would degrade retrieval performance for this very common use case.

I also don't support adding `--assignee=me` by default to narrow the scope. Implicit assignee filtering creates clear false-negative bad cases: users may be looking for collaborative or team-wide tasks, and silently dropping those results without user awareness would break the search expectations set by the official client.

If we do want a narrower default for ambiguous queries, that preference belongs in the agent's system prompt or in caller-side instructions, rather than altering the command routing within the skill itself.

PR #1531, opened 2026-06-22
docs(lark-base): document signature field (attachment style variant)

Neseria:
## What

Documents the new **Signature** field in the `lark-base` skill. Signature is not a top-level field type — it is the **attachment field's `style.type` variant**: `{"type":"attachment","style":{"type":"signature"}}`. Underneath it is `FieldType.Attachment` (17) distinguished only by `FieldUIType=Signature`.

**Signature cells are read-only.** Write, replace and clear are all rejected — there is no "you can clear it but not write it" special case to remember. Nothing errors: the field lands in the response's `ignored_fields` with `reason: SIGNATURE_READONLY`, and the rest of the request goes through. Signing happens in the Lark client.

## Scope: 3 reference files, +13/-2. `SKILL.md` untouched.

A signature field **reports itself as `type: "attachment"`** — `mapAttachmentFieldToOpenAPI` emits `type: attachment` unconditionally, the cell is byte-identical to an attachment cell, and the filter / lookup / workflow paths never consult `isSignatureField`. So every rule already written for `attachment` applies to signature automatically, and restating it in `data-query.md` / `field-lookup.md` / `workflow-schema.md` buys zero behavioural change while creating three places to drift. Those notes were in an earlier revision and have been dropped. For reference, adding the whole `button` field type (#2437) touched exactly one reference file.

What is left is only what the `attachment` rules do **not** already cover:

- **`field-schema.md` §2** — the 速查表 row for `attachment` said supplementary fields were `无`, which is now false.
- **`field-schema.md` §3.12** — the canonical creation payload, plus the read-only cell rule covering all three commands (`+record-*`, `+record-upload-attachment`, `+record-remove-attachment`). `+record-upload-attachment` gets its own sentence: it uploads the media *before* the append call drops the field, and that media is not rolled back, so a clean exit is not evidence the signature landed.
- **`field-create.md`** — `signature` is absent from `AgentFieldType` *and* from the alias map, so `{"type":"signature"}` hard-fails and the natural retry `{"type":"attachment"}` **succeeds as a plain attachment**. Silent wrong outcome, so it needs a pointer at the point of creation.
- **`field-update.md`** — `PUT` is a full overwrite and `style` defaults to `plain`, so omitting it silently downgrades the column while `type` stays `attachment` the whole time. **There is no safe self-service recovery**: re-`PUT`ing `style.type:"signature"` clears the column's attachment values, and recreate-and-migrate cannot work either because signature cells are not writable through the API. The rule is to stop and report the data risk instead. The workflow section's blanket "not changing `type` is relatively safe" now carries this exception — the type-change whitelist below it only governs `type` transitions and never sees this one.

## Notes

- Rebased onto `main` @ `20ef0af8a`. An earlier revision documented the write restriction in `references/lark-base-cell-value.md`; #2529 deleted that file and inlined the CellValue protocol into `SKILL.md`, so the rule now lives in `field-schema.md` §3.12 and this PR deliberately does not touch the `SKILL.md` index.
- **No CLI code change is needed** for this PR to be correct. `validateFieldCreate` does not whitelist field types or style keys — it only checks the formula/lookup guide ack and forwards the JSON verbatim. A separate follow-up could make `+record-upload-attachment` / `+record-remove-attachment` short-circuit on a signature field: both already fetch the field definition and type-check it before touching any media (`shortcuts/base/record_upload_attachment.go:273` and `:320`), so the check has everything it needs at the point where the orphan upload happens today.
- Server side: `bitable-mcp` MR !775 implements the `style.type` mapping and the read-only cell enforcement.

### DIFF
--- a/skills/lark-base/references/lark-base-field-create.md
+++ b/skills/lark-base/references/lark-base-field-create.md
@@ -54,7 +54,7 @@ POST /open-apis/base/v3/bases/:base_token/tables/:table_id/fields
   - `link`：必须有 `link_table`，可选 `bidirectional`、`bidirectional_link_field_name`。
   - `formula`：必须有 `expression`；先读 formula guide，再创建。
   - `lookup`：必须有 `from`、`select`、`where`；先读 lookup guide，再创建。
-  - `attachment`：签字字段没有 `"type":"signature"`，用 `style.type:"signature"` 创建；见 [Field Schema](lark-base-field-schema.md) §3.12。
+  - `attachment`：签字字段没有 `"type":"signature"`，用 `style.type:"signature"` 创建；见 [Field Schema](lark-base-field-schema.md) §3.12。若接口拒绝该 style，停止并报告当前环境尚未支持；不要删掉 `style` 重试，否则会创建普通附件字段。
 
 **正确（base +field-create）**
 
--- a/skills/lark-base/references/lark-base-field-create.md
+++ b/skills/lark-base/references/lark-base-field-create.md
@@ -54,6 +54,7 @@ POST /open-apis/base/v3/bases/:base_token/tables/:table_id/fields
   - `link`：必须有 `link_table`，可选 `bidirectional`、`bidirectional_link_field_name`。
   - `formula`：必须有 `expression`；先读 formula guide，再创建。
   - `lookup`：必须有 `from`、`select`、`where`；先读 lookup guide，再创建。
+  - `attachment`：签字字段没有 `"type":"signature"`，用 `style.type:"signature"` 创建；见 [Field Schema](lark-base-field-schema.md) §3.12。
 
 **正确（base +field-create）**
 
--- a/skills/lark-base/references/lark-base-field-schema.md
+++ b/skills/lark-base/references/lark-base-field-schema.md
@@ -40,7 +40,8 @@
 | `formula` | `type` `name` `expression` | 无 |
 | `lookup` | `type` `name` `from` `select` `where` | `aggregate` |
 | `auto_number` | `type` `name` | `style.rules` |
-| `attachment` / `location` / `checkbox` | `type` `name` | 无 |
+| `attachment` | `type` `name` | `style.type`（`plain` / `signature`） |
+| `location` / `checkbox` | `type` `name` | 无 |
 | `button` | `type` `name` `button_config.title` | 无 |
 
 所有类型都可额外传 `description`；上表的“常见补充字段”只列类型特有配置。
@@ -417,6 +418,14 @@
 ```json
 { "type": "attachment", "name": "附件" }
 ```
+
+签字字段是附件的 `style.type` 变体，没有 `"type":"signature"` 这个顶层类型（不传 `style` 按 `plain`）：
+
+```json
+{ "type": "attachment", "name": "店长确认签字", "style": { "type": "signature" } }
+```
+
+签字单元格**只读**：写入、替换、清空都不生效，签名只能由用户在飞书客户端完成。记录接口和 `+record-upload-attachment` / `+record-remove-attachment` 都不报错，只把该字段放进返回的 `ignored_fields`（`SIGNATURE_READONLY`），别当成成功。读取和下载与普通附件一致。
 
 ```json
 { "type": "location", "name": "位置" }
--- a/skills/lark-base/references/lark-base-field-schema.md
+++ b/skills/lark-base/references/lark-base-field-schema.md
@@ -425,7 +425,7 @@
 { "type": "attachment", "name": "店长确认签字", "style": { "type": "signature" } }
 ```
 
-签字单元格**只读**：写入、替换、清空都不生效，签名只能由用户在飞书客户端完成。记录接口和 `+record-upload-attachment` / `+record-remove-attachment` 都不报错，只把该字段放进返回的 `ignored_fields`（`SIGNATURE_READONLY`），别当成成功。读取和下载与普通附件一致。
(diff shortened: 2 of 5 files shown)


skills/lark-base/references/lark-base-cell-value.md
```diff
@@ -125,6 +125,18 @@
 - 删除附件：使用 `lark-cli base +record-remove-attachment --record-id <record_id> --field-id <field_id> --file-token <file_token> --yes`；可重复 `--file-token` 一次删除同一单元格里的多个附件。
 - 下载附件：使用 `lark-cli base +record-download-attachment --record-id <record_id> --file-token <file_token> --output <dir>`；不传 `--file-token` 时下载整行所有附件，也可重复 `--file-token` 只下载指定附件。Base 附件必须用这个命令下载，用其他下载入口可能失败。
 
+### 2.10 signature（签字：只读 / 可清空，不能写入）
+
+签字字段底层是附件的 `style:signature` 变体，单元格为一张签名图片（`image/png`），由用户在飞书客户端完成签名。**命令侧不能写入新的签名图**：在记录写入接口（`+record-upsert` / `+record-batch-create` / `+record-batch-update`）里给签字字段塞非空值会被拒绝并报错；用 `+record-upload-attachment` 上传到签字字段时，该字段会被跳过并出现在返回的 `ignored_fields` 里（不会上传）。
```
zhouyue-bytedance: 拦截签字字段写入的逻辑，在哪里

skills/lark-base/references/lark-base-field-create.md
```diff
@@ -29,6 +29,11 @@ lark-cli base +field-create \
   --base-token <base_token> \
   --table-id <table_id> \
   --json '{"name":"负责人","type":"user","multiple":false,"description":"用于标记记录的直接负责人；协作约定可参考[团队字段约定](https://example.com/field-spec)"}'
+
+lark-cli base +field-create \
+  --base-token <base_token> \
+  --table-id <table_id> \
+  --json '{"name":"店长确认签字","type":"attachment","style":{"type":"signature"}}'
```
zhouyue-bytedance: 这个得解释一下，这样写代表“签字”这个字段类型

zhouyue-bytedance: 哦，我看103行有，不过离这里太远了

skills/lark-base/references/lark-base-field-create.md
```diff
@@ -95,6 +100,7 @@ POST /open-apis/base/v3/bases/:base_token/tables/:table_id/fields
 
 - ⚠️ 这是写入操作，执行前必须确认。
 - ⚠️ 当 `type` 是 `formula` 或 `lookup` 时，先读对应 guide，再创建。
+- ⚠️ 签字字段用 `{"type":"attachment","style":{"type":"signature"}}` 创建（附件的签字变体），不是独立顶层类型；创建后可用 `+field-get` 确认返回的 `style.type` 为 `signature`，若仍是 `plain`（普通附件），说明当前环境暂未开放签字字段能力。签字字段不能通过命令写入签名图，见 [lark-base-cell-value.md](lark-base-cell-value.md)。
```
zhouyue-bytedance: 附件字段的style.type=plain吗？

skills/lark-base/references/lark-base-field-create.md
zhouyue-bytedance: 要不这里不用加这个例子了，签字字段也不是很高频的场景

Neseria: 好的我改一下，去掉这个例子

skills/lark-base/references/lark-base-field-create.md
Neseria: 是的，也兼容没有 style 字段的场景（默认附件）

skills/lark-base/references/lark-base-cell-value.md
Neseria: Skill 里面提及不能写入，真正拦截的逻辑是写在服务端代码的

skills/lark-base/references/lark-base-field-create.md
```diff
@@ -54,6 +54,7 @@ POST /open-apis/base/v3/bases/:base_token/tables/:table_id/fields
   - `link`：必须有 `link_table`，可选 `bidirectional`、`bidirectional_link_field_name`。
   - `formula`：必须有 `expression`；先读 formula guide，再创建。
   - `lookup`：必须有 `from`、`select`、`where`；先读 lookup guide，再创建。
+  - `attachment`：签字字段没有 `"type":"signature"`，用 `style.type:"signature"` 创建；见 [Field Schema](lark-base-field-schema.md) §3.12。若接口拒绝该 style，停止并报告当前环境尚未支持；不要删掉 `style` 重试，否则会创建普通附件字段。
```
zhouyue-bytedance, line 57: 为啥要特意说：没有 `"type":"signature"`

skills/lark-base/references/lark-base-field-schema.md
```diff
@@ -418,6 +419,14 @@
 { "type": "attachment", "name": "附件" }
 ```
 
+签字字段是附件的 `style.type` 变体，没有 `"type":"signature"` 这个顶层类型（不传 `style` 按 `plain`）：
```
zhouyue-bytedance, line 422: 同上，为啥特意说：没有 `"type":"signature"` 这个顶层类型

zhouyue-bytedance (approved):

kongenpei (approved):

PR #1533, opened 2026-06-22
fix(sheets): mark +delete-dimension high-risk-write

### LINKED ISSUE #1532: sheets +delete-dimension is missing the high-risk-write confirmation gate
## Problem

`sheets +delete-dimension` deletes entire rows or columns — an irreversible structural change — but is annotated `Risk: "write"`, so it does **not** trigger the framework's high-risk confirmation gate (it never requires `--yes`).

Its modern replacement `+dim-delete` and the sibling `+delete-sheet` are both `high-risk-write`. The deprecated `+delete-dimension` alias was left at `write`, so it is inconsistent with the rest of the destructive sheets commands and can delete data with no gate.

## Current behavior

- `lark-cli sheets +delete-dimension --help` shows `Risk: write` (no `--yes` flag).
- `lark-cli sheets +dim-delete --help` shows `Risk: high-risk-write`.
- Code: `shortcuts/sheets/backward/lark_sheets_row_column_management.go` → `SheetDeleteDimension.Risk = "write"`.

## Impact

Agents/LLMs that decide whether to ask the human based on the schema `risk` field will run row/column deletion **without** the human-confirmation protocol that `skills/lark-shared/SKILL.md` mandates for `high-risk-write`. The deprecated alias still works, so this is reachable today.

## Expected

Mark `+delete-dimension` as `high-risk-write` so it returns a structured `confirmation_required` error (exit code 10) unless `--yes` is passed — matching `+dim-delete` and `+delete-sheet`. `--dry-run` should still preview without `--yes`.

Precedent: #936 (`fix: mark base field update high risk`).

hehanlin1996:
## Summary

`sheets +delete-dimension` deletes entire rows/columns (irreversible) but was annotated `Risk: "write"`, so it never triggered the high-risk confirmation gate. Its modern replacement `+dim-delete` and the sibling `+delete-sheet` are both `high-risk-write`. This aligns the deprecated alias with them.

## Changes

- `SheetDeleteDimension.Risk`: `"write"` → `"high-risk-write"`, so without `--yes` it returns a structured `confirmation_required` error (exit 10); `--yes` executes; `--dry-run` still previews without confirmation.
- Added a metadata test locking the risk level (regression guard).
- Added gate-behavior tests: without `--yes` blocked + API not called / with `--yes` executes / `--dry-run` bypasses.
- Updated the existing execute tests to pass `--yes` under the gate.

## Test Plan

- [x] Unit tests pass: `go test ./shortcuts/sheets/...`
- [x] Manual local verification confirms the `lark-cli sheets +delete-dimension` flow works as expected:
  - `--help` shows `Risk: high-risk-write` and `--yes`
  - without `--yes` → exit 10, `confirmation_required`, no network call
  - `--dry-run` → previews the DELETE, no `--yes` needed
  - `--yes` → passes the gate and reaches the API

## Related Issues

- Closes larksuite/cli#1532

## Summary by CodeRabbit

* **New Features**
  * The `+delete-dimension` command now uses a higher-risk safety level and requires an explicit `--yes` confirmation to proceed, since row/column deletion is irreversible.
  * Use `--dry-run` to preview the DELETE request and response output without triggering the confirmation gate.
* **Tests**
  * Updated and added coverage to verify the new risk level, blocking behavior without `--yes`, successful execution with `--yes`, and confirmation bypass during `--dry-run`.

### DIFF
--- a/shortcuts/sheets/backward/lark_sheets_row_column_management.go
+++ b/shortcuts/sheets/backward/lark_sheets_row_column_management.go
@@ -309,9 +309,11 @@ var SheetDeleteDimension = common.Shortcut{
 	Service:     "sheets",
 	Command:     "+delete-dimension",
 	Description: "Delete rows or columns",
-	Risk:        "write",
-	Scopes:      []string{"sheets:spreadsheet:write_only", "sheets:spreadsheet:read"},
-	AuthTypes:   []string{"user", "bot"},
+	// Deleting rows/columns is an irreversible structural change, so it
+	// requires the same confirmation gate (--yes) as +delete-sheet.
+	Risk:      "high-risk-write",
+	Scopes:    []string{"sheets:spreadsheet:write_only", "sheets:spreadsheet:read"},
+	AuthTypes: []string{"user", "bot"},
 	Flags: []common.Flag{
 		{Name: "url", Desc: "spreadsheet URL"},
 		{Name: "spreadsheet-token", Desc: "spreadsheet token"},
--- a/shortcuts/sheets/backward/lark_sheets_sheet_dimension_test.go
+++ b/shortcuts/sheets/backward/lark_sheets_sheet_dimension_test.go
@@ -6,6 +6,7 @@ package backward
 import (
 	"context"
 	"encoding/json"
+	"net/http"
 	"strconv"
 	"strings"
 	"testing"
@@ -911,6 +912,7 @@ func TestSheetDeleteDimensionExecuteSuccess(t *testing.T) {
 		"--dimension", "ROWS",
 		"--start-index", "3",
 		"--end-index", "7",
+		"--yes",
 		"--as", "user",
 	}, f, stdout)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


hehanlin1996:
Thanks @coderabbitai — addressed in fe84b955. The two error-path tests now assert typed metadata via `errs.ProblemOf` instead of nil/substring checks:
- confirmation path → `Category=confirmation` / `Subtype=confirmation_required`
- API-error path → `Category=api` / `Code=90001`

The success and `--dry-run` paths already assert behavior (executed/preview) so they're left as-is.

hehanlin1996:
recheck

hehanlin1996:
recheck

xiongyuanwen-byted:
Thanks for the PR and the thorough test coverage!

Same note as on #1532: `+delete-dimension` is a deprecated shortcut that we plan to remove — it is kept only as a temporary backward-compat alias for `+dim-delete`. The modern replacement `+dim-delete` is already annotated `high-risk-write`, so it already returns the structured `confirmation_required` error (exit 10) unless `--yes` is passed, with `--dry-run` still previewing.

So rather than adding new code and tests to the deprecated alias, the recommended path is to switch to `+dim-delete`, which already has the high-risk-write gate. Since the alias is on its way out, we are inclined not to invest further in it.

PR #1606, opened 2026-06-26
feat(sheets): use office_sheet_file parent_type for imported office spreadsheets

xiongyuanwen-byted:
## Summary
Image uploads to a spreadsheet hard-coded `parent_type=sheet_image` at every entry point. Imported "office" spreadsheets carry a token prefixed with `fake_office_`, for which the drive backend requires `parent_type=office_sheet_file`. This PR funnels the parent_type selection through a single sheets-domain helper so the rule lives in one place and every image-upload path stays consistent.

## Changes
- Add `sheetMediaParentType(token)` in the sheets domain: returns `office_sheet_file` for `fake_office_`-prefixed tokens, otherwise `sheet_image`. Single source of truth for the mapping.
- Add an `uploadSheetImage(...)` collector that builds the `DriveMediaUploadAllConfig` (including parent_type) once, replacing the per-call-site hand-rolled configs.
- Route both main-domain image entries through the collector — float-image local upload and `+cells-set-image` — covering Execute **and** the dry-run preview body/desc.
- Cover the `backward` `+media-upload` entry: single-part, multipart (>20MB), and both dry-run bodies. `backward` is a separate package and an intentional verbatim mirror of `shortcuts/sheets/`, so it keeps its own copy of the helper rather than importing the main domain (avoids a cross-package dependency and preserves the mirror).
- Leave the shared `common.UploadDriveMediaAllTyped` upload layer untouched — the `fake_office_` rule is sheets-specific and must not leak into mail/slides/doc/drive/base.

## Test Plan
- [x] Unit tests pass (`go test ./shortcuts/sheets/...`)
- [x] New tests: pure-function `TestSheetMediaParentType` (6 cases incl. prefix-only and mid-string non-match), main-domain dry-run `TestCellsSetImage_DryRunOfficeParentType`, and `backward` real-multipart-body `TestSheetMediaUploadExecuteOfficeParentType` asserting `office_sheet_file` + `parent_node` actually go out on the wire
- [x] `gofmt` clean; `golangci-lint run --new-from-rev=HEAD` reports 0 new issues

## Related Issues
- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sheet image and media uploads to automatically choose the correct Drive upload destination type for native vs imported/office-style spreadsheets, with `parent_node` matching the provided spreadsheet token.
  * Ensured consistency between dry-run and execute for both floating images and embedded cell image uploads.

* **Tests**
  * Added unit and e2e coverage for office-style uploads, including multipart form `parent_type`/`parent_node` assertions and dry-run validation.
  * Added execute-path tests and local file-open error handling coverage.

### DIFF
--- a/shortcuts/sheets/backward/lark_sheets_float_images.go
+++ b/shortcuts/sheets/backward/lark_sheets_float_images.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/larksuite/cli/errs"
 	"github.com/larksuite/cli/extension/fileio"
@@ -14,7 +15,25 @@ import (
 	"github.com/larksuite/cli/shortcuts/common"
 )
 
-const sheetImageParentType = "sheet_image"
+// Drive media parent_type values for uploading an image into a spreadsheet.
+// Native spreadsheets use "sheet_image"; imported "office" spreadsheets carry a
+// synthetic token prefixed with "fake_office_" and the backend requires
+// "office_sheet_file" instead.
+const (
+	sheetImageParentType      = "sheet_image"
+	officeSheetFileParentType = "office_sheet_file"
+	fakeOfficeTokenPrefix     = "fake_office_"
+)
+
+// sheetMediaParentType returns the drive media parent_type to use when
+// uploading an image whose parent_node is spreadsheetToken, mapping the
+// "fake_office_" imported-spreadsheet token prefix to "office_sheet_file".
+func sheetMediaParentType(spreadsheetToken string) string {
+	if strings.HasPrefix(spreadsheetToken, fakeOfficeTokenPrefix) {
+		return officeSheetFileParentType
+	}
+	return sheetImageParentType
+}
 
 var SheetMediaUpload = common.Shortcut{
 	Service:     "sheets",
@@ -49,7 +68,7 @@ var SheetMediaUpload = common.Shortcut{
 				POST("/open-apis/drive/v1/medias/upload_prepare").
 				Body(map[string]interface{}{
 					"file_name":   fileName,
-					"parent_type": sheetImageParentType,
+					"parent_type": sheetMediaParentType(parentNode),
 					"parent_node": parentNode,
 					"size":        "<file_size>",
 				}).
@@ -71,7 +90,7 @@ var SheetMediaUpload = common.Shortcut{
 			POST("/open-apis/drive/v1/medias/upload_all").
 			Body(map[string]interface{}{
 				"file_name":   fileName,
-				"parent_type": sheetImageParentType,
+				"parent_type": sheetMediaParentType(parentNode),
 				"parent_node": parentNode,
 				"size":        "<file_size>",
 				"file":        "@" + filePath,
@@ -141,13 +160,14 @@ func resolveSheetMediaUploadParent(runtime *common.RuntimeContext) (string, erro
 }
 
 func uploadSheetMediaFile(runtime *common.RuntimeContext, filePath, fileName string, fileSize int64, parentNode string) (string, error) {
+	parentType := sheetMediaParentType(parentNode)
 	if fileSize <= common.MaxDriveMediaUploadSinglePartSize {
 		pn := parentNode
 		return common.UploadDriveMediaAllTyped(runtime, common.DriveMediaUploadAllConfig{
 			FilePath:   filePath,
 			FileName:   fileName,
 			FileSize:   fileSize,
-			ParentType: sheetImageParentType,
+			ParentType: parentType,
 			ParentNode: &pn,
 		})
 	}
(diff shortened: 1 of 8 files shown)


shortcuts/sheets/backward/lark_sheets_float_images.go
```diff
@@ -49,7 +68,7 @@ var SheetMediaUpload = common.Shortcut{
 				POST("/open-apis/drive/v1/medias/upload_prepare").
 				Body(map[string]interface{}{
 					"file_name":   fileName,
-					"parent_type": sheetImageParentType,
+					"parent_type": sheetMediaParentType(parentNode),
```
fangshuyu-768, line 71: This fixes the multipart dry-run preview, but the small-file dry-run branch just below still hard-codes `sheetImageParentType` in the `upload_all` body. For `fake_office_...` spreadsheet tokens, that preview will show `sheet_image` even though execute sends `office_sheet_file`, which makes dry-run output misleading. Could we use `sheetMediaParentType(parentNode)` in the small-file branch as well, and add a small-file dry-run test for a `fake_office_...` token?

wittam-01, line 71: The multipart dry-run now uses `sheetMediaParentType(parentNode)`, but the small-file `upload_all` dry-run branch just below still emits `parent_type: sheetImageParentType`. For `fake_office_...` tokens that preview remains wrong even though Execute uses `office_sheet_file`, so agents/users following `--dry-run` will copy the bad request body.

Please mirror this helper in the `upload_all` body as well and add a `fake_office_` small-file dry-run test.

shortcuts/sheets/sheet_media_parent_type_test.go
```diff
@@ -0,0 +1,164 @@
...
+// error and no upload request is attempted.
+func TestUploadSheetImage_FileOpenError(t *testing.T) {
+	runtime, _ := newSheetMediaTestRuntime(t)
+	cmdutil.TestChdir(t, t.TempDir())
+	// No upload stub is registered: a file-open failure must short-circuit
+	// before any request reaches the wire.
+	_, err := uploadSheetImage(runtime, "shtcnTOK123", "missing.png", "missing.png", 1)
+	if err == nil {
+		t.Fatal("expected error for missing file, got nil")
+	}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in 903c60ae. Strengthened `TestUploadSheetImage_FileOpenError` to assert typed metadata via `errs.ProblemOf` (category=validation, subtype=invalid_argument) and `errors.Is(err, fs.ErrNotExist)` for cause preservation. Endpoint-not-hit is now strictly implied by the same metadata: no stub is registered, so any real `upload_all` POST would surface from `httpmock` as a non-validation network error and trip the category assertion. Tried a Reusable+OnMatch stub first, but `TestFactory`'s `t.Cleanup(reg.Verify)` flags an unmatched stub as a failure and can't be opted out, so the implicit form is what's left.

shortcuts/sheets/sheet_media_parent_type_test.go
```diff
@@ -0,0 +1,164 @@
...
+	reader := multipart.NewReader(bytes.NewReader(stub.CapturedBody), params["boundary"])
+	body := sheetMediaCapturedMultipart{Fields: map[string]string{}, Files: map[string][]byte{}}
+	for {
+		part, err := reader.NextPart()
+		if err != nil {
+			break
+		}
+		buf := new(bytes.Buffer)
+		_, _ = buf.ReadFrom(part)
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in 903c60ae. `decodeSheetMediaMultipartBody` now treats only `io.EOF` as the loop terminator and `t.Fatalf`s on any other `NextPart()` or `ReadFrom()` error, so a malformed/truncated body can no longer produce a partial captured payload.

shortcuts/sheets/backward/lark_sheets_float_images.go
xiongyuanwen-byted: Done in 903c60ae. The small-file `upload_all` branch now also uses `sheetMediaParentType(parentNode)`, so dry-run and Execute agree on `office_sheet_file` for `fake_office_`-prefixed tokens. Added `TestSheetMediaUploadDryRunSmallFileOfficeParentType` to pin it — asserts the output contains `"office_sheet_file"` and not `"sheet_image"`. Thanks for catching this.

xiongyuanwen-byted: Done in 903c60ae. The small-file `upload_all` body now mirrors the multipart branch via `sheetMediaParentType(parentNode)`, and `TestSheetMediaUploadDryRunSmallFileOfficeParentType` covers the `fake_office_` small-file dry-run. Thanks for the catch.

fangshuyu-768 (approved):

PR #1704, opened 2026-07-01
Retry E2E retryable service errors by default

yxzhaao:
## What changed

- Make the CLI E2E `RunCmd` helper automatically retry structured service errors when the CLI returns `error.retryable: true`.
- Split the actual process execution into `runCmdOnce` so `RunCmdWithRetry` remains a custom retry entrypoint and does not nest the default `RunCmd` retry policy.
- Share JSON payload extraction between retryable-error detection and cleanup-error suppression, covering stderr that contains progress text before the JSON envelope.
- Add fake CLI coverage for retryable errors, non-retryable errors, and the no-nested-retry contract.

## Why

Live E2E ownership is now distributed across feature owners. Transient server contention, such as Drive delete returning `1061045` with `retryable: true`, should be handled by the framework by default instead of depending on every test author remembering to call `RunCmdWithRetry`.

### DIFF
--- a/tests/cli_e2e/core.go
+++ b/tests/cli_e2e/core.go
@@ -26,6 +26,13 @@ const projectRootMarkerDir = "tests"
 const cliBinaryName = "lark-cli"
 const CleanupTimeout = 30 * time.Second
 
+var (
+	defaultRetryAttempts        = 4
+	defaultRetryInitialDelay    = 1 * time.Second
+	defaultRetryMaxDelay        = 6 * time.Second
+	defaultRetryBackoffMultiple = 2
+)
+
 func SkipWithoutUserToken(t *testing.T) {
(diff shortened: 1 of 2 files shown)


yxzhaao (commented):
What changed
Make the CLI E2E RunCmd helper automatically retry structured service errors when the CLI returns error.retryable: true.
Split the actual process execution into runCmdOnce so RunCmdWithRetry remains a custom retry entrypoint and does not nest the default RunCmd retry policy.
Share JSON payload extraction between retryable-error detection and cleanup-error suppression, covering stderr that contains progress text before the JSON envelope.
Add fake CLI coverage for retryable errors, non-retryable errors, and the no-nested-retry contract.

liangshuo-1 (approved):

PR #1720, opened 2026-07-02
docs: add lark slides generation constraints

ethan-zhx:
## Summary
Add a high-priority lark-slides skill constraint for PPT generation and template editing behavior, and remove the older built-in template-tool path to avoid conflicting template routing.

## Changes
- Add a CRITICAL rule covering PPT canvas bounds, image-rich generation, shape usage limits, background image prompt constraints, and template-preserving edit behavior.
- Remove the lark-slides template toolchain: template_tool.py, its tests, template catalog/index, and bundled XML template assets.
- Remove SKILL/planning references that routed natural-language template/theme/layout requests through template_tool.

## Test Plan
- [x] Unit tests pass
- [x] Manual local verification confirms the documentation diff is scoped as expected
- [x] Ran `git diff --check`
- [x] Confirmed no `template_tool` / `template-catalog` / `template-index` / `assets/templates` references remain under `skills/lark-slides`

## Related Issues
- None

## Summary by CodeRabbit

* **Documentation**
  * Updated quick reference and execution rules with stricter PPT constraints (960×540, keep content within page bounds) and emphasized “plan first, then write.”
  * Added a clearer workflow for using user-provided templates: preserve original layout and fonts, then only fine-tune text/images/shapes as needed.
* **New Features**
  * Improved background image prompt guidance to prevent any text from appearing in slide backgrounds.
* **Chores / Template Updates**
  * Removed multiple built-in slide deck templates from the available assets.


fangshuyu-768:
Thanks for tightening the slide-generation guidance. I think two parts here should be narrowed before merge:

1. The new template instruction says that when a user specifies a PPT template, the agent should import it through `lark-drive` and edit the resulting Lark Slides directly. That conflicts with the existing template flow a few lines below, which says natural-language template/theme/layout requests must go through `template_tool.py search -> summarize -> extract`. As written, a request like “use a product launch template” could be routed to `lark-drive` even when the user has not provided a PPTX/cloud-drive file. Could we scope this to “when the user provides/uploads a PPTX template file or cloud-drive PPT resource”, and explicitly leave natural-language template/theme/layout requests on the `template_tool` path?

2. “Use more generated images, search images as support, must be image-and-text rich” feels too strong as a global CRITICAL requirement. For technical, research, data, or accuracy-sensitive decks, forcing generated/searched images can introduce incorrect assets and also pulls against the existing asset-planning rule that only metadata should be planned before XML generation. Could we make this conditional, e.g. use generated/searched images when the user allows it and it helps the story, while preferring user-provided assets, charts, screenshots, whiteboard diagrams, or XML-native fallbacks when accuracy matters?

fangshuyu-768:
I found one blocking regression after the template toolchain removal.

`skills/lark-slides/scripts/xml_text_overlap_lint_test.py` still depends on `skills/lark-slides/assets/templates/*.xml` as required fixtures:

```py
TEMPLATES_DIR = Path(__file__).resolve().parents[1] / "assets" / "templates"
...
template_paths = sorted(TEMPLATES_DIR.glob("*.xml"))
self.assertTrue(template_paths)
```

This PR deletes the entire `skills/lark-slides/assets/templates/` directory, so the test now fails because `template_paths` is empty. I reproduced it locally from `skills/lark-slides/scripts` with:

```bash
python3 -m unittest xml_text_overlap_lint_test.py
```

Result:

```text
FAIL: test_xml_text_overlap_lint_accepts_all_template_xml_files
AssertionError: [] is not true
```

Could you either keep a small fixture XML set for this lint coverage, or update the test to use inline XML samples instead of the removed template directory?

fangshuyu-768 (approved):

PR #1760, opened 2026-07-06
docs: require word count validation for doc creation

fangshuyu-768:
## Summary
- make word-count validation a required gate when doc creation requests include explicit length constraints
- delegate the concrete validation procedure to lark-doc-word-stat.md instead of repeating it in the create workflow
- require final reporting of the target range, word_count, and pass/fail result when validation runs

## Testing
- Not run; documentation-only change.

## Summary by CodeRabbit

* **Documentation**
  * Updated the workflow to treat “专项校验” as a standard step only when users provide an explicit word-count requirement (e.g., a specific range or upper/lower limits).
  * If no clear word-count requirement is given, the check is now skipped.
  * When run, the documentation clarifies that results are only shown after word-count evaluation, including the target range, the computed word count, and a clear pass/fail conclusion.


fangshuyu-768:
Superseded by #1759, which now includes this word-count validation change together with the duplicate-title guidance.

PR #1761, opened 2026-07-06
Soften Drive cleanup visibility checks

yxzhaao:
## What changed

- Treat Drive cleanup visibility lag differently from delete failures: if `drive +delete` succeeds but meta remains visible after the post-delete wait, report a cleanup warning instead of failing the business E2E.
- Keep failed delete commands fatal when the resource still exists.
- Extend cleanup context to 90s and Drive delete visibility polling to 60s.
- Add `scripts/e2e_scope.sh` so CI resolves CLI E2E scope from changed files as `subset`, `full`, or `skip`.
- Wire both dry-run and live CLI E2E jobs to use resolved `dry_packages` / `live_packages` instead of always running every domain package.
- Add script-level coverage for domain subset, doc/docs expansion, shared-harness full fallback, runtime full fallback, docs-only skip, and skill-domain mapping.

## Why

Drive delete can be accepted by the server while metadata remains visible briefly because of eventual consistency. Teardown should not fail otherwise-passed business workflows, such as `TestDocs_UpdateWorkflow`, when the destructive cleanup command has already succeeded.

Live E2E should also stay focused on the business domains touched by a PR. Domain-specific changes now run only matching `tests/cli_e2e/<domain>` packages, while shared/runtime/workflow changes conservatively fall back to full coverage and docs-only changes skip CLI E2E.

## Validation

- `make script-test`
- `bash scripts/e2e_scope.test.sh`
- `bash scripts/ci-workflow.test.sh`
- `go test ./tests/cli_e2e ./tests/cli_e2e/drive -run TestDeleteDriveResourceAndVerify -count=1`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,12 +263,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          fetch-depth: 0
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
         with:
           go-version-file: go.mod
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: '3.x'
+      - name: Resolve CLI E2E scope
+        id: e2e_scope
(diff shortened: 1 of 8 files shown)


yxzhaao:
Superseded by #1709; moved the Drive cleanup and CLI E2E domain-selection changes into that PR.

PR #1799, opened 2026-07-08
feat: use remote scopes.json for login scope recommendations

dc-bytedance:
## Summary

`auth login` previously computed its recommended authorization scopes from a compiled-in local table, so different CLI versions could request different scope sets and drift from the platform's own scope computation. This PR switches the scope source to a remotely published per-brand `scopes.json` fetched at login time (~1s timeout, whole-file validation, silent fallback to the local computation on any failure), and removes the terminal interactive domain-selection page and the local auto-approve filter chain so `--recommend` is equivalent to `--domain all`.

## Changes

- Add `internal/auth/remote_scopes.go`: brand-addressed GET with ~1s timeout and whole-file validation — a valid file is used verbatim (including domains/scopes unknown to this CLI build); any failure (non-2xx, empty body, bad JSON, missing/empty `scopes`, a domain missing `user_scopes`, or a malformed scope string: fewer than two `:`-separated segments, an empty segment, or characters outside `[a-z0-9_.]`) falls back silently to the local computation
- `cmd/auth/login.go`: fetch remote scopes once per login and use them for domain validation, `--domain all` / bare-login / `--recommend` expansion, and per-domain scope selection; the `--scope`-only path never touches the catalog
- Remove `cmd/auth/login_interactive.go`, the auto-approve loader chain in `internal/registry/loader.go`, and the service-description getter orphaned by the interactive-page removal; drop the now-unused interactive-selection and no-flags-hint message strings in `cmd/auth/login_messages.go`
- Behavior changes (release-notes relevant): bare `auth login` in a non-TTY environment no longer fails fast — it initiates full-domain authorization and blocks up to ~10 minutes awaiting authorization (agent harnesses should use `--no-wait`); `--recommend` now requests the full domain scope set rather than the former auto-approve subset (the authorization page still shows every requested scope)

## Test Plan

- [x] `gofmt` clean on all changed files
- [x] `go build ./...` passed
- [x] `go vet ./cmd/... ./internal/...` passed
- [x] `go test ./...` unit suite passed on the rebased head
- [x] manual verification: fetched the production `scopes.json` for both brands; whole-file validation accepts them; per-domain diff against the local synthesis — 20/21 domains identical
- [ ] sandbox E2E / acceptance-reviewer: re-run via CI on the rebased head

## Related Issues

N/A


cmd/auth/login_result.go
```diff
@@ -215,9 +255,8 @@ func authorizationCompletePayload(openId, userName string, summary *loginScopeSu
 	}
 	if issue != nil {
 		payload["warning"] = map[string]interface{}{
-			"type":    "missing_scope",
-			"message": issue.Message,
-			"hint":    issue.Hint,
+			"type": "missing_scope",
+			"hint": summary.StatusMessage,
```
dc-bytedance: **`authorizationCompletePayload`: empty `status_message` leaves the JSON warning with no readable content, and dropping `message` is a consumer-visible break.**

1. When the token response does not carry `status_message` (older server deployments, or any path where the field is absent), the warning payload becomes `{"type":"missing_scope","hint":""}`. The `missing` scope list is still elsewhere in the payload, but the consumer loses all human-readable guidance. The text path already has a fallback for this case (it prints the missing-scope list when `StatusMessage` is empty) — the JSON path doesn't. Suggest: fall back to the previous locally generated hint when `summary.StatusMessage == ""`, and add a unit test pinning the empty-`status_message` JSON shape so it can't regress.

2. Removing the `message` field is a breaking change for any JSON consumer reading `.warning.message` today. Probably an acceptable break, but it should be stated in the PR body so it reaches the release notes.

cmd/auth/login_result.go
```diff
@@ -129,14 +146,43 @@ func emptyIfNil(s []string) []string {
...
 func writeLoginScopeBreakdown(errOut *cmdutil.IOStreams, msg *loginMsg, summary *loginScopeSummary) {
 	if summary == nil {
 		summary = &loginScopeSummary{}
 	}
-	fmt.Fprintf(errOut.ErrOut, msg.RequestedScopes, formatScopeList(summary.Requested, msg.NoScopes))
-	fmt.Fprintf(errOut.ErrOut, msg.NewlyGrantedScopes, formatScopeList(summary.NewlyGranted, msg.NoScopes))
+	fmt.Fprintln(errOut.ErrOut)
+	fmt.Fprintln(errOut.ErrOut, msg.GrantedScopes)
+	fmt.Fprintf(errOut.ErrOut, "  %s\n", formatScopeList(grantedRequestedScopes(summary), msg.NoScopes, msg.ScopeSeparator))
+	if summary.StatusMessage != "" {
```
dc-bytedance, line 157: **`writeLoginScopeBreakdown`: "not granted" header can sit on top of non-failure text.**

The function prints the not-granted header + `status_message` whenever `StatusMessage != ""`, even when `summary.Missing` is empty. If the server can return a non-empty `status_message` on a fully granted login (e.g. a success summary), the output would show a "not granted" header over success text.

If the contract guarantees `status_message` is only populated when some scopes were not granted, a short code comment stating that assumption here would keep someone from "fixing" it later; otherwise gate the header on `len(summary.Missing) > 0` (or use a neutral header when `Missing` is empty).

liangshuo-1 (approved):

PR #1801, opened 2026-07-08
feat(drive): Strengthen lark-drive high-risk write operations and read-only recognition boundaries.

yballul-bytedance:
## Summary

  Strengthen the lark-drive skill’s guardrails for high-risk write operations, including deletion, public permission changes, owner transfer, version deletion/rollback, and bulk move/overwrite/
  sync. Distinguish “target state” from “specific target plus execution confirmation.” Also define URL type/token recognition as a read-only responsibility of drive +inspect, and add common usage
  guidance for files list sorting, file.statistics, and file.view_records typed flags to reduce Drive command misuse and unauthorized execution.

## Changes

- skills/lark-drive/SKILL.md
    - Expanded the description to include scenarios for identifying link type, real token, and title.
    - Added two quick-decision rules:
        - URL type/token recognition should use lightweight judgment or +inspect, only for identification/disambiguation.
        - High-risk write operations must have both a specific target and explicit confirmation from the user in the current turn. Do not treat target-state requests such as “delete useless
          ones,” “open it to everyone,” or “overwrite/move these” as execution confirmation.

    - Changed the files list section from “must” to “read the reference before use and pass parameters according to the template,” avoiding overly rigid binding.
    - Clarified that file.statistics get and file.view_records list should use typed flags such as --file-token, --file-type, --page-size, and --viewer-id-type, instead of rewriting them as
      --params JSON.

- skills/lark-drive/references/lark-drive-delete.md
    - Clarified the meaning of “specific target”: a concrete URL/token, or a resource list confirmed item-by-item or in batch by the user. Candidates found through heuristic search are pending-
      confirmation targets.

    - Added a “pre-deletion gate” table requiring both a specific target and execution confirmation. If either is missing, list candidates and stop to wait for confirmation.

- skills/lark-drive/references/lark-drive-files-list.md
    - Added a “sort by time” section with templates for CreatedTime / EditedTime plus ASC / DESC.
    - Updated parameter rules:
        - Do not pass order_by / direction by default.
        - Use minimal --params only when sorting.
        - Add page_size / page_token only for pagination, recursive inventory, full export of a large directory, or follow-up pagination after the first page returns has_more=true.
        - Do not omit server-side sorting and then sort locally with scripts.

- skills/lark-drive/references/lark-drive-inspect.md
    - Expanded the scenario table: use drive +inspect when asking for resource type, canonical token, title, canonical URL, or when the path looks like docx / Base / Sheet / Slides / Drive folder
      but the task is only identification.

    - Clarified that identifying link type/token/title is a Drive inspect task. Do not replace it with body reading, Base table commands, or manual URL path extraction.

- skills/lark-drive/references/lark-drive-permission-guide.md
    - Added a “pre-public-permission-change gate” table: --yes is allowed only when all three are present: specific target, user-selected link_share_entity level, and current-turn execution
      confirmation.

    - Clarified that phrases like “open it up,” “share it with everyone,” or “let everyone view it” only express the target state, not the specific public scope. List available levels first, then
      stop and wait for the user’s choice.

## Test Plan

- [x] Applied the MR net patch with git apply --3way. SKILL.md required semantic merge to preserve the GitHub base’s existing “copy document / create duplicate” decision item; the other 4
  reference files applied cleanly.
- [x] git diff --check reported no whitespace errors.
- [x] git diff --cached --stat matches the MR file list: 5 files changed, 58 insertions(+), 4 deletions(-).
- [ ] Later, in an integration environment, regress the skill prompts for drive +delete, drive permission.public patch, drive files list, drive file.statistics get, drive file.view_records list,
  and drive +inspect, confirming that agents reliably hit the new gates and typed-flag guidance.
## Related Issues

Auto research task: 01KWW7ZHEWMRH9JSBMDAMSNJZ4

## Summary

* **Documentation**
  * Updated cloud-space guidance to correctly handle link type, real token, and title routing for Feishu/doubao 云空间, without relying on fallback URL fetching.
  * Expanded `drive +inspect` usage with typical scenarios and reinforced “don’t infer from page content/Base/manual path truncation” rules.
  * Added “deletion threshold” and “public permission change threshold” guardrails: high-risk write actions now require a specific target plus explicit, per-round confirmation.
  * Refined `drive files list` sorting/pagination and tightened `--params`/typed `--format json` examples for related native commands.


skills/lark-drive/SKILL.md
```diff
@@ -21,6 +21,8 @@ metadata:
 ## 快速决策
 
 - 用户要**复制文档 / 创建副本 / 另存为副本**时，使用 `lark-cli drive files copy`。先用 `lark-cli schema drive.files.copy --format json` 确认参数；如果来源是 wiki URL/token，先用 `lark-cli drive +inspect` 获取底层 `token` 和 `type`，不要把 wiki token 直接当 `file_token`。`params.file_token` 传源文档 token，`data.folder_token` 传目标文件夹 token，`data.name` 传副本名称，`data.type` 传源文件类型（如 `docx` / `sheet` / `bitable` / `slides`）。示例：`lark-cli drive files copy --params '{"file_token":"<DOC_TOKEN>"}' --data '{"folder_token":"<FOLDER_TOKEN>","name":"<COPY_NAME>","type":"docx"}'`。如返回 `confirmation_required`，按 `lark-shared` 高风险审批协议向用户确认后，在原命令末尾追加 `--yes` 重试。
+- 用户要**识别飞书 / doubao 云空间 URL 的类型和 token**时，可以先按 URL 路径形态做轻量判断；当路径已明确指向 docx / sheet / bitable / slides / file / folder 等资源时，可直接提取对应 token/type。传入 wiki URL、需要识别标题或 canonical URL、URL/token 有歧义，或后续操作依赖底层真实资源时，再使用 `lark-cli drive +inspect --url '<url>'` 进行识别。`+inspect` 失败后不要自动切到写接口继续尝试，先按错误提示处理权限、scope 或链接问题。`inspect` 仅用于识别/消歧，不作为所有操作的前置步骤。
```
fangshuyu-768: Could we avoid treating URL-shape parsing as canonical identity resolution here? This says agents can directly extract token/type when the path looks clear, while `lark-drive-inspect.md` says not to replace `+inspect` with manual path slicing for type/token/title. I’d limit lightweight URL checks to routing hints, and require `drive +inspect` for canonical type, real token, title, wiki unwrapping, or any follow-up operation that depends on the true resource identity.

skills/lark-drive/SKILL.md
```diff
@@ -21,6 +21,8 @@ metadata:
...
+- 高风险写操作（删除、公开权限修改、owner 转移、版本删除/回滚、批量移动/覆盖/同步）必须同时满足两个条件才执行：目标是具体 URL/token/资源列表，且用户在本轮明确确认执行该具体目标。用户只说“删除没用的文件”“开放/共享给大家”“改成开放”“覆盖/移动这些”只表示目标状态；先只读发现并列出候选或权限档位，停止等待用户确认。
```
fangshuyu-768: Please tighten this top-level high-risk write gate. As written, a concrete URL/token/list plus current-turn confirmation is enough for all listed operations, but several of them need operation-specific executable details before they are safe to run: public permission patches need the selected scope and field diff, owner transfers need the target owner/options, version rollback/delete needs the version id, and move/overwrite/sync needs destination/conflict policy. Since this quick-decision rule is read before the detailed references, it can accidentally override the stricter per-operation gates. I’d frame it as: resolved executable target(s) + required operation-specific parameters/plan + explicit confirmation.

fangshuyu-768:
One broader consistency concern: the new guardrails are valuable, but the quick-decision rules and the detailed references currently give agents different signals in a couple of places.

- For URL identity, `SKILL.md` says agents may directly extract token/type when the URL path looks clear, while `lark-drive-inspect.md` says link type/token/title recognition should not be replaced with manual URL slicing and should go through `drive +inspect`.
- For high-risk writes, `SKILL.md` frames the common gate as concrete URL/token/list plus current-turn confirmation, while the detailed permission/owner/version/sync guidance requires more operation-specific executable details before running the write.

Because agents usually read the quick-decision section first, these broader rules can accidentally loosen or contradict the stricter reference-level rules. I think the safest model is to make the quick-decision section act only as routing guidance, and leave canonical identity resolution plus executable write prerequisites to the detailed references.

skills/lark-drive/SKILL.md
wittam-01: 这里主要是担心所有操作都是用inspect，会多一次openapi调用，所以建议明确的url，可以直接提取type和token

wittam-01 (approved):

wittam-01 (approved):

wittam-01 (approved):

wittam-01 (approved):

PR #1818, opened 2026-07-09
fix(doc): reject unsafe str_replace code spans

### LINKED ISSUE #1680: docs +update str_replace strips <code> tags around pure punctuation
## Bug Description

When using `lark-cli docs +update --command str_replace`, the `<code>` inline tag is silently stripped if its content is pure punctuation (e.g. `::` or `.`).

The same markup is preserved correctly when using `--command block_replace`, so this appears to be specific to how the server parses the `str_replace` replacement content.

## Environment

- `lark-cli` version: `1.0.60`
- Installation method: `pnpm add -g @larksuite/cli`
- OS: macOS

## Reproduction

1. Insert a paragraph containing `<code>::</code>` and `<code>.</code>`:

```bash
lark-cli docs +update --doc "<doc_id>" --command block_insert_after \
  --block-id "<last_block_id>" \
  --content '<p>TEST <code>::</code> and <code>.</code></p>'
```

2. Run `str_replace` with the same content:

```bash
lark-cli docs +update --doc "<doc_id>" --command str_replace \
  --pattern 'TEST <code>::</code> and <code>.</code>' \
  --content 'TEST <code>::</code> and <code>.</code>'
```

## Dry-run request

```json
{
  "command": "str_replace",
  "content": "TEST <code>::</code> and <code>.</code>",
  "format": "xml",
  "pattern": "TEST <code>::</code> and <code>.</code>",
  "revision_id": -1
}
```

## Actual result

The paragraph becomes:

```xml
<p>TEST :: and .</p>
```

The `<code>` tags are stripped.

## Expected result

The `<code>` tags should be preserved:

```xml
<p>TEST <code>::</code> and <code>.</code></p>
```

## Workaround

Use `block_replace` instead of `str_replace` for content that contains punctuation-only code spans.

## Additional context

- The `lark-doc` skill (v2.0.0) documents `str_replace` as supporting inline tags like `<b>`, `<a>`, `<cite>`, `<latex>`, but does not warn about this edge case.
- This behavior makes it unsafe to edit technical documentation that uses `<code>::</code>` (e.g. C++ scope-resolution operators) or `<code>.</code>` (e.g. struct field access) via `str_replace`.

xu91102:
## Summary
Reject unsafe `docs +update --command str_replace` content that contains punctuation-only inline `<code>` spans, because the upstream str_replace path strips those tags while `block_replace` preserves them.

## Changes
- Detect punctuation-only XML `<code>` inline spans for `str_replace` replacement content.
- Return a typed validation error bound to `--content` with guidance to use `block_replace`.
- Add shortcut and CLI dry-run regression coverage.
- Document the `str_replace` edge case in the lark-doc update reference.

## Test Plan
- [x] `GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go test ./shortcuts/doc -count=1`
- [x] `GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go build -o ./lark-cli .`
- [x] `LARK_CLI_BIN=$PWD/lark-cli GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go test ./tests/cli_e2e/docs -run 'TestDocs_DryRunDefaultsToV2OpenAPI|TestDocs_UpdateRejectsStrReplacePunctuationOnlyInlineCode' -count=1`
- [x] `git diff --check`
- [x] conflict marker scan

## Related Issues
- Fixes #1680

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for document updates so punctuation-only inline code is rejected in cases where it could produce unreliable replacements.
  * Clarified error feedback to guide users toward a safer block-based replacement workflow.
* **Documentation**
  * Expanded best-practice guidance for rich-text replacements, including when to use block replacements instead of inline code updates.
* **Tests**
  * Added coverage for both rejected and accepted inline-code replacement scenarios, plus an end-to-end CLI check.

### DIFF
--- a/shortcuts/doc/docs_update_test.go
+++ b/shortcuts/doc/docs_update_test.go
@@ -207,6 +207,42 @@ func TestValidateUpdateV2RejectsInvalidReferenceMap(t *testing.T) {
 	}
 }
 
+func TestValidateUpdateV2RejectsStrReplacePunctuationOnlyInlineCode(t *testing.T) {
+	t.Parallel()
+
+	runtime := newUpdateShortcutTestRuntime(t, "", map[string]string{
+		"command": "str_replace",
+		"pattern": "TEST :: and .",
+		"content": `<p>TEST <code>::</code> and <code>.</code> and <code>ok</code></p>`,
+	})
+
+	err := validateUpdateV2(context.Background(), runtime)
+	assertValidationContract(t, err, errs.SubtypeInvalidArgument, "--content")
+	for _, want := range []string{
+		"punctuation-only <code>",
+		"docs_ai str_replace",
+		"block_replace",
+	} {
+		if !strings.Contains(err.Error(), want) {
+			t.Fatalf("validateUpdateV2() error missing %q: %v", want, err)
+		}
+	}
+}
+
+func TestValidateUpdateV2AllowsStrReplaceAlphanumericInlineCode(t *testing.T) {
+	t.Parallel()
+
+	runtime := newUpdateShortcutTestRuntime(t, "", map[string]string{
+		"command": "str_replace",
+		"pattern": "TEST fmt.Println",
+		"content": `<p>TEST <code>fmt.Println</code></p>`,
+	})
+
+	if err := validateUpdateV2(context.Background(), runtime); err != nil {
+		t.Fatalf("validateUpdateV2() error = %v", err)
+	}
+}
+
 func TestDocsUpdateRejectsLegacyFlags(t *testing.T) {
 	tests := []struct {
 		name     string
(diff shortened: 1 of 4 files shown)


fangshuyu-768:
Thank you for preparing this defensive validation and the accompanying tests and documentation. We have now re-tested the original #1680 scenario on lark-cli 1.0.87 against a live document. Both `<code>::</code>` and `<code>.</code>` were preserved after `docs +update --command str_replace` and a full XML fetch, with no warnings. Since the underlying issue is no longer reproducible, rejecting these otherwise valid inputs is no longer necessary. We are therefore closing this PR together with #1680. We appreciate the thoughtful contribution and the effort put into making the update path safer.

PR #1819, opened 2026-07-09
fix(doc): warn on markdown heading numbering loss

### LINKED ISSUE #1781: docs +fetch --doc-format markdown silently drops heading seq/seq-level (auto-numbering) attribute
## Problem

`docs +fetch --api-version v2 --doc-format markdown` silently drops the `seq`/`seq-level` attributes on heading blocks. When a document has native Feishu heading auto-numbering enabled (e.g. `<h1 seq="auto" seq-level="auto">`), the markdown export renders the heading as plain unnumbered text, with no indication that the heading is actually auto-numbered in the live document.

This is a real-world footgun: an agent (or human) reading only the markdown export has no way to tell "this heading has no numbering" apart from "this heading is auto-numbered but the export doesn't show it." I hit this directly — I told a user a document's section numbering had been "removed" based on a markdown fetch, when in fact the headings still had `seq="auto" seq-level="auto"` the whole time; I only discovered the discrepancy by separately fetching the same document with `--doc-format xml`.

## Repro

```bash
# 1. Create a doc with auto-numbered headings via XML (per the resolution on #316, this path works)
lark-cli docs +create --api-version v2 --content - <<XML
<title>Seq export test</title>
<h1 seq="auto" seq-level="auto">Section A</h1>
<h2 seq="auto" seq-level="auto">Section A.1</h2>
<p>body</p>
XML

# 2. Fetch the same doc two ways
lark-cli docs +fetch --api-version v2 --doc <doc_id> --doc-format xml --jq '.data.document.content'
# -> <h1 seq="auto" seq-level="auto">Section A</h1>...  (attribute preserved)

lark-cli docs +fetch --api-version v2 --doc <doc_id> --doc-format markdown --jq '.data.document.content'
# -> # Section A   (no numbering, no seq info, indistinguishable from a heading that was never auto-numbered)
```

Confirmed on lark-cli 1.0.63 and again after `lark-cli update` to 1.0.65 — behavior unchanged.

## Expected behavior

One of:

1. **Preserve the numbering as literal text** in the markdown export (e.g. resolve `seq="auto"` against sibling headings at fetch time and render `# 1. Section A`, `## 1.1 Section A.1`), since Markdown has no attribute syntax to carry `seq`/`seq-level` losslessly.
2. **At minimum, surface a warning** on the fetch response (the CLI already has precedent for this — e.g. `docs +create --doc-format markdown` returns a `missing_document_title` warning when it can't extract a title) so callers know heading numbering info was dropped during markdown conversion and should re-fetch with `--doc-format xml` if they need it.

## Context

Related to (but distinct from) two existing issues:
- #671 (open) — requests a way to *enable* the document-level heading-numbering setting via CLI. That's about the write side of a different setting.
- #316 (closed) — requested `seq`/`seq-level` support in `docs +create`/`+update`; resolved via the v2 XML flow, and a maintainer comment confirmed XML fetch preserves `seq="auto"`/`seq-level="auto"`.

This issue is specifically about **markdown fetch/export losing that already-preserved XML information**, not about writing or enabling numbering.

## Environment

- lark-cli version: 1.0.65 (also reproduced on 1.0.63)
- OS: macOS (arm64)

xu91102:
## Summary
Warn callers when `docs +fetch --doc-format markdown` or `im-markdown` cannot preserve heading `seq` / `seq-level` metadata, so agents know to fetch XML when numbering fidelity matters.

## Changes
- Add fetch response warnings for Markdown formats that drop heading numbering metadata.
- Cover the warning path in docs fetch tests.
- Update the lark-doc fetch reference with XML guidance for heading numbering metadata.

## Test Plan
- [x] `GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go test ./shortcuts/doc -count=1`
- [x] `GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go build -o ./lark-cli .`
- [x] `LARK_CLI_BIN=$PWD/lark-cli GOCACHE=$PWD/.cache/go-build GOPROXY=https://goproxy.cn,direct go test ./tests/cli_e2e/docs -run 'TestDocs_DryRunDefaultsToV2OpenAPI' -count=1`
- [x] `git diff --check`
- [x] conflict marker scan

## Related Issues
- Fixes #1781

## Summary by CodeRabbit

* **New Features**
  * Added support for showing multiple warnings during document fetches.
  * Markdown fetches now warn when heading auto-numbering metadata may be lost, with guidance to use XML when needed.
* **Bug Fixes**
  * Improved warning output so users can see all relevant notices instead of only one.
* **Documentation**
  * Updated fetch guidance to clarify Markdown and im-markdown format limitations.

### DIFF
--- a/shortcuts/doc/docs_fetch_v2.go
+++ b/shortcuts/doc/docs_fetch_v2.go
@@ -74,8 +74,10 @@ func executeFetchV2(_ context.Context, runtime *common.RuntimeContext) error {
 	if err := processHTML5BlockReferenceMapForFetch(runtime, effectiveFetchFormat(runtime), ref.Token, data); err != nil {
 		return err
 	}
-	if warning := addFetchDetailDowngradeWarning(runtime, data); warning != "" && runtime.Format == "pretty" {
-		fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", warning)
+	for _, warning := range addFetchWarnings(runtime, data) {
+		if runtime.Format == "pretty" {
+			fmt.Fprintf(runtime.IO().ErrOut, "warning: %s\n", warning)
+		}
 	}
 	if isIMMarkdownFetch(runtime) {
 		applyFetchIMMarkdown(data, runtime.Str("doc"))
@@ -192,6 +194,17 @@ func effectiveFetchDetail(runtime *common.RuntimeContext) string {
 	return detail
 }
 
+func addFetchWarnings(runtime *common.RuntimeContext, data map[string]interface{}) []string {
+	var warnings []string
+	if warning := addFetchDetailDowngradeWarning(runtime, data); warning != "" {
+		warnings = append(warnings, warning)
+	}
+	if warning := addFetchMarkdownHeadingSeqWarning(runtime, data); warning != "" {
+		warnings = append(warnings, warning)
+	}
+	return warnings
+}
+
 func addFetchDetailDowngradeWarning(runtime *common.RuntimeContext, data map[string]interface{}) string {
 	format := strings.TrimSpace(runtime.Str("doc-format"))
 	detail := strings.TrimSpace(runtime.Str("detail"))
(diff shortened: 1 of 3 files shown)


zhaojiaxing-coding:
Thanks for working on this.

After reviewing the proposal, we decided not to add a warning for this behavior. The conclusion for issue #1781 is that this is an expected limitation of the Markdown format, not a defect to be fixed. Markdown is intentionally content-oriented and does not guarantee lossless preservation of Feishu-specific heading-numbering configuration or other Docx metadata.

Users who need complete document semantics, including heading auto-numbering, should use XML instead. The absence of numbering metadata in Markdown should not be interpreted as the document having no numbering configured.

Since issue #1781 has been closed as not planned, this PR is no longer needed and will not be merged. Closing the PR accordingly.

PR #1827, opened 2026-07-09
docs: clarify Mindnote token handling

fangshuyu-768:
## Summary
- clarify that `mindnote_id` is the Mindnote document token, not a node ID
- document using `drive +inspect` to resolve Mindnote and Wiki URLs before calling `mindnotes`
- update the recommended workflow to confirm a real Mindnote token before node operations

## Tests
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `--mindnote-id` requires the Mindnote document token, not a node ID.
  * Added instructions for retrieving the correct token from Mindnote and Wiki URLs.
  * Updated the recommended workflow to verify the document type before use.
  * Added a reference link for drive inspection.


fangshuyu-768:
Verified with a real Wiki-wrapped Mindnote URL.

- `lark-cli drive +inspect --url <wiki-url> --as user --format json` unwraps the Wiki node to `type: mindnote` and returns the canonical Mindnote token.
- `lark-cli mindnotes nodes list --mindnote-id <mindnote-token> --as user --format json` succeeds and returns nodes.
- Passing the Wiki node token directly to `--mindnote-id` fails with `3410003 resource not found`.

This confirms the doc needs to explicitly route Wiki/title lookup through Drive before calling `mindnotes nodes list`.

SunPeiYang996 (approved):

PR #1833, opened 2026-07-09
Merge lark sheets development branch

zhengzhijiej-tech:
## Summary
- Merge the lark sheets development branch into main.
- Includes the current sheets CLI updates from feat/lark-sheets-develop.

## Testing
- Not run; merge PR only.

## Summary by CodeRabbit

* **New Features**
  * Added Sheets shortcuts: `+revision-get`, `+changeset-get`, `+formula-verify`, and `+history-list/+history-revert/+history-revert-status`.
  * Enhanced `+rows-resize` / `+cols-resize` with `--height/--width`, `--heights/--widths` map form, plus `font-family` support and expanded chart gradients.
* **Bug Fixes**
  * Updated `sheets --help` to hide deprecated aliases from listings while keeping them executable.
  * Strengthened validation and error hints, including batch/resize and fan-out size/range caps; improved image token compatibility and `+csv-put` non-interactive stdin fallback.
* **Documentation**
  * Updated Sheets guides/references for the new shortcuts and revised resize/flag usage.


zhengzhijiej-tech:
Follow-up attribution notes from the review:

- @xiongyuanwen-byted `+batch-update` sub-ops currently bypass the standalone cobra enum normalization/validation for flat flag-defs enum flags. Schema-backed enums are covered, but flat flags such as `+rows-resize` / `+cols-resize --type` can still enter the batch translator through `mapFlagView` without the `chainEnumNormalization` gate. Please add enum validation/normalization to the batch sub-op validation path or otherwise keep batch/standalone parity.
- @xiongyuanwen-byted the `--size` -> `--height` / `--width` surface change in `+rows-resize` / `+cols-resize` is intentional if the new CLI contract is allowed to break compatibility. If old commands must keep working, this needs a compatibility alias at the spec/CLI surface.
- @wenzhuozhen `lark_sheet_history_revert.go` has a stale file header: it says both shortcuts target `--history-version-id`, but `+history-revert-status` now correctly uses `--transaction-id`.
- @xiongyuanwen-byted `validateDropdownRanges` is now also used by `+cells-batch-set-style`; behavior looks fine, but the name is narrower than its current responsibility.

I opened #1839 only for the `+changeset-get` issue introduced by `zhengzhijiej-tech`.

fangshuyu-768:
I found a few issues that look worth fixing before merging:

1. `+batch-update` sub-operations skip enum validation.

   In `translateBatchOp`, the batch path builds a `mapFlagView` and only calls `validateRawTypes`, which checks numeric and boolean fields but intentionally leaves string fields permissive. Standalone shortcuts still get enum validation from Cobra/common runner, so the same bad value is rejected standalone but can be accepted in a batch sub-op.

   For example, a `+cells-clear` sub-op with an invalid `scope` falls through `normalizeClearType` and becomes `contents`, and a `+range-copy` sub-op with an invalid `paste-type` falls through to `all`. That can turn a typo into a different write operation. The batch translator should validate enum-bearing string flags from `flag-defs` before calling the shortcut input builder.

2. `+revision-get` does not resolve wiki URLs during execution.

   `RevisionGet.Execute` calls `resolveSpreadsheetToken`, while the other sheets execute paths call `resolveSpreadsheetTokenExec`. For a `/wiki/<node>` URL, this sends the wiki node token as `excel_id` to `get_workbook_structure` instead of resolving it to the backing spreadsheet token first. This should use the Exec resolver, matching `+workbook-info` and the rest of the sheets commands.

3. `+formula-verify --exit-on-error` exits successfully for `partial`.

   `formulaVerifyExitOnError` treats `status="partial"` the same as `success`, but the new reference docs say `partial` means the scan was truncated and must not be treated as a clean verification. In CI or agent self-check flows, this can report success while part of the workbook was never scanned. With `--exit-on-error`, `partial` should return a non-zero typed error or otherwise force the caller to narrow/split the scan and retry until `status="success"`.

4. Resize map operations are nondeterministic for overlapping ranges with the same start.

   `resizeMapInput` sorts generated operations only by the parsed start index. Since the input is a Go map, overlapping keys with the same start, such as `{"A:C":100,"A:F":200}`, can be emitted in either order. That changes the final width/height for the overlapping cells. Please add a deterministic tiebreaker, or reject overlapping ranges explicitly.

I also ran:

```bash
go test ./shortcuts/sheets/...
go test ./shortcuts/common ./shortcuts
```

zhengzhijiej-tech:
@fangshuyu-768 Follow-up on items 1 and 3:

1. The batch enum-validation gap is real, but it was not introduced by this PR. At the review base (`74d84586`), `translateBatchOp` already called only `validateRawTypes()` before `mapping.translate`, so flat string enums were already bypassing standalone normalization. We still fixed the pre-existing gap in `d8cffa94` by adding `normalizeAndValidateEnums()` to the batch path, with contract tests for invalid values, canonical casing, aliases, and hyphen/underscore keys.

3. The `partial` exit-code finding does not match the declared `--exit-on-error` contract. Both `flag-defs.json` and the command reference define this flag as returning non-zero only for `status=errors_found`, and `TestFormulaVerifyExitOnError_StatusMatrix` explicitly locks `partial` to a zero exit. `partial` is still emitted in the JSON with `has_more=true`, and the workflow docs require the caller to narrow/split the scan until `status=success`; it is not converted to `success`. Making `partial` non-zero would be a separate contract change rather than a correction to the current implementation, so we are keeping the existing behavior.

caojie0621 (dismissed):

liangshuo-1 (approved):

PR #1892, opened 2026-07-14
fix: support safe file input for IM message bodies

SiyiZhu1:
## What changed

- allow `im +messages-send` and `im +messages-reply` body flags (`--text`, `--markdown`, and `--content`) to read from a cwd-relative `@file` or stdin (`-`)
- document the safe input forms for long and generated messages
- add unit coverage for the input contract and a dry-run regression test containing shell metacharacters and Chinese text

## Why

Long or agent-generated message bodies are currently passed only as inline command-line strings. Markdown code fences, `$()` expressions, dollar signs, quotes, and backticks can therefore be interpreted by the caller's shell before lark-cli receives them, producing corrupted or incomplete Feishu messages.

The shortcut framework already has validated `@file` and stdin input resolution. The IM send/reply body flags simply did not opt into it.

## User impact

Agents and scripts can safely send exact long-form content without shell interpolation:

```bash
lark-cli im +messages-send --chat-id oc_xxx --markdown @./message.md
generate-message | lark-cli im +messages-send --chat-id oc_xxx --markdown -
```

## Validation

- `git diff --check`
- added shortcut contract coverage
- added a dry-run E2E regression test that preserves code fences, `$HOME`, `$()`, and Chinese content
- local Go execution was unavailable in the authoring environment; CI is expected to run the repository's Go checks

## Summary by CodeRabbit

* **New Features**
  * Message send and reply commands now accept message content from files or standard input using `--content`, `--text`, and `--markdown`.
  * Added guidance for sending long or generated Markdown messages through files or pipelines.

* **Bug Fixes**
  * Markdown content preserves shell-like syntax and special characters without unintended interpretation.


liangshuo-1:
Hi @SiyiZhu1, thanks for the PR! Please sign the [CLA](https://cla-assistant.io/larksuite/cli?pullRequest=1892) so we can merge this. Note that your commit email ("ZSY") isn't linked to a GitHub account — you'll need to [link it](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user) first for the check to pass.

liuxinyanglxy:
Hi@SiyiZhu1 
Thanks for the PR! Letting IM message bodies accept @file and stdin input is a good direction — it reuses the shortcut framework's already-vetted input mechanism and genuinely helps avoid long or agent-generated content getting mangled by shell interpretation.

That said, it's not mergeable yet. Three things need to be addressed first:

The new dry-run test currently fails, and a one-line path fix won't be enough. TestIMMessagesSendMarkdownFileDryRunPreservesShellSyntax reads data.request.body.content, which doesn't exist in the dry-run output (the correct path is data.api.0.body.content, matching every other dry-run test in the repo); that's also why e2e-dry-run is currently red in CI. We actually ran it both ways: even after switching to the correct path, the test still fails, because that field holds the JSON-serialized message content (& becomes &, newlines become \n), not the raw text. Verifying "shell metacharacters are preserved" requires parsing that field as JSON first and asserting on the decoded text field.
Ordinary messages starting with "@" now fail to send. Something like --markdown "@channel please check this" — very normal chat text — now gets treated as a file path and fails with "cannot read file". There is an existing @@ escape to recover the literal value, but neither updated doc mentions it.
A literal "-" message is now silently swallowed as a stdin marker, with no escape available. A caller who wants to send the literal one-character message "-" now gets an empty value read from stdin instead, followed by an error message that doesn't point at the real cause.

Also, it looks like the CLA is still unsigned — please take a look at liangshuo-1's earlier comment on that.

PR #1899, opened 2026-07-15
refactor: converge success output through a single Emitter that owns the write

sang-neo03:
## Why

The success-output path (content-safety scan → envelope → jq → format render → write) had no single owner: it was assembled at two envelope points (`RuntimeContext.emit`, `WriteSuccessEnvelope`) and re-implemented across several format-routing functions, which had drifted (raw/Meta support, `MarkRaw`, the jq safety-warning). Write failures were silently swallowed on most paths, and `_notice` injection mutated the caller's map.

## What this PR does (now)

Converges all success output onto a single command-scoped `internal/output.Emitter` **and makes that Emitter own the write** — proven byte-for-byte on the success path by a differential golden harness driven from the real legacy entry points.

> **Scope note.** An earlier revision of this description called the PR "deliberately unwired / no production caller migrated / six production files empty diff / skip acceptance." That is no longer accurate — all success-output callers are migrated and the Emitter now owns write-failure semantics. The description below reflects HEAD.

### Migrated onto the Emitter (byte-for-byte)
- `RuntimeContext.Out` / `OutRaw` / `OutFormat` / `OutFormatRaw` / `OutPartialFailure` — signatures unchanged; `emit` / `outFormat` deleted.
- `WriteSuccessEnvelope` — thin delegate to `Emitter.Success`.
- `internal/client.HandleResponse` — the non-JSON structured-response presentation branch.
- api / service pagination — per-page output via `Emitter.StreamPage`.

### Write-ownership contract (this PR)
- Leaf writers gained error-returning `Write*` cores (`WriteJSON`, `WriteNDJSON`, `WriteTable`, `WriteCSV`, `WriteAlertWarning`, `PaginatedFormatter.WritePage`); legacy `Print*` / `FormatValue` wrappers keep their exact behavior for unmigrated callers.
- The Emitter renders to a buffer then `io.Copy`s to stdout, so a marshal/render failure leaves stdout empty and returns a typed internal error (with cause via `errors.Unwrap`/`errors.Is`); a stdout write failure is propagated, not swallowed.
- `handleEmitterError` now captures every error (not only the jq/safety branches).
- `_notice` is injected into a `maps.Clone` of the payload — the caller's map is never mutated and an existing `_notice` is never overwritten.
- jq's own typed error (validation for a bad expression, api for a runtime failure) is preserved; only genuine stdout write failures are wrapped.

## Contract table (per format)

| Format | stdout shape | envelope | jq | content-safety | writer failure |
|---|---|---|---|---|---|
| `""` / `json` | standard `Envelope` | yes | — | alert in envelope; block → typed policy error, no stdout | typed internal error, stdout empty |
| `--jq` | filtered `Envelope` | yes | jq's own typed error preserved | alert also warned to stderr (when `JQSafetyWarning`) | typed internal error |
| `pretty` | naked business data (command renderer) | no | — | alert warned to stderr | typed internal error |
| `table` / `csv` / `ndjson` | naked business data | no | — | alert warned to stderr | typed internal error, stdout empty |

## Behavior changes

- **Intended fixes:** write/marshal failures now propagate as typed errors with cause (was: swallowed / `os.Stderr` only); `_notice` no longer mutates the caller's map; struct/named-map payloads keep `_notice` on the unknown-format → JSON fallback; jq's typed error is preserved through the buffer.
- **Strictly preserved (byte-for-byte):** all success-path stdout/stderr bytes across raw / ok / meta / jq / notice, every format, content-safety fail-open/fail-closed, and pagination — locked by the golden double-diff.
- **Known residuals (deferred, see Follow-ups):** unknown `--format` still falls back to JSON instead of a typed validation error; the pretty renderer does not yet receive the scanned data; the streaming formatter advances column state before confirming the write (write failures are propagated, but the poisoned-state machine is follow-up); notice still flows via the global `PendingNotice` provider and `JQSafetyWarning` is still caller-selected.

## Oracle provenance

- Legacy base: `4a56748bfa941ff0ee0bfec92e65acac427732b0`.
- Golden fixtures are frozen from the legacy entry points; the differential harness runs a **double diff** (Emitter vs golden AND the migrated `RuntimeContext.Out*` vs golden) so it cannot self-reference.
- The golden self-update env vars were removed; regeneration is allowed only from the fixed base, never from the current SUT.

## Test Plan

- [x] `go build ./...` / `go vet ./...` clean; `gofmt` clean
- [x] `GOTOOLCHAIN=go1.25.4 go test -race -count=1 ./internal/output ./shortcuts/common ./cmd/api ./cmd/service`
- [x] golden double-diff byte-for-byte (success path frozen)
- [x] normative contract suite: `err==nil ⇒ fully written`, marshal failure → typed internal error + empty stdout, writer failure preserves `errors.Is` cause, Emitter does not mutate the caller map, existing `_notice` not overwritten, notice read at most once, jq runtime error preserves its typed error
- [x] write-failure matrix: `OutRaw` propagates (flipped from the old ignore test), `runShortcut` non-zero exit on a failing writer, api/service mid-page write failure stops further pages
- [ ] acceptance-reviewer — running (owed now that production callers are migrated)

## Follow-ups (out of scope; separate PRs, one verifiable invariant each)

1. Strict typed `Format`: unknown format → typed `--format` validation error, reject illegal combos (`jq+table`, `raw+csv`). *(user-visible; needs changelog)*
2. `PrettyRenderer(w, data, color)` receives the scanned data.
3. Streaming: commit formatter state only after a successful page write; poison the stream on failure; unify the format-lock.
4. Remove the global `PendingNotice`; notice only in `Envelope.Notice`; drop caller-selected `JQSafetyWarning` (always warn).
5. Consolidate the duplicated api/service pagination; migrate remaining direct-`PrintJson` paths (binary/download metadata, event consume) and the pretty dry-run safety bypass; remove the `WriteSuccessEnvelope` delegate once callers are migrated.

## Reviewer guide

contract/golden (`emitter_contract_test.go`, `emitter_legacy_compat_test.go`) → `Emitter` (`emitter.go`) → `RuntimeContext` adapter + `handleEmitterError` → api/service pagination.

## Related Issues

N/A (follow-ups tracked out-of-band, listed above).


sang-neo03:
Both CodeRabbit findings are addressed in 001e2f80:

- **Critical (raw JSON path in `emitter.go`):** the encoder error is now returned (`return enc.Encode(env)`) instead of discarded, so `Success` no longer reports success on a failed write. Locked by a new `TestEmitterRawJSONPropagatesWriteError`.
- **Minor (`StreamPage` `--jq`):** added `TestEmitterStreamPageRejectsJQ`, which asserts the `*errs.ValidationError` and its `--jq` param via `errors.As`.

The only remaining red check is `e2e-live`, which is failing on upstream `HTTP 429` rate-limiting of the shared test account (unrelated to this change, which adds two unwired files and touches no live API path).

@coderabbitai review

liangshuo-1 (approved):

PR #1911, opened 2026-07-16
feat(sheets): eval round 2 — cut agent error rate and --help lookups

xiongyuanwen-byted:
## Background

Round 2 of eval-driven sheets optimization, rebased onto the latest `feat/lark-sheets-develop` (`8897196d`).

## Changes

- **feat(sheets): cut agent error rate and --help lookups (eval round 2)** — targets the top failure modes from round 2 evals, reducing agent error rate and the number of `--help` lookups.
- **chore(sheets): sync skill docs and flag data from sheet-skill-spec** — syncs skill docs and flag data from sheet-skill-spec.

## Notes

- During rebase, the "import mislabeled .xls workbooks by sniffing content" fix already existed on the target branch (identical patch-id), so it was auto-skipped — no duplicate.
- The target branch was force-rewritten and advanced in the meantime; the two new commits were cleanly replayed onto the new tip via `--onto` with no conflicts. One hunk touching the `--type` description in `lark-sheets-workbook.md` was auto-dropped because upstream already has the same end state — no content lost.

### DIFF
--- a/internal/cmdutil/confirm.go
+++ b/internal/cmdutil/confirm.go
@@ -4,6 +4,10 @@
 package cmdutil
 
 import (
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/larksuite/cli/errs"
 )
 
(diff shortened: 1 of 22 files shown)


internal/cmdutil/confirm.go
```diff
@@ -14,12 +18,58 @@ import (
...
+// (falling back to the plain hint) when any argument reads stdin ("-",
+// whose piped data a bare re-run would not reproduce) or when the rendered
+// command would be unreasonably long to echo back.
 func RequireConfirmation(action string) error {
-	return errs.NewConfirmationRequiredError(errs.RiskHighRiskWrite, action,
-		"%s requires confirmation", action).
-		WithHint("add --yes to confirm")
+	err := errs.NewConfirmationRequiredError(errs.RiskHighRiskWrite, action,
+		"%s requires confirmation", action)
+	if retry := retryCommandWithYes(os.Args); retry != "" {
```
zhengzhijiej-tech, line 32: [P1] Avoid echoing the complete argv in confirmation errors

`RequireConfirmation` embeds the full `os.Args` in a typed error hint that is JSON-encoded to stderr. High-risk commands accept inline sensitive payloads (for example, `apps +db-execute --sql` and the generic service `--data` flag), so any invocation below the 300-byte cutoff can copy SQL, request bodies, or secrets into logs and agent transcripts.

Please keep the fixed `add --yes to confirm` hint and let callers reuse their original argv array with `--yes` appended, instead of echoing the full command.

zhengzhijiej-tech (approved):

PR #1921, opened 2026-07-16
feat(docs): add local authoring and resource workflows

SunPeiYang996:
## Summary

Add an XML-first local authoring workflow for agents: initialize isolated drafts, run tolerant profile, Presentation Decision, and resource checks, then publish through document create/update with safe local and remote resource handling.

## Changes

- Add `docs +script --command init-draft`, which creates a unique `draft_<8-hex>_folder`, saves the Presentation Decision, and returns a new `draft_path` without pre-creating the XML file.
- Add `docs +script --command parse` for local XML drafts or remote documents. Top-level `ok` reports command execution; Profile Check pass/fail is reported by `data.assessment.status`.
- Return actionable `data.diagnostics[]` for failed profile, decision, and resource checks. Remote images with the same failure reason are grouped and identified by `image_indices`.
- Prepare local and remote resources for `docs +create` and `docs +update`, including bounded-concurrency uploads, token binding, verification, image dimensions, and cleanup. Reject disallowed remote image sources before the main Docs write and keep download-time validation for DNS or redirect changes.
- Route failed draft cleanup through the same FileIO backend used for saving, removing only the generated sidecar and empty workspace directory.
- Keep `docs +script parse` XML-only; remove the old word-stat script and superseded `create-temp-xml`, `markdown-to-xml`, and `--file-name` surfaces.
- Refresh the `lark-doc` authoring workflow and genre guidance, and expand unit, dry-run, workflow, concurrency, platform, and compatibility coverage.

For agents: run `parse` against the returned `draft_path`, use `data.assessment.status` as the check result, and fix only the locations identified by `data.diagnostics[]`. Rebuild the full draft only when it is empty, truncated, or structurally invalid. `parse` is a tolerant profile/preflight check, not an XML schema or server-acceptance guarantee.

## Test Plan

- [x] `make unit-test` (race-enabled)
- [x] `go vet ./...`
- [x] Formatting and affected-package `golangci-lint v2.1.6`
- [x] Targeted `docs +script` and local/remote resource unit and E2E coverage
- [x] Real built CLI invocation for the draft workflow

## Related Issues

- None


shortcuts/doc/internal/docxparse/schema.go
```diff
@@ -0,0 +1,268 @@
...
+		if !exists {
+			continue
+		}
+		rule := rules[alias]
+		if rule.transform != nil {
+			var ok bool
+			value, ok = rule.transform(value)
+			if !ok {
+				continue
+			}
```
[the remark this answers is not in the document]
SunPeiYang996: Verified against the current HEAD. The alias-rewrite schema path was removed together with schema.go. Current parsing preserves source input instead of silently normalizing aliases, so this finding targets deleted behavior and no current fix applies.

shortcuts/common/drive_media_upload_test.go
```diff
@@ -7,20 +7,40 @@ import (
...
+			if !errors.Is(err, context.Canceled) {
+				t.Fatalf("error = %v, want context.Canceled cause", err)
+			}
+			problem, ok := errs.ProblemOf(err)
+			if !ok || problem.Category != errs.CategoryNetwork || problem.Subtype != errs.SubtypeNetworkTransport {
+				t.Fatalf("problem = %#v, ok = %v", problem, ok)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
SunPeiYang996: Verified against the current HEAD. waitDriveMediaMultipartRequest and the associated pacing implementation were removed from the PR, so this cancellation test target no longer exists and no current fix applies.

skills/lark-doc/references/lark-doc-create.md
```diff
@@ -56,25 +51,11 @@ lark-cli docs +create --doc-format markdown --title "项目计划" --content $'#
...
-
-- [`lark-doc-create-workflow.md`](style/lark-doc-create-workflow.md) — 从零创作工作流（Code-Act Loop、单 Agent 串行撰写）
-- [`lark-doc-style.md`](style/lark-doc-style.md) — 文档写作原则（默认段落、按体裁、组件克制）
-- [`lark-doc-xml.md`](lark-doc-xml.md) — XML 语法规范
-- [`lark-doc-fetch.md`](lark-doc-fetch.md) — 获取文档
-- [`lark-doc-update.md`](lark-doc-update.md) — 更新文档
-- [`lark-doc-media-insert.md`](lark-doc-media-insert.md) — 插入图片/文件到文档
+|参数|必填|说明|
+|-|-|-|
+|`--title`|否|文档标题，Markdown 导入时使用；XML 创建推荐在 `--content` 开头写 `<title>...</title>`；多个标题仅保留第一个并在 `warnings` / `degrade_details` 提示|
```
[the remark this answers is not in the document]
SunPeiYang996: Verified against the current command paths. The CLI faithfully forwards the full XML and both context flags to docs_ai; it does not discard title or context input. The cited text documents server response semantics for the document title and nested excerpts. Rejecting these valid requests client-side would change API behavior, so no change was made.

shortcuts/doc/docs_script.go
```diff
@@ -0,0 +1,748 @@
...
+	}
+	resolvedPath, err := runtime.ResolveSavePath(path)
+	if err != nil {
+		return errs.NewInternalError(errs.SubtypeFileIO,
+			"resolve reserved draft XML path %s: %s", path, err).
+			WithCause(err)
+	}
+	rawDecision := strings.TrimSpace(runtime.Str("presentation-decision"))
+	decisionPath := filepath.Join(filepath.Dir(path), docsScriptDecisionFile)
+	if _, err := runtime.FileIO().Save(decisionPath, fileio.SaveOptions{
```
fangshuyu-768: [P2] Reserve the draft workspace atomically

`FileIO.Save` uses replacement semantics, so a collision with an existing `draft_<8-hex>_folder` overwrites its Presentation Decision and may return a `draft_path` whose XML already exists. If output subsequently fails, cleanup removes that pre-existing workspace entirely. Create the workspace with an exclusive/no-replace operation, retry on collision, and only remove directories proven to be owned by this invocation.

skills/lark-doc/references/lark-doc-xml.md
```diff
@@ -1,183 +1,54 @@
...
-- 任何包含 `<cite type="user">` 的 XML 在导入、新建或编辑回写时，都必须显式传入 `user-id`；其值为用户的 `open_id`，不得省略。
-- 当从 IM 消息、日历、审批、任务等来源获取到用户的 `open_id` 时，写入文档**必须**使用 `<cite type="user" user-id="open_id">` 标签，而非纯文本名字。这样文档中会渲染为可点击的 @人。
-- 典型场景：IM 消息的 `sender`、`mentions`、reactions 的 `operator`、卡片消息中引用的用户、系统消息中的用户名、合并转发中的用户名。
-- 当只有纯文本名字而没有 `open_id` 时（如系统消息、合并转发内容），先通过 `lark-cli contact +search-user --query "名字" --as user` 反查 `open_id`，再写入 cite 标签。
+- `p, h1-h9, blockquote, hr, img, b, em, u, del, br, span` 语义不变。普通文档建议只使用 `h1-h6`，`h7-h9` 仅在确需更深层级时使用。
+- `<a type="url-preview" href="URL">链接标题</a>`
+- `<latex>E = mc^2</latex>`：适用行内公式，也适用于上标、下标写法。
+- `<ol><li>第一项<ul><li>子项</li></ul></li><li>第二项</li></ol>`：子列表放在 `<li>` 内；新增列表项必须放在 `<ul>` 或 `<ol>` 内。
+- `<pre lang="go" caption="示例"><code>fmt.Println(&quot;hello&quot;)</code></pre>`：代码必须放在 `<code>` 内，禁止直接放在 `<pre>` 下；`caption` 可省略。
+- `<img path="@./photo.png"/>`：上传当前工作目录内的本地图片。也可用 `<img href="URL"/>` 上传公开 HTTP(S) 网络图片，或用 `<img src="token"/>` 复制原始图片；三者任选一个，可选 `width`、`height`、`caption`、`name`。使用 `href` 时，CLI 会先把图片下载到进程内存，全部下载成功后再创建或更新文档并上传图片；响应须为 PNG、JPEG、GIF 或 WebP，单图不超过 20MiB。内部网络图片须先下载到本地再使用 `path`。
```
fangshuyu-768: [P2] Describe the actual post-write remote-image flow

This says every remote image is fully downloaded before the document create/update, but `executeCreateV2` and `executeUpdateV2` perform the Docs API write first and only download images from `finalizeLocalDocResources`. A full-download failure after preflight therefore leaves an already-created or updated document and returns a partial failure. Either pre-download all images before the primary write or document the post-write partial-success behavior and required `local_resource_failures` handling.

SunPeiYang996: Addressed in f9c1b68f. The guidance now states that href-based remote images are converted to local resources and uploaded, and no longer claims that every download completes before the document write.

shortcuts/doc/docs_script.go
SunPeiYang996: Thanks for flagging this. We verified the replacement and cleanup behavior, but given the cryptographically random 8-hex workspace name and the expected low workspace volume, this PR explicitly accepts the residual collision risk in favor of the shorter user-facing path. No code change is planned for this PR.

shortcuts/doc/docs_script.go
```diff
@@ -0,0 +1,776 @@
...
+	docsScriptDraftTip              = "The workspace directory has been created successfully. draft_path points to a new XML file that does not exist yet. Create and write the file directly without reading it first."
+	docsScriptListBlockType         = "list"
+)
+
+var DocsScript = common.Shortcut{
+	Service:     "docs",
+	Command:     "+script",
+	Description: "Initialize a document draft workspace, or parse, preflight, and profile documents",
+	Risk:        "read",
+	AuthTypes:   []string{"user", "bot"},
```
fangshuyu-768, line 56: **[P1] Preserve authentication-free local script execution**

`runShortcut` resolves identity and configuration before this shortcut reaches validation or execution. Since `DocsScript` has no local execution branch, both `init-draft` and `parse --content` now fail with `config/not_configured` on a clean machine even though they make no Lark API call. This also contradicts `TestDocsScriptLocalParseWithoutLarkConfiguration` and the local-first workflow. Please restore a command-aware local bypass for `init-draft` and `parse` with `--content`, while retaining normal authentication and scope checks for `--doc`.

shortcuts/doc/local_doc_resources.go
```diff
@@ -0,0 +1,2829 @@
...
+	}
+
+	for i, resource := range resources {
+		outcome := &localDocResourceOutcome{
+			Resource:      resource,
+			Status:        "pending",
+			CleanupStatus: "not_needed",
+		}
+		matches := byMarker[resource.Marker]
+		if len(matches) != 1 {
```
fangshuyu-768, line 1128: **[P2] Do not report a missing placeholder correlation as `not_needed`**

When `len(matches) == 0`, the outcome keeps its initial `cleanup_status` of `not_needed` while `SafeToCleanup` remains false. The cleanup phase then skips this outcome, so `local_resource_failures` says cleanup was unnecessary even though the response did not identify the placeholder and an unbound block may remain in the document. Please mark this case as `skipped_ambiguous` (or another explicit missing-correlation state) and add a test for the emitted failure payload.

shortcuts/doc/docs_script.go
```diff
@@ -0,0 +1,776 @@
...
+	Document *struct {
+		Content *string `json:"content"`
+	} `json:"document"`
+}
+
+func installDocsScriptHelp(cmd *cobra.Command) {
+	installDocsContentPathCapture(cmd)
+	cmd.Example = `  lark-cli docs +script --command init-draft --presentation-decision '<JSON>'
+  lark-cli docs +script --command parse --content "@./draft.xml"
+  lark-cli docs +script --command parse --content "@./draft.md"
```
fangshuyu-768: **[P2] Remove the stale Markdown example from `+script` help**

`parse` now unconditionally calls `ParseCompatibleXML`, and `TestDocsScriptParseRejectsMarkdownFromFile` explicitly pins Markdown rejection, but this example still tells users to pass `@./draft.md`. The embedded `lark-doc` Skill and coverage documentation make the same claim, so agents following the shipped guidance will deterministically fail. Either restore Markdown parsing or update all user-facing guidance and coverage text to describe this command as XML-only.

shortcuts/doc/local_doc_resources.go
SunPeiYang996: Thanks, this is valid. When the API response contains no block for an expected local-resource marker (`len(matches) == 0`), the old default `cleanup_status` remained `not_needed` while `SafeToCleanup` was false. That made `local_resource_failures` claim cleanup was unnecessary even though the placeholder correlation was missing and an unbound block could remain. Commit `2828eb6e` now reports `skipped_ambiguous` for that case and adds a regression assertion for the emitted failure payload.

shortcuts/doc/docs_script.go
SunPeiYang996: Fixed in `2828eb6e`. `docs +script --command parse` now documents XML-only input: removed the `@./draft.md` help example, updated the lark-doc Skill and E2E coverage text, and updated the help test. Markdown create/update workflows are unchanged.

shortcuts/doc/docs_script.go
SunPeiYang996: Thanks for flagging this. This shortcut is intentionally only reachable through the normal authenticated CLI pipeline; callers entering this interface have already resolved identity and configuration. We do not support invoking `docs +script` as a standalone unauthenticated local utility, including `init-draft` and `parse --content`. The local-first behavior means the authenticated shortcut can perform local parsing without an additional Lark API call; it does not bypass CLI authentication or configuration. No code change is planned for this thread.

fangshuyu-768 (dismissed):

liangshuo-1 (approved):

PR #1928, opened 2026-07-17
feat(slides): report resolved table size mismatches

ethan-zhx:
## Summary

Improve the Slides XML lint for table geometry. It detects table bounds outside the canvas and reports informational size mismatches when the renderer's resolved row or column totals differ from the declared table dimensions.

## Changes

- Add table canvas-boundary errors for tables that overflow the slide.
- Add `info` output for weighted row and column layout resolution mismatches, including resolved dimensions and correction guidance.
- Add direct and command-line black-box coverage for table layout cases; remove obsolete table-size guidance.

## Test Plan

- [x] `python3 -m unittest xml_text_overlap_lint_test` (42 tests)
- [x] Command-line black-box coverage verifies exact sizing, declared-size overflow, insufficient remaining space, and missing table target dimensions.

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Improved table linting with new validations for tables that extend beyond slide canvas boundaries.
  * Added informational reporting for when table layout sizing resolves differently from declared dimensions, including resolved size details.
* **Bug Fixes**
  * Canvas boundary checks now use the resolved table geometry rather than only declared bounds.
* **Documentation**
  * Updated the XML format guide by removing outdated table sizing guidance.
* **Tests**
  * Added extensive coverage for overflow detection, resolved-size behavior, and end-to-end CLI lint output.


BD-ZERO (approved):

R0bynZhu (approved):

fangshuyu-768:
Thanks for the table-geometry lint improvements. I verified the current PR head (`ab27e2b`) locally: the declared test command passes (`python3 -m unittest xml_text_overlap_lint_test`, 47 tests), and CI is green. I still see a few issues worth fixing before merge:

1. `extract_table_column_sizes` expands `<col span>` directly into a Python list with no upper bound. A very large but well-formed `span`, e.g. `span="10000000"`, can make the lint process hang or run out of memory before it has a chance to report anything. Please validate/bound `span_count` before `sizes.extend(...)`, ideally using a table column limit and emitting a lint error instead of allocating unbounded entries.

2. `extract_elements` only keeps a table when all of `x`, `y`, resolved `width`, and resolved `height` are present. Because table `width`, `height`, and `colgroup` are optional, this can skip real overflow cases. For example, a table with rows but no declared width/colgroup can still resolve height and overflow the bottom edge, but it is dropped because width is `None`; similarly, a table with columns but no rows/height can overflow horizontally and be dropped. The canvas check should either resolve one effective bbox with sensible defaults, or validate each axis independently when only one dimension is known.

3. The table content extraction uses `slide_xml.find("</table>", match.end())` even for self-closing `<table .../>` tags. If a self-closing table is followed by another table, the first table can accidentally consume the second table’s content and report false `table_resolved_size_mismatch` diagnostics. Please detect self-closing opening tags and treat their content as empty, or switch this portion to parsed XML traversal.

4. Minor test maintainability: in the CLI black-box test, assert `completed.returncode` before `json.loads(completed.stdout)`. If the subprocess crashes, this preserves the useful stderr instead of masking it with a JSON decode error.

Separately, my read of the new sizing behavior is: table width/height are treated as target minimum dimensions; filled column/row sizes are preserved; empty columns/rows share remaining space; when all child sizes are filled and the target is larger, sizes are scaled by their existing weights; when the target is too small, filled sizes/defaults win and the resolved total can exceed the declared table size. The resolved bbox is then used for `table_out_of_canvas`, while declared-vs-resolved differences are reported as `info`.

PR #1987, opened 2026-07-21
fix(slides): reindent xml-get output for readability

tianyouskrrr:
## Summary
- `slides +xml-get` now reindents presentation/slide XML (via `etree`) before writing it to `--output` files and `--raw` stdout — the text surfaces a human or line tool reads directly
- The default JSON envelope carries the server's original XML **verbatim**: inside a JSON string every newline is escaped to `\n`, so formatting there adds no readability while inflating the payload; passthrough also keeps that read path byte-exact without even parsing the content (no `pretty_printed` field, no fallback needed on this path)
- Schema-documented whitespace character references such as `&#32;`/`&#9;` (and CR/LF forms `&#13;`/`&#10;`, incl. hex/zero-padded spellings) are preserved lexically across the parse/write pass on the formatted surfaces
- `--raw` means formatted XML stdout without the JSON envelope; for the unmodified server payload, use the default envelope
- If formatting fails on `--raw`/`--output`, the command returns the original server content, writes an explicit warning to stderr, and (for `--output`) reports `pretty_printed: false` in file metadata
- Adds `github.com/beevik/etree` as a direct dependency

## Why
The API returns XML as a single unindented line, which is unreadable for decks with many shapes (for example, PPTX-imported presentations). Reformatting client-side in this read-only command keeps the change contained while preserving whitespace references that can be significant in later SML write operations.

Formatting is deliberately scoped to the file/stdout text surfaces: agents consuming the JSON envelope pay per token and gain nothing from escaped newlines, and the untouched envelope doubles as a byte-exact escape hatch that needs no fidelity machinery at all.

The fallback on the formatted surfaces remains best-effort by design: unusual or non-strict XML from the service is still retrievable, but callers are no longer left unaware that formatting was skipped.

## Test plan
- [x] `make unit-test`
- [x] `go vet ./...`
- [x] `gofmt -l .`
- [x] `go mod tidy` (no additional changes)
- [x] `make build`
- [x] Regression coverage for decimal/hex/zero-padded whitespace character references, mixed-content sibling boundaries, non-mixed text leaves, idempotency, `--raw` help semantics, explicit fallback signaling for raw/file output, and verbatim envelope passthrough (well-formed and malformed content, both scopes)
- [x] Manual PPTX import/read/write validation with repeated spaces, a tab, and a whitespace-only rich-text run
- [x] Mutation-tested the output wiring: re-adding envelope formatting, or dropping formatting from `--raw`/`--output`, each breaks pinned tests


fangshuyu-768:
One behavior risk worth tightening: `prettyPrintXML` currently parses entity references and serializes them back as literal characters. In particular, SML documents whitespace-sensitive text by using `&#32;` / `&#9;` inside `<p>` and inline text-bearing elements, but after this pass those become literal spaces/tabs in the `slides +xml-get` output.

That is usually equivalent at the XML layer, but `+xml-get` is also used as the read step in read-modify-write slide workflows. If a user saves the pretty-printed XML and feeds it back into replace/create, the server may treat those literal spaces/tabs according to the normal SML text whitespace rules, potentially collapsing or dropping whitespace that was originally preserved by numeric character references.

Could we either preserve/re-escape sensitive whitespace inside text-bearing subtrees, or add a regression test proving that a readback containing `&#32;` / `&#9;` remains safe to round-trip through the write APIs?

shortcuts/slides/slides_xml_get.go
```diff
@@ -289,3 +293,133 @@ func outputSlidesXMLGetContent(runtime *common.RuntimeContext, content string, o
...
+// across stdout, --output files, and the JSON envelope alike. There is no
+// flag to disable this: the server returns XML as a single unbroken line,
+// and this command has no other reason to exist than to hand a human
+// something they can read.
+//
+// Reindentation never enters a textBearingTags element: those subtrees are
+// serialized exactly as parsed, so any deliberately preserved whitespace
+// survives untouched. CDATA sections are also preserved rather than
+// collapsed into escaped text.
+func prettyPrintXML(xmlContent string) (string, error) {
```
fangshuyu-768, line 342: This parse/write pass does more than indentation: etree resolves character references and serializes them back as literal text. In SML, whitespace-sensitive text is represented with `&#32;` / `&#9;` inside `<p>` and inline text-bearing elements; after this pass those become literal spaces/tabs in `slides +xml-get` output.

That may break read-modify-write workflows where users save the readback XML and feed it back into replace/create, because the server can then apply normal SML whitespace collapsing/ignoring rules to whitespace that was originally intentionally preserved. Could we either preserve/re-escape sensitive whitespace inside text-bearing subtrees, or add a regression test proving that XML containing `&#32;` / `&#9;` remains safe to round-trip through the write APIs?

shortcuts/slides/slides_xml_get_test.go
```diff
@@ -515,3 +535,236 @@ func TestSlidesXMLGetRejectsRemoveAttrIDForSingleSlide(t *testing.T) {
...
+// incidental formatting whitespace, so a naive reindent (etree's bare
+// Indent()) silently deletes them. See the mixed="true" content types
+// (p/strong/em/u/span/del/a/shadow/outline/chartTitle/chartSubTitle).
+func TestPrettyPrintXMLPreservesEscapedWhitespaceAsSoleLeafContent(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{"space in p", `<content><p>&#32;</p></content>`, "<content>\n  <p> </p>\n</content>\n"},
```
fangshuyu-768: This test name/comment says escaped whitespace is preserved, but the expected output actually turns `&#32;` / `&#9;` into literal space/tab characters. That only proves the whitespace was not deleted; it does not prove the SML-preserving escaped representation survives or is safe to feed back into write APIs.

Could we either rename this test to reflect the weaker guarantee, or add a separate regression test for preserving/re-escaping numeric character references in text-bearing content?

shortcuts/slides/slides_xml_get.go
```diff
@@ -183,6 +185,7 @@ func fetchSlidesXMLGetContent(runtime *common.RuntimeContext, presentationID str
 		if content == "" {
 			return "", nil, errs.NewInternalError(errs.SubtypeInvalidResponse, "slides xml get returned empty slide.content")
 		}
+		content = prettyPrintXMLOrOriginal(content)
```
fangshuyu-768: `--raw` still advertises "raw XML", but the content has already been passed through `prettyPrintXMLOrOriginal` here before `outputSlidesXMLGetContent` decides whether to print raw stdout. That changes the contract for callers using `--raw` for exact server readback, diffs, caches, or round-trip debugging.

Could `--raw` bypass formatting, or should the flag/help text be updated to make clear that it now prints formatted XML rather than the raw API payload?

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 require (
 	github.com/atotto/clipboard v0.1.4 // indirect
 	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
+	github.com/beevik/etree v1.5.1
```
fangshuyu-768: This package is imported directly by `shortcuts/slides/slides_xml_get.go`, but it is listed in the second require block, which otherwise appears to hold indirect dependencies. Could we move `github.com/beevik/etree` into the direct dependency block so dependency ownership stays clear?

shortcuts/slides/slides_xml_get.go
```diff
@@ -289,3 +293,133 @@ func outputSlidesXMLGetContent(runtime *common.RuntimeContext, content string, o
...
+	return out, nil
+}
+
+// prettyPrintXMLOrOriginal keeps xml-get best-effort: if the server returns
+// content that is not strictly valid XML, callers still receive the original
+// content instead of losing the previously available read path.
+func prettyPrintXMLOrOriginal(xmlContent string) string {
+	out, err := prettyPrintXML(xmlContent)
+	if err != nil {
+		return xmlContent
```
fangshuyu-768: This silently falls back to the unformatted original when parsing/formatting fails, which conflicts with the PR description that says pretty-printing is always on and malformed input is rejected with a clear internal error.

If best-effort output is the intended behavior, could we document that contract and expose some indication that formatting was skipped? Otherwise this should return a typed error so users do not unknowingly receive a mix of formatted and unformatted XML.

shortcuts/slides/slides_xml_get.go
```diff
@@ -232,13 +239,15 @@ func fetchSlidesXMLGetContent(runtime *common.RuntimeContext, presentationID str
...
 	}
+	content, prettyPrinted := prettyPrintXMLOrOriginal(runtime, content)
 	presentationOut := map[string]interface{}{
 		"content": content,
 	}
 	out := map[string]interface{}{
 		"xml_presentation_id": presentationID,
 		"scope":               "presentation",
 		"xml_presentation":    presentationOut,
+		"pretty_printed":      prettyPrinted,
```
ethan-zhx: 这会导致模型拿到两份数据，占用更多上下文？

tianyouskrrr: 不会拿到两份数据，但是占用的上下文会增加一些

fangshuyu-768 (dismissed):

fangshuyu-768 (approved):

PR #1994, opened 2026-07-21
feat(drive): add +member-remove shortcut

zhaojiaxing-coding:
Add a high-risk Drive shortcut for removing one collaborator permission through the permission member delete API. The command validates resource and member combinations, previews the exact DELETE request, and returns an auditable success receipt.

Key features:

- Reuse existing Drive member target, member type, and wiki permission helpers without changing +member-add

- Require explicit confirmation and reject unsupported bot department removal before the API call

- Cover request shaping, typed errors, dry-run behavior, and the live add/remove workflow

## Test Plan

- Unit (`shortcuts/drive/drive_member_remove_test.go`): request construction, query/body shaping, conditional output fields, typed validation errors, and API-error passthrough.
- Dry-run E2E (`tests/cli_e2e/drive/drive_member_remove_dryrun_test.go`): exact DELETE method/URL/params/body per resource type, plus regression cases rejecting ambiguous multi-segment URLs, %2F-encoded separators, `--type` conflicts, and slash/comma-containing member IDs.
- Live workflow E2E (`tests/cli_e2e/drive/drive_member_remove_workflow_test.go`): docx add→remove round trip, and a fixture-gated apps (Miaoda) add→remove round trip (`LARK_CLI_E2E_DRIVE_APPS_TOKEN`, skipped when unset) that only mutates a transient bot collaborator and reports cleanup failures via `clie2e.ReportCleanupFailure`.
- Targeted package check: `go test ./shortcuts/drive/ -run MemberRemove` (pass), `go vet ./shortcuts/drive/`, `gofmt -l` (clean).

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Added `drive +member-remove` to remove exactly one collaborator’s Drive permission, with wiki-specific handling and structured progress/output.
  * Enforces strict validation and requires confirmation for real deletions.
* **Documentation**
  * Added command reference and skill guidance for safe usage (`--dry-run`, parameter rules, and returned field meanings).
* **Tests**
  * Added unit tests for request construction, conditional output fields, validation, and typed permission error passthrough.
  * Added CLI end-to-end coverage for dry-run and a successful user workflow.

### DIFF
--- a/shortcuts/drive/drive_member_remove.go
+++ b/shortcuts/drive/drive_member_remove.go
@@ -1,0 +1,393 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package drive
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/larksuite/cli/errs"
+	"github.com/larksuite/cli/internal/validate"
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+var driveMemberRemoveIDTypes = []string{
+	"email", "openid", "openchat", "opendepartmentid",
+	"userid", "unionid", "groupid", "wikispaceid",
+}
+
+// DriveMemberRemove removes one collaborator/member permission from a Drive resource.
+var DriveMemberRemove = common.Shortcut{
+	Service:     "drive",
+	Command:     "+member-remove",
+	Description: "Remove one collaborator/member permission from a Drive resource",
+	Risk:        "high-risk-write",
+	Scopes:      []string{"docs:permission.member:delete"},
+	AuthTypes:   []string{"user", "bot"},
+	Flags: []common.Flag{
+		{Name: "token", Desc: "target token or document URL; type is auto-inferred from URL path when omitted", Required: true},
+		{Name: "type", Desc: "target resource type; required when --token is a bare token"},
+		{Name: "member-id", Desc: "single collaborator ID to remove; comma-separated values are rejected", Required: true},
+		{Name: "member-type", Desc: "ID type for --member-id; supported: email|openid|openchat|opendepartmentid|userid|unionid|groupid|wikispaceid", Required: true},
+		{Name: "member-kind", Desc: "request body type when --member-type=wikispaceid; one of wiki_space_member|wiki_space_viewer|wiki_space_editor"},
+		{Name: "perm-type", Desc: "wiki permission scope; defaults to container; rejected for non-wiki types and wiki-space members"},
+	},
+	Tips: []string{
+		"This command removes exactly one collaborator; run separate commands for multiple members.",
+		"Resource type is auto-inferred from URL paths; pass --type when --token is a bare token.",
+		"When --member-type=wikispaceid, pass --member-kind wiki_space_member, wiki_space_viewer, or wiki_space_editor.",
+		"For ordinary wiki collaborators, --perm-type defaults to container; use single_page to remove only the current-page permission.",
+		"Department collaborator removal (--member-type=opendepartmentid) requires --as user; bot identity is not supported.",
+		"A successful response confirms that the removal request completed; it does not prove the permission previously existed.",
+	},
+	Validate: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		_, err := readDriveMemberRemoveSpec(runtime)
+		return err
+	},
+	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
+		spec, err := readDriveMemberRemoveSpec(runtime)
+		if err != nil {
+			return common.NewDryRunAPI().Set("error", err.Error())
+		}
+		return buildDriveMemberRemoveDryRun(spec)
+	},
+	Execute: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		spec, err := readDriveMemberRemoveSpec(runtime)
+		if err != nil {
+			return err
+		}
+		return executeDriveMemberRemove(runtime, spec)
+	},
+}
+
+type driveMemberRemoveSpec struct {
+	Token        string
+	ResourceType string
+	MemberID     string
+	MemberType   string
+	MemberKind   string
+	PermType     string
+}
+
+// APIQueryParams returns the query parameters sent with the member-remove DELETE request.
+func (spec driveMemberRemoveSpec) APIQueryParams() map[string]interface{} {
+	return map[string]interface{}{
+		"type":        spec.ResourceType,
+		"member_type": spec.MemberType,
+	}
+}
+
+// Body builds the request body, carrying the wiki member kind and permission scope only when set.
+func (spec driveMemberRemoveSpec) Body() map[string]interface{} {
+	body := make(map[string]interface{}, 2)
+	if memberKind := driveMemberAddBodyType(spec.MemberType, spec.MemberKind); memberKind != "" {
+		body["type"] = memberKind
+	}
+	if spec.PermType != "" {
+		body["perm_type"] = spec.PermType
+	}
+	return body
+}
+
+// readDriveMemberRemoveSpec reads and validates the command flags into a driveMemberRemoveSpec,
+// enforcing single-member, member-type, wiki, and identity constraints before any request is issued.
+// rejectDriveMemberRemoveDotSegment rejects a value that is an RFC 3986 dot
+// segment ("." or "..") in any "/"-split part. validate.ResourceName only stops
+// "..", but a lone "." (or an encoded "%2e" that PathUnescape decodes to ".")
+// survives EncodePathSegment unchanged and can be normalized away by an HTTP
+// stack or proxy, redirecting this high-risk DELETE to a different route. This
+// is a local, remove-only guard so the shared validate.ResourceName behavior is
+// left untouched for its many other callers.
+func rejectDriveMemberRemoveDotSegment(value, flagName string) error {
+	for _, seg := range strings.Split(value, "/") {
+		if seg == "." || seg == ".." {
+			return errs.NewValidationError(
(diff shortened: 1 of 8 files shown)


tests/cli_e2e/drive/drive_member_remove_workflow_test.go
```diff
@@ -0,0 +1,108 @@
...
+	suffix := clie2e.GenerateSuffix()
+	folderToken := CreateDriveFolder(t, t, ctx, "lark-cli-e2e-member-remove-"+suffix, "user", "")
+	docToken := createMemberRemoveWorkflowDoc(t, ctx, folderToken, suffix)
+	botOpenID := getMemberRemoveWorkflowBotOpenID(t, ctx)
+
+	addResult, err := clie2e.RunCmd(ctx, clie2e.Request{
+		Args: []string{
+			"drive", "+member-add",
+			"--token", docToken,
+			"--type", "docx",
```
fangshuyu-768, line 39: [P2] Cover the advertised apps path with a live round trip

This workflow hard-codes `docx`, so the live gate never exercises the new apps/Miaoda branch. The apps cases in `drive_member_remove_dryrun_test.go` only prove client-side URL inference and request shaping; if the server rejects `DELETE ...?type=apps`, every current test still passes while the help and skill docs claim support. Please add an add -> remove -> member-list live round trip for an apps resource, or avoid advertising apps support until that path can be verified.

zhaojiaxing-coding: Good catch — the workflow only exercised docx, so nothing proved the server accepts type=apps. I verified the apps path end-to-end against a real Miaoda app on production: both POST and DELETE with ?type=apps
succeed (removed=true, resource_type=apps), and the DELETE genuinely mutates state (removing a visible collaborator dropped the list 7→6, adding it back restored 7) — so the "server rejects ?type=apps while tests
stay green" case is disproven. I also added a fixture-gated live test (TestDrive_MemberRemoveAppsWorkflowAsUser) that runs add(view) → remove against an apps resource and asserts removed=true + resource_type=apps.
It's fixture-gated rather than self-provisioning because Miaoda apps have no create/delete endpoint (self-provisioning would leak tenant state); the app is injected via LARK_CLI_E2E_DRIVE_APPS_TOKEN (never a real
token in source, skips when unset) and only the transient bot collaborator it adds is mutated and cleaned up — matching how the apps role/file live workflows are gated. I dropped the member-list-after-remove
check you suggested because apps +member-list doesn't surface bot-type members; the test asserts on the remove response envelope instead, which is what actually proves the round trip.

shortcuts/drive/drive_member_remove.go
```diff
@@ -0,0 +1,266 @@
...
+		body["type"] = memberKind
+	}
+	if spec.PermType != "" {
+		body["perm_type"] = spec.PermType
+	}
+	return body
+}
+
+func readDriveMemberRemoveSpec(runtime *common.RuntimeContext) (driveMemberRemoveSpec, error) {
+	token, resourceType, err := resolveDriveMemberAddTarget(runtime.Str("token"), runtime.Str("type"))
```
fangshuyu-768: [P2] Reject ambiguous URL targets before deleting

`resolveDriveMemberAddTarget` parses the decoded URL path and truncates at the first slash, so both `/docx/doxReal/other` and `/docx/doxReal%2Fother` resolve to `doxReal`. With `--yes`, this command can therefore delete a collaborator from `doxReal` even though the supplied URL does not identify that resource. The same resolver also lets a conflicting explicit `--type` override the URL type. Please use strict exact-segment parsing (for example `EscapedPath`, as in `resolveSecureLabelTarget`) and reject extra or encoded separators and type conflicts, with dry-run regression cases.

tests/cli_e2e/drive/drive_member_remove_workflow_test.go
```diff
@@ -0,0 +1,203 @@
...
+	botOpenID := getMemberRemoveWorkflowBotOpenID(t, ctx)
+
+	// Best-effort cleanup: remove the transient collaborator even if a later
+	// assertion fails before the explicit remove step runs. Removing an absent
+	// member is idempotent, so this is safe when the test path already removed it.
+	t.Cleanup(func() {
+		cleanupCtx, cleanupCancel := clie2e.CleanupContext()
+		defer cleanupCancel()
+
+		_, _ = clie2e.RunCmd(cleanupCtx, clie2e.Request{
```
fangshuyu-768: [P2] Preserve the shared fixture pre-existing membership

This cleanup always removes `botOpenID` and discards the cleanup result. Because `+member-add` is idempotent, if this bot already has access to the shared app, the add can succeed and the test then permanently deletes a pre-existing collaborator; if the add commits but cleanup fails, the leaked permission is silent. Please establish or encode a disposable-member baseline (or restore the prior state), arm cleanup only for membership created by this run, and report cleanup failures via `clie2e.ReportCleanupFailure`, as the existing apps role fixture workflows do.

shortcuts/drive/drive_member_remove.go
```diff
@@ -0,0 +1,349 @@
...
+		DELETE(driveMemberRemovePath(spec)).
+		Params(spec.APIQueryParams()).
+		Body(spec.Body())
+}
+
+// driveMemberRemovePath builds the permissions member DELETE path with each segment escaped.
+func driveMemberRemovePath(spec driveMemberRemoveSpec) string {
+	return fmt.Sprintf(
+		"/open-apis/drive/v1/permissions/%s/members/%s",
+		validate.EncodePathSegment(spec.Token),
```
fangshuyu-768, line 267: [P2] Reject dot segments before constructing the DELETE path

The strict resolver still accepts `https://.../docx/%2e%2e` (as well as a bare `--token ".."` or `--member-id ".."`). After `PathUnescape`, `EncodePathSegment` leaves `..` unchanged, so dry-run emits `/open-apis/drive/v1/permissions/../members/...` and the live request can be normalized to a different route by the HTTP stack or proxy. Please validate both decoded or bare tokens and member IDs with `validate.ResourceName` in addition to the single-segment check, and add dry-run regressions for dot segments.

tests/cli_e2e/drive/drive_member_remove_workflow_test.go
```diff
@@ -0,0 +1,320 @@
...
+
+	// Fail closed on the baseline guard's assumption. The baseline check above
+	// treated "bot absent from member-list" as "safe to run", but that is only
+	// sound if apps +member-list actually surfaces this bot when it IS a member.
+	// Assert exactly that here: right after adding, the bot must appear. If it
+	// does not, member-list cannot observe apps bot membership, the baseline
+	// guard would silently fail open, and continuing could delete a pre-existing
+	// collaborator — so skip instead of proceeding to the destructive remove.
+	if !driveMemberListShows(ctx, appToken, "apps", botOpenID) {
+		t.Skipf("FIXTURE: apps +member-list does not surface bot %s after add; baseline guard cannot be trusted, refusing to run the destructive remove", botOpenID)
```
fangshuyu-768: [P2] Avoid cleanup when membership ownership is ambiguous

When apps `+member-list` does not surface this bot, the baseline can miss a membership that already existed. `+member-add` is idempotent, so it can still return success without creating anything; this `t.Skipf` then runs `t.Cleanup` with `needsCleanup=true`, and cleanup removes the pre-existing collaborator. Please use a disposable member or fixture with an authoritative absent baseline, or otherwise preserve and restore the prior state; skipping before the explicit remove is not safe while cleanup still deletes.

shortcuts/drive/drive_member_remove.go
```diff
@@ -0,0 +1,367 @@
...
+// enforcing single-member, member-type, wiki, and identity constraints before any request is issued.
+func readDriveMemberRemoveSpec(runtime *common.RuntimeContext) (driveMemberRemoveSpec, error) {
+	token, resourceType, err := resolveDriveMemberRemoveTarget(runtime.Str("token"), runtime.Str("type"))
+	if err != nil {
+		return driveMemberRemoveSpec{}, err
+	}
+	// The resolver already rejects separators, but not "..". Validate the resolved
+	// (decoded) token so a bare ".." or an encoded "%2e%2e" in a URL path cannot
+	// reach the DELETE path as a traversal segment.
+	if err := validate.ResourceName(token, "--token"); err != nil {
```
fangshuyu-768, line 125: [P2] Reject the single-dot path segment too

`validate.ResourceName` rejects `..` but still accepts `.`, so the strict resolver continues to accept `https://.../docx/%2e` as well as a bare token or member ID of `.`. Dry-run then emits `/permissions/./members/...` or `/members/.`, which HTTP intermediaries can normalize to a different route. Please reject both RFC dot segments (`.` and `..`) for decoded/bare tokens and member IDs, and add matching dry-run regressions.

tests/cli_e2e/drive/drive_member_remove_workflow_test.go
```diff
@@ -0,0 +1,297 @@
...
+			"--member-id", botOpenID,
+			"--member-type", "openid",
+			"--yes",
+		},
+		DefaultAs: "user",
+	})
+	require.NoError(t, err)
+	removeResult.AssertExitCode(t, 0)
+	removeResult.AssertStdoutStatus(t, true)
+	require.True(t, gjson.Get(removeResult.Stdout, "data.removed").Bool(), "stdout:\n%s", removeResult.Stdout)
```
fangshuyu-768, line 70: [P2] Verify the docx removal through observable state

The asserted add fields can be backfilled from the request spec, while the remove receipt is synthesized with `removed=true`; the delete endpoint can also succeed when the member is already absent. Removing the member-list checks therefore leaves this live workflow able to pass even if the bot was never granted access or the removal is a no-op. Please use an observable user fixture as the apps workflow now does, or verify bot access before and after deletion with a bot-identity read request.

SunPeiYang996 (approved):

PR #2022, opened 2026-07-23
  feat(slides): add layout density lint for sparse/empty containers        

BD-ZERO:
Summary                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  - Extended xml_text_overlap_lint.py (the existing XML layout static-check script) with three new detection rules:                                                                                                                
    - blank_slide: the slide canvas has no visible content at all                                                                                                                                                                  
    - sparse_container_content: a large container (card) has disproportionately low visible-content coverage                                                                                                                       
    - sparse_slide_content: the whole slide has low effective-content coverage                                                                                                                                                     
  - Expanded out-of-canvas detection (*_out_of_canvas) from covering only table/chart/text elements to all element kinds, and added geometry extraction for icon and line elements                                                 
  - Standardized the issue output format (schema_version 2.0): every issue now includes rule (rule name/threshold/comparison), measurement (actual measured values), related_objects, element_ids, message/hint; the top-level     
  summary gained status, release_ready, and screenshot_review_required fields                                                                                                                                                      
  - Adjusted CLI exit-code semantics: only error blocks release (exit 1); when only warnings are present, the full JSON is still emitted with exit 0, so the downstream screenshot-review step can proceed                         
  - Updated validation-checklist.md / SKILL.md / troubleshooting.md accordingly: script path is now resolved relative to the parent directory of the loaded SKILL.md (no guessing a global install path), plus a handling-guidance 
  table for the newly added issue codes                                                                                                                                                                                            
                                                                                                                                                                                                                                   
  Test plan                                                                                                                                                                                                                        
                                                                                                                                                                                                                                   
  - [x] python3 skills/lark-slides/scripts/xml_text_overlap_lint_test.py passes                                                                                                                                                    
  - [x] New/updated test cases cover: blank_slide blocking, sparse-content warnings for both large containers and short cards, whole-slide sparse-content warnings, icons/own <content> counted toward visible coverage, exemption 
  paths for image-overlay cards and edge-spanning layout panels, out-of-canvas detection across all element kinds                                                                                                                  
  - [x] Verified sparse_container_content fires on a known sparse-container sample XML, and produces no false positives on a normal-density page        

## Summary by CodeRabbit

* **New Features**
  * Expanded automated slide validation to flag blank pages, sparse/low-coverage layouts, text/geometry overlap risks, and out-of-canvas elements.
  * Validation now uses rule-based, structured results with clearer gating: errors block delivery/replacement; warnings can require screenshot review.
* **Bug Fixes**
  * Improved handling for rotated/geometry-heavy elements and reduced false positives for minor rounding noise.
* **Documentation**
  * Updated quick reference, validation checklist, and troubleshooting to follow the new lint-and-review workflow and repair order (fix errors, then review warnings).
* **Tests**
  * Updated expectations for structured error/warning output and added coverage for density/blank-slide and richer geometry diagnostics.


skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1194,12 +1195,7 @@ def detect_elements_out_of_canvas(
     elements: list[dict[str, Any]], slide_width: int | float, slide_height: int | float
 ) -> list[dict[str, Any]]:
     issues: list[dict[str, Any]] = []
-    for element in (
-        element
-        for element in elements
-        if element["kind"] in {"table", "chart"}
-        or (element["kind"] == "shape" and element["type"] == "text")
-    ):
+    for element in elements:
```
fangshuyu-768, line 1222: **[P1] Account for rotation for every element kind checked here.** `detect_elements_out_of_canvas` now iterates over all elements, but `element_canvas_bbox` still applies the rotated bounding-box calculation only to charts and text shapes. For example, a 100x100 rect at `x=0` with `rotation=45` visibly extends about 20.7 px beyond the left edge, yet the lint returns `release_ready: true`. Rotated images and icons have the same gap. Please either apply the rotation calculation to every rotatable kind or explicitly exclude unsupported kinds, and add edge-rotation regression tests.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1336,56 +1332,617 @@ def lint_slide(
...
+def slide_content_visual_bbox(
+    element: dict[str, Any], slide_bbox: dict[str, int | float]
+) -> dict[str, int | float] | None:
+    if is_text_element(element):
+        estimated = estimate_text_visual_bbox(element)
+        return clipped_bbox(estimated, slide_bbox) if estimated else None
+    if element["kind"] == "shape" and has_text_content(element):
+        estimated = own_text_visual_bbox(element)
+        return clipped_bbox(estimated, slide_bbox) if estimated else None
+    if element["kind"] in {"img", "chart", "table", "whiteboard", "icon"}:
```
fangshuyu-768: **[P1] Invisible elements must not satisfy the visible-content rule.** This path does not filter `alpha <= 0`: a slide containing only `` is currently reported as `passed` instead of `blank_slide`. The same visibility omission also lets transparent image overlays or large visual children suppress sparse-container warnings. Please introduce a shared visibility predicate and use it consistently in blank-slide, slide-density, and container-density calculations (including icons, whose extracted density record currently omits alpha).

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1336,56 +1332,617 @@ def lint_slide(
...
+
+def issue_measurement(
+    issue: dict[str, Any], elements_by_id: dict[str, dict[str, Any]]
+) -> dict[str, Any]:
+    if issue.get("measurement") is not None:
+        return issue["measurement"]
+    if issue["code"] == "bbox_overlap" and len(issue.get("elements", [])) == 2:
+        left = elements_by_id.get(issue["elements"][0])
+        right = elements_by_id.get(issue["elements"][1])
+        if left and right:
```
fangshuyu-768, line 1899: **[P2] Measure the same visual geometry that triggered the overlap rule.** Detection can use estimated text visual bounds or horizontal text intrusion, but this code always measures the declared element boxes. The existing `test_lint_xml_detects_horizontal_text_overflow_across_declared_box_gap` therefore emits a `bbox_overlap` error whose `measurement.intersection_area` is `0`, contradicting `rule.comparison`. Please reuse the actual visual bounds used by detection, or emit dedicated intrusion metrics for the horizontal-overflow path.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1336,56 +1332,617 @@ def lint_slide(
...
+        normalized.setdefault(
+            "message",
+            f"slide visible content coverage {ratio:.1%} is below {threshold:.1%}",
+        )
+        normalized.setdefault(
+            "hint",
+            "Review the rendered screenshot to decide whether the page is intentionally sparse.",
+        )
+    else:
+        normalized.setdefault("message", normalized["code"].replace("_", " "))
```
fangshuyu-768, line 2019: **[P2] Ensure every schema-v2 diagnostic has a `hint`.** `validation-checklist.md` now promises that every error/warning contains `target`, `message`, and `hint`, but this fallback only supplies `message`; `bbox_overlap` is one current example with no `hint`. Please provide a default or rule-specific hint here and add a contract test that asserts the required keys for every normalized issue.

fangshuyu-768 (commented):
I found four actionable issues in the new layout-lint release gate. The existing test suite passes, but these cases leave gaps in the advertised detection and schema-v2 contract.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1336,645 @@ def lint_slide(
...
+        and element["kind"] == "shape"
+        and element["type"] == "rect"
+        and is_edge_spanning_layout_panel(element, slide_width, slide_height)
+        and contains(element, container, tolerance=DENSITY_CONTAINMENT_TOLERANCE)
+        for element in elements
+    )
+
+
+def extract_density_elements(slide_xml: str) -> list[dict[str, Any]]:
+    elements = extract_elements(slide_xml)
```
fangshuyu-768, line 1499: **[P1] Parse density elements directly from the XML tree.** This still bootstraps from the regex-based extractor and repairs shapes only when an explicit `id` is available. Valid XML using single-quoted attributes produces no density elements and is blocked as `blank_slide`; id-less self-closing shapes can also inherit the following shape's content because the repair lookup cannot match them. Both forms are valid here (`id` is optional/server-filled). Please build the geometry records from the ElementTree nodes directly and generate fallback IDs only after parsing.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1394,6 +1983,104 @@ def lint_xml(xml: str, source_path: str | None = None) -> dict[str, Any]:
...
+    for index, slide_xml in enumerate(presentation["slides"]):
+        slide_number = index + 1
+        geometry = lint_slide(
+            slide_xml,
+            slide_number,
+            presentation["width"],
+            presentation["height"],
+        )
+        density_elements = extract_density_elements(slide_xml)
+        extra_elements = [
```
fangshuyu-768, line 2110: **[P1] Include `polyline` in canvas validation.** The SXSD lists `polyline` as a supported slide data element, but neither `density_elements` nor `extra_elements` extracts it. A slide with normal visible text and a polyline at `x=900`, `width=120` on a 960px canvas currently returns `release_ready: true` with no `polyline_out_of_canvas` issue. Please extract its geometry, rotation, and alpha and include it in overflow and visibility handling.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1336,645 @@ def lint_slide(
...
+        touches_vertical_edge and element["width"] >= slide_width * LAYOUT_PANEL_SPAN_RATIO
+    )
+
+
+def has_matching_image_overlay(container: dict[str, Any], elements: list[dict[str, Any]]) -> bool:
+    container_area = element_area(container)
+    return any(
+        element["kind"] == "img"
+        and is_visually_rendered(element)
+        and intersection_area(container, element)
```
fangshuyu-768: **[P2] Require an image overlay to cover the container.** Dividing by `min(container_area, image_area)` treats every fully contained image as a matching overlay regardless of size. For example, a 20x20 image inside a 300x300 card gets a ratio of 1 and suppresses `sparse_container_content`, even though its actual container coverage is only 0.44%. Please compare overlap against `container_area`, or separately require the image and container areas to be similar.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1336,645 @@ def lint_slide(
...
+    if not is_visually_rendered(element):
+        return False
+    return element_area(element) / element_area(container) >= LARGE_VISUAL_CHILD_RATIO
+
+
+def detect_sparse_container_content(
+    elements: list[dict[str, Any]], slide_number: int, slide_width: int | float, slide_height: int | float
+) -> list[dict[str, Any]]:
+    issues: list[dict[str, Any]] = []
+    for container in (
```
fangshuyu-768, line 1687: **[P2] Exclude invisible containers and layout panels as well as invisible children.** The latest fix filters child visuals, but candidate containers are still unfiltered: an `alpha=0` large rect emits `sparse_container_content`. Conversely, an `alpha=0` edge-spanning rect can exempt a visible sparse card through `is_nested_in_layout_panel`. Please apply `is_visually_rendered` to both candidate containers and the panels used by exemption predicates.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1336,645 @@ def lint_slide(
...
-        error_count = sum(1 for issue in top_level_issues if issue["level"] == "error")
-        warning_count = sum(1 for issue in top_level_issues if issue["level"] == "warning")
-        info_count = sum(1 for issue in top_level_issues if issue["level"] == "info")
+
+def issue_measurement(
+    issue: dict[str, Any], elements_by_id: dict[str, dict[str, Any]]
+) -> dict[str, Any]:
+    if issue.get("measurement") is not None:
+        return issue["measurement"]
+    if issue["code"] == "bbox_overlap" and len(issue.get("elements", [])) == 2:
```
fangshuyu-768, line 1896: **[P2] Measure ordinary text overlaps with the visual bounds used by detection.** The new explicit measurement fixes only the horizontal-overflow branch. Ordinary text overlaps still fall through here and are calculated from declared shape boxes, although `should_flag_overlap` used `estimate_text_visual_bbox`. In a simple reproduction this reports area `13300` while the visual intersection that triggered the rule is about `1456`. Please reuse the estimated visual boxes so schema-v2 measurements describe the geometry that actually triggered the issue.

fangshuyu-768 (commented):
Re-review of the latest head: the rotation and schema-v2 hint issues are fixed. The alpha and overlap-measurement fixes are only partial, and the expanded regression pass found three additional gaps.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -79,8 +80,10 @@ def parse_args(argv: list[str]) -> dict[str, Any]:
 
 
 def extract_attribute(tag_source: str, name: str) -> str | None:
-    match = re.search(fr'{re.escape(name)}="([^"]+)"', tag_source)
-    return match.group(1) if match else None
+    match = re.search(fr"{re.escape(name)}=(?:\"([^\"]+)\"|'([^']+)')", tag_source)
+    if not match:
+        return None
+    return match.group(1) if match.group(1) is not None else match.group(2)
```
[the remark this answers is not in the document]
fangshuyu-768: Reopening because this remains reproducible at d191fc9. A well-formed element such as `<shape id = "visible" type = "text" ...>` is still dropped by geometry extraction, and the slide is incorrectly blocked as `blank_slide`. Please keep this thread open until optional whitespace around `=` is accepted, or the attributes are parsed from the XML tree.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1350,700 @@ def lint_slide(
...
+) -> bool:
+    # Deliberately permissive, unlike slide_content_visual_bbox: blank_slide is asking "is
+    # *anything* rendered here", not the richer "counts toward meaningful content density" bar
+    # that sparse_slide_content/sparse_container_content apply. A plain shape with no text (a
+    # decorative rect/ellipse/etc.), <undefined>, or any future SXSD data element should all
+    # count here — deny-list only what's actually invisible (alpha<=0 or zero on-canvas area)
+    # instead of maintaining an allow-list that silently treats unlisted kinds as blank.
+    if not is_visually_rendered(element):
+        return False
+    bbox = line_stroke_bbox(element) if element["kind"] == "line" else element
```
fangshuyu-768, line 1671: **[P1] Do not let a background-only slide bypass `blank_slide`.** This predicate treats every rendered shape as slide content, so a full-canvas background rectangle with an explicit fill produces `status: passed` and `release_ready: true` with no other content. That contradicts `validation-checklist.md`, which says an empty background or empty shape must not pass. Please exclude background/layout-only shapes here, or reconcile the implementation and the documented release contract.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -1330,62 +1350,700 @@ def lint_slide(
...
+
+
+def has_similar_short_card_peer(element: dict[str, Any], elements: list[dict[str, Any]]) -> bool:
+    return sum(
+        other is not element
+        and other["kind"] == "shape"
+        and other["type"] == "rect"
+        and other["width"] >= MIN_CONTAINER_WIDTH
+        and other["height"] >= MIN_SHORT_CARD_HEIGHT
+        and element_area(other) >= MIN_CONTAINER_AREA
```
fangshuyu-768, line 1428: **[P2] Ignore invisible cards when counting similar peers.** This count includes rectangles with `alpha="0"`. Two invisible 300x100 rectangles can therefore make one visible rectangle qualify as a repeated short-card container and emit `sparse_container_content`, forcing unnecessary screenshot review. Please require `is_visually_rendered(other)` when counting peers and add a regression test for transparent peers.

fangshuyu-768 (approved):

PR #2035, opened 2026-07-23
fix: omit zero total from chat search

arnold9672:
## Summary

Fix `im +chat-search` output when the upstream API omits `total` or returns
zero. The CLI now avoids reporting `total: 0` alongside non-empty results and
keeps positive upstream totals unchanged.

## Changes

- Omit `data.total` when the upstream value is missing, invalid, or zero.
- Use the number of displayed chats for the pretty-output footer when no
  positive upstream total is available.
- Add regression coverage for missing, invalid, zero, and positive totals.
- Add dry-run coverage for the chat-search request contract.

## Test Plan

- [x] Focused `shortcuts/im` contract tests pass.
- [x] Chat-search dry-run request contract test passes.
- [x] `go vet ./...` passes.
- [x] `gofmt -l` and `git diff --check` pass.
- [ ] `make unit-test` is fully green. The affected `shortcuts/im` package
  passes, but the full suite is blocked by unrelated host-environment failures:
  the test Git user config is overridden by the global config, and the system
  Git does not support `git init --initial-branch`.

## Related Issues

- None

sa: none
doc: none
cfg: none
test: unit test

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `im +chat-search` handling of the API-provided `total` when it’s missing, invalid, zero, or fractional.
  - `total` is now omitted from structured output unless it’s a positive value.
  - The “chat(s) found” message continues to display the parsed count from `total`.

- **Tests**
  - Added unit coverage to verify the `total` field contract across `json` and `pretty` modes.
  - Added an end-to-end dry-run CLI test validating the generated request, filters, and pagination parameters.

### DIFF
--- a/shortcuts/im/im_chat_search.go
+++ b/shortcuts/im/im_chat_search.go
@@ -108,7 +108,7 @@ var ImChatSearch = common.Shortcut{
 
 		rawItems, _ := resData["items"].([]interface{})
 		total, hasPositiveTotal := util.ToFloat64(resData["total"])
-		hasPositiveTotal = hasPositiveTotal && total > 0
+		hasPositiveTotal = hasPositiveTotal && int(total) > 0
 		hasMore, pageToken := common.PaginationMeta(resData)
 
 		// Extract MetaData from each item
--- a/shortcuts/im/im_chat_search.go
+++ b/shortcuts/im/im_chat_search.go
@@ -107,8 +107,8 @@ var ImChatSearch = common.Shortcut{
 		}
 
 		rawItems, _ := resData["items"].([]interface{})
-		totalF, _ := util.ToFloat64(resData["total"])
-		total := totalF
+		total, hasPositiveTotal := util.ToFloat64(resData["total"])
+		hasPositiveTotal = hasPositiveTotal && total > 0
 		hasMore, pageToken := common.PaginationMeta(resData)
 
 		// Extract MetaData from each item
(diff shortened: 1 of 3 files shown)


shortcuts/im/im_chat_search.go
```diff
@@ -144,10 +144,12 @@ var ImChatSearch = common.Shortcut{
 
 		outData := map[string]interface{}{
 			"chats":      items,
-			"total":      int(total),
 			"has_more":   hasMore,
 			"page_token": pageToken,
 		}
+		if hasPositiveTotal {
+			outData["total"] = int(total)
```
BytedanceSearch, line 151: **[P1] Apply the fallback to pretty output as well.** This only omits data.total from structured output. The pretty formatter still converts total unconditionally, so a response containing one chat with a missing or zero total still prints a row followed by 0 chat(s) found. That is the same inconsistency described by this PR. Please use len(items) when hasPositiveTotal is false and restore pretty-output regression coverage for missing and zero totals.

arnold9672: Using `len(items)` as a fallback is not appropriate here. It represents only the number of items returned on the current page, not the total number of matching chats, so displaying it as the total would be inaccurate (especially with pagination). This PR is intentionally scoped to omitting the zero-valued `total` field from structured output; it preserves the existing pretty-output behavior rather than substituting a different count with different semantics.

PR #2043, opened 2026-07-24
docs(doc): clarify formula rendering

### LINKED ISSUE #2042: docs(lark-doc): clarify formula rendering for XML document writes
## Summary

Clarify the `lark-doc` Skill guidance for mathematical formulas so AI agents do not write Markdown/LaTeX delimiters as literal text when the document command is using the default XML format.

## Problem

`lark-cli docs +create` and `docs +update` default to `--doc-format xml`. The XML reference currently lists `<latex>E = mc^2</latex>` in the inline-component table, but the format-dependent rule is easy to miss.

As a result, an agent may generate content such as:

```xml
<p>The result is $E = mc^2$.</p>
<p>$$\sum_{i=1}^{n} i$$</p>
```

In XML mode, `$...$`, `$$...$$`, `\(...\)`, and `\[...\]` are ordinary text. They can therefore appear verbatim in the Feishu/Lark document instead of becoming native formula nodes.

The correct XML is:

```xml
<p>The result is <latex>E = mc^2</latex>.</p>
<p><latex>\sum_{i=1}^{n} i</latex></p>
```

Markdown writes are different: `$...$` is valid only when the command explicitly uses `--doc-format markdown`.

## Proposed Change

Update the existing `lark-doc` Skill rather than adding a duplicate Skill:

1. Add a formula-routing rule to `skills/lark-doc/SKILL.md`.
2. Add a dedicated formula section to `references/lark-doc-xml.md` covering:
   - conversion from `$...$`, `$$...$$`, `\(...\)`, and `\[...\]` to `<latex>...</latex>`;
   - inline versus standalone-paragraph examples;
   - XML escaping for `&`, `<`, and `>` inside formulas;
   - avoiding `<pre>` / `<code>` for formulas;
   - relative `@file` input for content containing `$` and backslashes.
3. Add matching Markdown guidance to `references/lark-doc-md.md`.
4. Recommend a post-write `docs +fetch --doc-format xml --detail full` check to verify that the content contains `<latex>` nodes rather than literal delimiters.

## Acceptance Criteria

- The Skill clearly distinguishes XML formula syntax from Markdown formula syntax.
- Agents are explicitly told not to write Markdown/LaTeX delimiters verbatim in XML mode.
- Examples cover inline and standalone formulas.
- Shell-safe relative `@file` input is recommended for formula-heavy content.
- The repository's Skill format check passes.

## Validation

I prepared the documentation-only change locally and verified it with:

```text
node scripts/skill-format-check/index.js
✅ Skill format check passed!

git diff --check
✅ no whitespace errors
```

I can open a PR with the prepared patch once a fork branch is available.

Alpha-Girl:
## Summary

Clarify formula authoring in the `lark-doc` Skill so agents do not write Markdown/LaTeX delimiters as literal text when `docs +create` or `docs +update` uses the default XML format.

The root cause is that native `<latex>...</latex>` support was only mentioned in the XML component table, while the format-dependent conversion rule was not prominent. This change makes the XML and Markdown paths explicit and adds a post-write verification step.

## Changes

- Add a formula-routing rule to `skills/lark-doc/SKILL.md`.
- Document XML conversion from `$...$`, `$$...$$`, `\(...\)`, and `\[...\]` to native `<latex>` nodes.
- Add inline and standalone formula examples, XML escaping guidance, and relative `@file` input guidance.
- Clarify Markdown formula syntax and recommend XML fetch verification after writes.

## Test Plan

- [ ] Unit tests pass — not run; this is a documentation-only Skill change.
- [x] `git diff --check`
- [ ] Manual live-tenant verification of the `lark-cli docs +create/+update` flow — not run; requires Feishu/Lark tenant access.

## Related Issues

- Closes #2042

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for writing and exporting Markdown content, including mathematical formulas, escaping, images, and unsupported block types.
  * Documented XML formula syntax using `<latex>...</latex>`, including escaping and validation requirements.
  * Added quick-decision guidance for selecting the appropriate formula format and safely passing formula content.

### DIFF
--- a/skills/lark-doc/SKILL.md
+++ b/skills/lark-doc/SKILL.md
@@ -49,6 +49,7 @@ lark-cli docs +update --doc "文档URL或token" --command append --content '<p>�
 - 用户明确要操作思维笔记时；已有**思维笔记**，走 [思维笔记链路](references/lark-doc-mindnote.md)；新建**思维笔记**，走 [lark-doc-whiteboard](references/lark-doc-whiteboard.md)
 - 拿到 spreadsheet URL/token 后 → 切到 `lark-sheets` 做对象内部操作
 - 用户需要统计文档的**总字数 / 总字符数**（word count / character count）时，先读取 [`lark-doc-word-stat.md`](references/lark-doc-word-stat.md)，并按其中流程调用 [`scripts/doc_word_stat.py`](scripts/doc_word_stat.py)；统计口径以该脚本为准，不要改用其他方式自行计算。
+- 文档包含数学公式时，先按所选格式读取对应公式规则：XML 使用 `<latex>...</latex>`，**禁止**把 `$...$`、`$$...$$`、`\(...\)` 或 `\[...\]` 原样写入 XML；Markdown 才使用 `$...$`。含 `$` 或反斜杠的内容优先通过相对 `@file` 传入，避免 Shell 改写公式。详见 [`lark-doc-xml.md`](references/lark-doc-xml.md#公式) / [`lark-doc-md.md`](references/lark-doc-md.md#数学公式)。
 - 用户说"给文档加评论""查看评论""回复评论""给评论加/删除表情 reaction" → 切到 `lark-drive` 处理
 - 文档内容中出现嵌入的 `<sheet>`、`<bitable>` 或 `<cite file-type="sheets|bitable">` 标签时 → **必须主动提取 token 并切到对应技能下钻读取内部数据**，不能只呈现标签本身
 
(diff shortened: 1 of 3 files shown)


SunPeiYang996:
Thanks for the contribution and for documenting this edge case.

The issue is valid, but it appears to be a very low-probability edge case. We do not want to keep patching the Skill for every boundary scenario: doing so would make the Skill unnecessarily large and increase maintenance and model-reading costs.

For now, we will not merge this PR and will close it. If we keep a reminder, please reduce the change to a single sentence:

> Markdown syntax is not supported in XML mode; we plan to improve this in the next release.

We can revisit the documentation when XML mode supports Markdown syntax at the product level. Thanks for understanding.

PR #2076, opened 2026-07-28
feat: add lark router and setup-lark-skills skills for opt-in skill isolation

AKAZIK-py:
## Summary

Users running many agent harnesses (Claude Code, pi, Kimi Code, etc.) currently get all 27 `lark-*` skill descriptions (~11.8 KB) injected into every session's system prompt, whether or not the session involves Lark. This PR adds two infrastructure skills — a router (`lark`) and a configurator (`setup-lark-skills`) — that give users an **opt-in** path to isolate the skill family behind explicit `/skill:` invocation, without changing the default auto-injected behavior of any existing skill.

## Changes

- **`skills/lark/SKILL.md`** (new): the authoritative routing index over all 27 `lark-*` skills — a 27-row intent→skill table, a dedicated disambiguation section for meeting-related skills (the highest-collision boundary: `lark-vc` / `lark-vc-agent` / `lark-minutes` / `lark-note` / `lark-workflow-meeting-summary`), and multi-intent chaining guidance. Marked `disable-model-invocation: true` so the router itself is user-invoked.
- **`skills/setup-lark-skills/SKILL.md`** (new): a configurator that lets users switch low-frequency `lark-*` skills to dormant (insert `disable-model-invocation: true`) or restore them, with idempotent, frontmatter-scanning commands (works regardless of where the harness installed the skills; includes install-path detection and Linux/macOS `sed` notes). Also marked `disable-model-invocation: true`.
- No changes to existing files. `disable-model-invocation` is an optional Agent Skills frontmatter field — harnesses that don't recognize it ignore it (verified against pi, which honors it natively, and this repo's own non-strict YAML frontmatter parser).

## Test Plan

- [x] `node scripts/skill-format-check/index.js skills` passes (29 skill dirs, no errors/warnings)
- [x] Route table verified 27/27 against `lark-cli skills list` (v1.0.71) and against `upstream/main` `skills/` directory — no drift
- [x] Description-volume figure re-measured: 27 descriptions ≈ 11.8 KB (mean ≈ 440 B), matching the doc
- [x] Markdown-only change; no Go code touched (`git diff --stat`: 2 new files only)
- [ ] `make unit-test` — not run locally (no Go toolchain on this machine); change does not touch Go sources, CI should be unaffected
- [x] The router + isolation setup has been battle-tested locally for a week across three agent harnesses, including an independent grading harness (intent→skill routing accuracy scored 100/100 by a separate examiner agent)

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Added a Lark/飞书 skill router for `/skill:lark`, including direct `/skill:lark-<name>` shortcuts, an agent fallback when skills aren’t loaded, and clearer meeting-related disambiguation (with a rule to avoid using it for non-飞书 requests).
  * Introduced `/skill:setup-lark-skills` to batch-manage `lark-*` skills’ “activated vs sleeping” behavior.
* **Documentation**
  * Documented strict detection and idempotent frontmatter editing for `disable-model-invocation`, plus preset selection profiles and restart/usage guidance.

### DIFF
--- a/skills/lark/SKILL.md
+++ b/skills/lark/SKILL.md
@@ -1,0 +1,65 @@
+---
+name: lark
+version: 1.0.0
+description: "飞书/Lark 能力路由器：27 个 lark-* skill 的权威用途索引与调用方式。当不确定某个飞书需求该用哪个 skill、或一个任务需要组合多个飞书 skill 时，输入 /skill:lark 加载本文件查询。"
+disable-model-invocation: true
+metadata:
+  role: router
+---
+
+# Lark 能力路由
+
+本文件是 27 个 lark-* skill 的唯一权威路由索引（description 不复述触发词，避免双写漂移）。27 个 skill 默认全部自动注入（model-invoked）；想把飞书 skill 与日常开发隔离的用户，可用 `setup-lark-skills` 将低频 skill 转为 user-invoked 休眠。
+
+## 调用方式
+
+- **查路由**：输入 `/skill:lark` 加载本文件，按下表选择。
+- **直接用**：已知目标时输入 `/skill:lark-<name>`，无需经过本文件。
+- **Agent 备用路径**：agent 在用户明确要求飞书能力、但相关 skill 未加载时，可直接 read 本文件获取路由，再 read 目标 skill 的 SKILL.md 执行（安装路径因 harness 而异，探测方法见 `setup-lark-skills`）。
+- **隔离配置**：运行 `/skill:setup-lark-skills` 把低频 skill 转为 user-invoked 休眠、或恢复自动注入（重启会话后生效）。
+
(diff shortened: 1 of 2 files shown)


AKAZIK-py:
Both open threads addressed in b56a540:

1. **CRLF (Major)**: all three awk scripts now strip a trailing `\r` into a normalized copy before matching (`/^---$/` on a CRLF delimiter previously never matched, so CRLF files were silently skipped). Output always prints the original line, and inserted `disable-model-invocation` lines follow the file's own line-ending style. Verified: CRLF files now get correct status detection, sleep/activate edits, idempotency, and the strict duplicate-field normalization — 17/17 fixture tests including CRLF/LF mixed and dirty-data cases.
2. **nullglob (Minor)**: the status scan enables `nullglob` and aborts with a clear error when `$ROOT/lark-*/SKILL.md` matches nothing (partial/missing install), instead of scanning a literal `*` path and reporting a fake `*: 激活`.

PR #2078, opened 2026-07-28
fix(base): align workflow validation and agent guidance

huarenmin13:
## Summary

Fix Base workflow cases 014 and 026 without adding a general workflow schema validator.

- **014:** Document `DeleteRecordTrigger` as known unsupported. When a requested capability is unavailable, agents must preserve the original intent, stop writes, and execute alternatives only after explicit user selection.
- **026:** Validate `LarkMessageAction.data` before workflow create/update: `data` must be an object, `receiver` and `content` must be non-empty arrays, and optional `send_to_everyone`/`btn_list` fields are type-checked when present.

## Changes

- Share one create/update validator for the `LarkMessageAction` contract and return typed `--json` field errors before transport.
- Preserve valid request bodies and leave non-target or unknown step shapes to server validation.
- Align workflow get/update guidance and the delivered Skill schema.
- Keep the effective diff within existing files.

## Validation

- [x] `make build`
- [x] `go test ./shortcuts/base -count=1`
- [x] Current-binary delivered-Skill E2E tests
- [x] `go vet ./shortcuts/base ./tests/cli_e2e/base`
- [x] `gofmt` and `git diff --check`

The opt-in live workflow round-trip was not run.

## Related

Base workflow evaluation cases 014 and 026.


huarenmin13:
@coderabbitai review

huarenmin13:
@coderabbitai review

PR #2089, opened 2026-07-28
fix(drive): use title for default download filename

wittam-01:
## Summary
Fix `drive +download` default naming when `--output` is omitted. The command now resolves metadata title before starting the download and saves with `Content-Disposition` filename > metadata title > token.

## Changes
- Add conditional metadata title lookup for omitted `--output`, so missing metadata permission fails before the download stream starts.
- Resolve the default save name from response filename, metadata title, then token while keeping explicit `--output` behavior unchanged.
- Add unit tests, dry-run E2E coverage, and update the lark-drive download reference.

## Test Plan
- [x] Unit tests pass: `go test ./shortcuts/drive -run 'TestDriveDownload(UsesContentDispositionWhenOutputOmitted|FallsBackToMetadataTitleWhenOutputOmitted|FallsBackToTokenWhenOutputOmittedAndMetadataEmpty|MetadataErrorBeforeDownloadWhenOutputOmitted|RejectsOverwriteWithoutFlag|AllowsOverwriteFlag)' -count=1`\n- [x] Dry-run E2E passes with a locally built binary: `LARK_CLI_BIN=/Users/bytedance/go/src/github.com/larksuite/cli/tmp/lark-cli-drive-download-default go test ./tests/cli_e2e/drive -run 'TestDriveDownloadDryRun_(DefaultNamePlansMetadataBeforeDownload|ExplicitOutputSkipsMetadata)' -count=1`\n- [x] `git diff --cached --check`\n\n## Related Issues\n- None\n

## Summary by CodeRabbit

- **New Features**
  - When `--output` is omitted, Drive downloads now derive a safe default filename/path (sanitizing `Content-Disposition`, including slash-only and Windows-reserved characters), with priority: `Content-Disposition` filename → Drive metadata title → file token.
  - Metadata planning/scope checks are performed only for the default-output flow; explicit `--output` skips them.
  - Overwrite protection now applies to the computed default output (unless `--overwrite` is set).
  - Drive export filename sanitization now treats Windows reserved device names as invalid and falls back.
- **Documentation**
  - Updated `drive +download` docs for the token-only default save location.
- **Tests**
  - Expanded unit and CLI E2E coverage for default naming, dry-run behavior (API call order), scope enforcement, and related validation/error handling.


skills/lark-drive/references/lark-drive-download.md
```diff
@@ -11,10 +11,23 @@
 # 下载到指定路径
 lark-cli drive +download --file-token boxbc_xxx --output ./report.pdf
 
-# 只提供 token，默认保存为当前目录下同名文件
+# 只提供 token，默认保存到当前目录
 lark-cli drive +download --file-token boxbc_xxx
 ```
 
+省略 `--output` 时，CLI 会先查询文件标题，再下载文件；本地文件名优先级为：
```
wittam-01: 下面这些skill描述可以去掉，不需要

shortcuts/drive/drive_download.go
```diff
@@ -16,47 +18,112 @@ import (
...
 	Scopes:      []string{"drive:file:download"},
-	AuthTypes:   []string{"user", "bot"},
+	// Metadata is only required when --output is omitted and the CLI needs the
+	// remote title as the pre-download fallback filename.
+	ConditionalScopes: []string{driveMetadataReadScope},
+	AuthTypes:         []string{"user", "bot"},
 	Flags: []common.Flag{
 		{Name: "file-token", Desc: "file token", Required: true},
-		{Name: "output", Desc: "local save path"},
+		{Name: "output", Desc: "local save path; omit to save in the current directory using Content-Disposition filename, metadata title, then token"},
```
wittam-01: 不需要加这个描述哈，改回去，内部兼容逻辑，不需要让用户感知

skills/lark-drive/references/lark-drive-download.md
```diff
@@ -11,7 +11,7 @@
 # 下载到指定路径
 lark-cli drive +download --file-token boxbc_xxx --output ./report.pdf
 
-# 只提供 token，默认保存为当前目录下同名文件
+# 只提供 token，默认保存到当前目录
 lark-cli drive +download --file-token boxbc_xxx
```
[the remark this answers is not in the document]
wittam-01: 这里不采纳。人工 review 已明确要求不要在 skill/help 暴露内部兼容策略；本 PR 保留简洁用法说明，权限缺失由 typed error / ConditionalScopes 反馈给调用方。

shortcuts/drive/drive_download.go
```diff
@@ -16,47 +18,112 @@ import (
...
-		if _, resolveErr := runtime.ResolveSavePath(outputPath); resolveErr != nil {
-			return errs.NewValidationError(errs.SubtypeInvalidArgument, "unsafe output path: %s", resolveErr).WithParam("--output")
-		}
-		if _, statErr := runtime.FileIO().Stat(outputPath); statErr == nil && !overwrite {
-			return errs.NewValidationError(errs.SubtypeInvalidArgument, "output file already exists: %s (use --overwrite to replace)", outputPath).WithParam("--output")
+		var metadataTitle string
+		if outputPath == "" {
+			title, err := common.FetchDriveMetaTitle(runtime, fileToken, "file")
+			if err != nil {
+				return err
```
fangshuyu-768: [P2] Keep downloads working when optional title lookup fails

`FetchDriveMetaTitle` is only needed for fallback naming, but returning every metadata error here prevents the actual file GET when `metas.batch_query` is rate-limited or temporarily unavailable, even though `Content-Disposition` or the token can still provide a safe filename. Please preserve the intentional missing-scope error if required, but for non-permission metadata failures emit a warning and continue with the existing Content-Disposition/token fallback, as the adjacent export flow does.

shortcuts/drive/drive_download.go
```diff
@@ -16,47 +18,112 @@ import (
...
+}
+
+func driveDownloadDefaultOutputPath(header http.Header, title, fileToken string) string {
+	fallback := driveDownloadFallbackFileName(title, fileToken)
+	fileName := common.ResolveDownloadFileName(header, fallback)
+	if strings.TrimSpace(fileName) == "" || fileName == "." || fileName == ".." || strings.Trim(fileName, "/") == "" {
+		fileName = fileToken
+	}
+	fileName, _ = common.AutoAppendDownloadExtension(fileName, header, "")
+	return fileName
```
fangshuyu-768: [P2] Sanitize default filenames for Windows

The new default path preserves remote title/header characters other than path separators. A valid Drive title or `Content-Disposition` value such as `Q1: forecast?.txt` therefore reaches `FileIO.Save` unchanged and fails on the supported Windows build, where `:` and `?` are not valid filename characters; the previous token-based default did not have this failure mode. Please run the derived name through the Drive filename sanitizer (or an equivalent cross-platform sanitizer) before extension resolution and add a regression case for Windows-reserved characters.

shortcuts/drive/drive_download.go
fangshuyu-768: [P2] Preserve cancellation from metadata lookup

Cancellation and deadline errors should remain terminal even though other non-authorization metadata failures can safely fall back. `CallAPITyped` preserves the underlying cause, so a canceled metadata request currently reaches this branch as a non-authorization error, is downgraded to a warning, and then proceeds into the download path. Please check `errors.Is(err, context.Canceled)` and `context.DeadlineExceeded` before applying the fallback, and add a regression test asserting that no download request is attempted after cancellation.

shortcuts/drive/drive_download.go
fangshuyu-768: [P2] Also reject Windows reserved device names

`sanitizeExportFileName` now replaces invalid Windows characters, but it still returns reserved device names such as `CON.txt`, `NUL`, or `COM1.pdf` unchanged. Those names fail at `FileIO.Save` on Windows, so a remote title or `Content-Disposition` filename can still trigger the same regression. Please reject reserved device basenames case-insensitively, before any extension, and add regression cases.

shortcuts/drive/drive_download.go
fangshuyu-768: [P2] Keep ordinary metadata timeouts on the fallback path

The `CategoryNetwork` / `network_timeout` check is broader than caller cancellation: `WrapDoAPIError` classifies SDK `ServerTimeoutError` and `ClientTimeoutError` this way even though neither wraps `context.Canceled` nor `context.DeadlineExceeded`. Those are transient failures of the optional title lookup, so this condition reintroduces the original regression by preventing the file GET even when `Content-Disposition` or the token can still name the download. The `errors.Is` checks above already make caller cancellation and deadlines terminal. Please keep ordinary SDK timeouts on the warning-and-fallback path and add a regression case proving that the download still runs.

shortcuts/drive/drive_download.go
```diff
@@ -16,47 +19,132 @@ import (
...
+
+func driveDownloadDefaultOutputPath(header http.Header, title, fileToken string) string {
+	fallback := driveDownloadFallbackFileName(title, fileToken)
+	fileName := common.ResolveDownloadFileName(header, fallback)
+	if strings.TrimSpace(fileName) == "" || fileName == "." || fileName == ".." || strings.Trim(fileName, "/") == "" {
+		fileName = fileToken
+	}
+	fileName = sanitizeExportFileName(fileName, fileToken)
+	fileName, _ = common.AutoAppendDownloadExtension(fileName, header, "")
+	return fileName
```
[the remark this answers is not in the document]
fangshuyu-768: [P2] Validate every derived candidate before moving down the fallback chain

This remains broader than the `CON.txt` case: a Content-Disposition filename or metadata title containing U+202E or another character rejected by `ResolveSavePath` passes the current sanitizer, then line 163 returns a `--output` validation error even though the user never supplied that flag. Please normalize, sanitize, and path-validate the header and metadata candidates independently, choose the first safe value in header → metadata title → token order, and reserve `--output` errors for explicit user paths.

tests/cli_e2e/drive/drive_download_dryrun_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
```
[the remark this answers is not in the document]
fangshuyu-768: [P2] Cover the omitted-output branch in the live workflow

The existing `TestDrive_UploadWorkflow` downloads with `--output downloaded.txt`, so the latest head still does not exercise the behavior changed by this PR against the live backend. Please extend that self-contained upload workflow, or add an equivalent one, to download the named fixture without `--output`, assert the actual saved basename and content, and clean up the local artifact. This is needed to verify the real metadata and Content-Disposition behavior required by `AGENTS.md` for shortcut behavior changes.

shortcuts/drive/drive_download.go
```diff
@@ -16,47 +19,180 @@ import (
...
+				lastErr = err
+				continue
+			}
+		}
+		return fileName, nil
+	}
+	if lastErr != nil {
+		return "", lastErr
+	}
+	return fileToken, nil
```
fangshuyu-768, line 88: [P2] Fail when no default filename candidate is safe

`fileToken` has already been tried as the final candidate in the loop, so returning it raw here bypasses both sanitization and `validatePath`. For example, `validate.ResourceName("/", "--file-token")` currently succeeds; with no usable header or metadata title, this helper returns `("/", nil)`, so the command misses the `cannot derive a safe default output path` branch and only fails later during `Stat` or `Save`. Please return an error when the candidate loop is exhausted, or tighten token validation so this fallback can never be unsafe.

skills/lark-drive/references/lark-drive-download.md
```diff
@@ -11,7 +11,7 @@
 # 下载到指定路径
 lark-cli drive +download --file-token boxbc_xxx --output ./report.pdf
 
-# 只提供 token，默认保存为当前目录下同名文件
+# 只提供 token，默认保存到当前目录
```
fangshuyu-768, line 14: [P2] Document the new default-name and permission contract

Omitting `--output` now performs a metadata request that may fail with `missing_scope`, and the resulting filename follows `Content-Disposition` → metadata title → file token. Neither `--help` nor this reference explains those details, so users with download-only credentials cannot predict the reauthorization requirement or which file will be created. Please document the filename precedence and the conditional `drive:drive.metadata:readonly` scope here.

wittam-01: 只是内部兼容逻辑变更，不需要在这里专门说明，预期绝大部分bot都已经有drive:drive.metadata:readonly了，极端场景没有也会自动提示

shortcuts/drive/drive_download.go
wittam-01: 不需要处理，file-token 传 / 下载就会报错

fangshuyu-768 (approved):

PR #2091, opened 2026-07-29
feat(sheets): error prescription overhaul, batch-update hardening, and read offload

xiongyuanwen-byted:
## Summary

Aggregates the sheets work on `feat/lark-sheets-develop` since #1833, driven by eval-batch error attribution (07-11 / 07-21 / 07-28 batches). Four themes:

### Error messages as prescriptions
- Inline schema hints on validation failure — the error itself teaches the expected shape.
- Collect all validation errors in one pass instead of failing on the first.
- Errors and help now name the enum, the required selector, and the missing envelope explicitly.
- Reachable border-vocabulary normalization (previously dead code), plus prescriptive fixes for flag and style-field mistakes.

### Batch-update contract hardening (#2028)
- Contract hardening and style-vocabulary acceptance for `+batch-update`, from a dedicated 137-error attribution round and five re-run validations, including a cross-model batch (Doubao).

### Read-path improvements
- `--include` truncation support on `+cells-get`.
- `--output-path` for full-read file offload on `cells` / `csv` / `table-get`, avoiding stdout truncation on large reads.

### Freeze semantics (verified live)
Freeze is full-state replacement server-side, not a per-axis patch. Verified on a real sheet: freezing 1 row and then 2 columns in two calls ends at `frozen_row_count 0 / frozen_column_count 2` — the second call drops the first axis.

- `+workbook-create` / `+table-put` / `+styles-put` now emit `freeze: {rows, cols}` as ONE operation. Two per-axis operations previously left only the last axis frozen.
- `+dim-freeze` gains `--rows` / `--cols`, which state the complete freeze state in one call. This is the only way to hold both axes, and the only way at all inside `+batch-update`, whose sub-ops are a static array that cannot read current state (`+styles-put`, the other carrier of a combined freeze, is not batchable).
- The single-axis `--dimension` / `--count` pair is a strict subset — every call has a byte-identical `--rows`/`--cols` spelling, pinned by `TestDimFreezeEquivalent`. It keeps working but is retired from the skill docs and `--help`, and every use (standalone, `--dry-run`, and each `+batch-update` sub-op) prints its exact replacement.

### --styles range prefix handling
A `Sheet!` prefix on a `--styles` range was only normalized for `+styles-put` `cell_styles`, so `row_sizes` like `"Sheet1!2:3"` failed `parseA1Range` on all three carriers. Normalization moved into the shared item parser, so `+workbook-create` / `+table-put` / `+styles-put` behave alike; a prefix naming a different sheet is stripped and reported alongside the item's other issues instead of masking them.

### Correctness fixes
- `+dim-insert --inherit-style` side mapping corrected (backend always inherits the anchor column's style; `side` only picks placement): `before` → `side=after` + `pos-1`, `after` → `side=before`; redundant `none` dropped.
- Reject path-shaped `--csv` values instead of silently writing them into the sheet.
- Skill docs and flag data resynced from sheet-skill-spec.
- `DEPRECATED(phase-2)` marks the phase-1 deprecations that are still executable, so `grep -rn "DEPRECATED(phase-2)"` enumerates what is pending removal; each carries its removal checklist.

### Cross-repo dependency
`shortcuts/sheets/data/**` and `skills/lark-sheets/**` are sync artifacts of `sheet-skill-spec`. This PR depends on that repo's `develop` commits `4888147` (flag-level doc retirement via `doc_hidden_flags`) and `40aa06c` (`+dim-freeze --rows/--cols`); both are pushed. Landing this PR against an older spec state would let the next `sync:consumers` revert the new flags.

`origin/main` has been merged back into this branch (conflicts resolved in `internal/validate/path.go` and `internal/vfs/localfileio/path.go`, keeping both `SafeTempAbsInputPath` and `LocalInputPath`).

### Rollback guide
The PR aggregates several independent themes; if something regresses, revert the theme rather than the PR:

| Theme | Where |
| --- | --- |
| Error prescriptions / schema hints | `flag_schema_validate.go`, `flag_ergonomics.go`, `style_vocab.go` |
| `+batch-update` contract | `batch_op_dispatch.go`, `lark_sheet_batch_update.go` (see #2028) |
| Read offload / truncation | `read_output.go`, `lark_sheet_read_data.go`, `lark_sheet_table_io.go` |
| `+styles-put` and `--styles` payloads | `lark_sheet_styles_put.go`, `lark_sheet_workbook.go` |
| Freeze semantics | `lark_sheet_sheet_structure.go` (`dimFreezeInput`), `workbookCreateVisualOpInput` |
| Docs / flag data | `skills/**`, `shortcuts/sheets/data/**` (regenerate from sheet-skill-spec, do not hand-edit) |

Phase-1 deprecations that remain executable are tagged `DEPRECATED(phase-2)`; each carries its own removal checklist.

## Test plan
- `go build ./...`, `go vet ./shortcuts/... ./internal/...` clean.
- `go test ./shortcuts/... ./internal/... ./tests/cli_e2e/sheets/...` all pass; package coverage for `shortcuts/sheets` is 84.9%.
- `go test ./shortcuts/sheets/...`, `./internal/vfs/localfileio/...`, `./internal/validate/...` all pass post-merge.
- Behavior changes validated through repeated eval-batch re-runs (see #2028 for the batch-update rounds).
- Freeze semantics established against a live sheet rather than inferred: two-call vs one-call ordering, omitted-axis behaviour, and the `+batch-update` sub-op path were each executed and read back via `grid_properties.frozen_row_count / frozen_column_count`.

## Summary by CodeRabbit

* **New Features**
  * Added declarative `+styles-put` (styles, merges, row/column sizing, freezes).
  * Added multi-region `+cells-set --writes` and multi-range `+dim-delete --ranges`.
  * Added `+chart-create --print-example` and improved `--print-schema --flag-name` for dotted paths.
* **Bug Fixes**
  * Improved high-risk confirmation to provide safer `re-run ... --yes` hints when appropriate.
  * Hardened `--csv`/file-path safety for “forgotten @ vs stdin” cases; improved validation/error flattening and aggregation.
  * Improved `--output-path` behavior (max-chars handling, truncation warnings) and unsafe-path reporting.
* **Documentation**
  * Updated Sheets flags, schemas, and references for new options and clarified `+batch-update` fail-fast/no-rollback semantics and `+cells-batch-set-style` deprecation.


internal/cmdutil/confirm.go
```diff
@@ -14,12 +18,106 @@ import (
...
+	if len(args) == 0 {
+		return ""
+	}
+	parts := make([]string, 0, len(args)+1)
+	parts = append(parts, filepath.Base(args[0]))
+	for _, a := range args[1:] {
+		if argReadsStdin(a) || argCarriesSecret(a) {
+			return ""
+		}
+		parts = append(parts, shellQuoteArg(a))
```
fangshuyu-768: **[P1] Avoid echoing free-form payloads into confirmation errors**

The guard only detects credential-looking flag names, but high-risk commands also accept sensitive free-form values such as `apps +db-execute --sql` and `base +form-submit --json`. A short inline password, SQL literal, or PII payload is therefore copied into the confirmation envelope, stderr, and downstream logs. Please avoid rendering arbitrary argv values here—either whitelist known-safe flags or fall back to the generic `add --yes` hint for payload-bearing invocations.

shortcuts/sheets/sheet_ai_api.go
```diff
@@ -100,6 +101,66 @@ func callTool(
...
+		out := detail.Message + " — " + strings.Join(parts, "; ")
+		// Partial failure is NOT rolled back server-side: the succeeded sub-ops
+		// stay applied. Spell out the recovery so agents don't resend the whole
+		// batch and double-apply the successes (observed in eval traces). With
+		// one failure (fail-fast) everything before it succeeded and nothing
+		// after it ran — resend from that index; with several (continue-on-error)
+		// only the listed failures need resending.
+		if strings.Contains(detail.Message, "succeeded") &&
+			!strings.Contains(detail.Message, " 0 succeeded") {
+			if len(detail.Failures) == 1 {
```
fangshuyu-768: **[P1] Do not infer fail-fast execution from a single failure**

A single failure does not imply fail-fast: `+batch-update --continue-on-error` can report one failed operation while later operations have already succeeded. Prescribing `operations[i:]` in that case replays those later successes and can double-apply inserts, deletes, or moves. Please pass the execution mode into this formatter, or emit recovery guidance that distinguishes failed operations from the not-run tail.

skills/lark-sheets/references/lark-sheets-styles-put.md
```diff
@@ -0,0 +1,91 @@
+# Lark Sheet Styles Put（+styles-put）
+
+> **本文定位**：对**已有**表格做美化收尾的默认入口——样式 / 边框 / 合并 / 行高列宽 / 冻结写成一份声明式规格，一次调用原子交付。样式**取什么值**（配色 / 字号 / 对齐 / 数字格式标准）以 `lark-sheets-visual-standards` 为唯一权威，本文只讲**怎么落地**。
+>
+> **边界（三分流判定，按操作组合选入口）**：目标是**样式 / 合并 / 行高列宽 / 冻结**的任意组合 → 本命令；**同一个写操作**打多个区域（如多区域清除、批量下拉）→ 用该命令自身的复数形态（`--ranges` / map 入参）；操作链**跨类型且有顺序依赖**（如插列 → 写表头 → 回填数据）→ `+batch-update`。美化收尾不需要也不应该拼 `--operations` 子操作数组。
+
+## 使用场景
+
+写入。对存量表格的多个子表批量应用视觉规格：新表美化、加汇总行后统一版式、按分组合并同类单元格、调列宽行高、冻结表头。整份规格展开为一次原子批量提交，全部生效或全部不生效；纯样式盖章可安全重放（同一份规格重发无副作用）。
```
fangshuyu-768: **[P1] Stop promising transactional batch semantics**

`+styles-put` is implemented with the same `batch_update` path that this PR documents as fail-fast without rollback. If operations for an earlier sheet succeed and a later sheet is missing or otherwise fails, the earlier changes remain applied, contradicting “all or nothing” and making whole-spec retries unsafe. Please replace the atomic/all-or-nothing wording here and in the matching help for the new batch-backed helpers with explicit no-rollback and recovery semantics.

shortcuts/sheets/lark_sheet_styles_put.go
```diff
@@ -0,0 +1,272 @@
...
+		if name == "" {
+			probs = append(probs, common.ValidationErrorf("%s.name is required (the real sheet name; check +workbook-info)", path))
+			continue
+		}
+		if seenName[name] {
+			probs = append(probs, common.ValidationErrorf("%s.name %q appears twice; merge the two items", path, name))
+			continue
+		}
+		seenName[name] = true
+		payload, itemProbs := parseWorkbookCreateStyleItem(item, path)
```
fangshuyu-768: **[P2] Reject unknown styles item keys before expansion**

`parseWorkbookCreateStyleItem` ignores unrecognized top-level keys. For example, an item containing valid `cell_styles` plus `freezee: {"rows":1}` succeeds, applies the styles, and silently drops the requested freeze. Please reject keys outside `name`, `cell_styles`, `row_sizes`, `col_sizes`, `cell_merges`, and `freeze`, ideally with a did-you-mean hint.

shortcuts/sheets/lark_sheet_styles_put.go
```diff
@@ -0,0 +1,272 @@
...
+			totalCells += int64(rows) * int64(cols)
+			if err := checkBatchStampBudget(totalCells); err != nil {
+				return nil, err
+			}
+			ops = append(ops, map[string]interface{}{
+				"tool_name": "set_cell_range",
+				"input": map[string]interface{}{
+					"excel_id":   token,
+					"sheet_name": spec.name,
+					"range":      stripSheetPrefix(cs.Range),
```
fangshuyu-768, line 165: **[P2] Validate sheet-prefixed ranges before stripping them**

This silently discards the range’s sheet prefix without checking it against `spec.name`. An item with `name: "Summary"` and `range: "Detail!A1:D1"` is therefore applied to `Summary!A1:D1` and exits successfully. Please reject prefixed ranges or require the prefix to match the item’s target sheet before removing it.

shortcuts/sheets/read_output.go
```diff
@@ -0,0 +1,75 @@
...
+	if path == "" {
+		runtime.Out(out, nil)
+		return nil
+	}
+	b, err := json.MarshalIndent(out, "", "  ")
+	if err != nil {
+		return err
+	}
+	b = append(b, '\n')
+	if _, err := runtime.FileIO().Save(path, fileio.SaveOptions{}, bytes.NewReader(b)); err != nil {
```
fangshuyu-768, line 171: **[P2] Preserve typed errors for read-output file failures**

Returning the raw `FileIO.Save` error causes an unsafe path such as `--output-path ../../outside.json` to fall through as `internal/unknown` instead of the required validation error, while write failures also lose the `file_io` subtype. Please return `common.WrapSaveErrorTyped(err)` here and add assertions for the typed category, subtype, and preserved cause.

fangshuyu-768 (dismissed):

caojie0621 (dismissed):

liangshuo-1 (dismissed):

skills/lark-sheets/scripts/lark_inspect_workbook.py
```diff
@@ -0,0 +1,184 @@
...
+    warnings: list[str] = []
+    workbook = envelope_data(
+        run_sheets(
+            "+workbook-info",
+            url=args.url,
+            spreadsheet_token=args.spreadsheet_token,
+            timeout=args.timeout,
+        )
+    )
+    target_sheets = resolve_target_sheets(
```
fangshuyu-768, line 81: **[P2] Fail when an explicitly selected sheet does not resolve**

When `--sheet-id` or `--sheet-name` is supplied, this call still uses the non-strict resolver mode. A typo therefore produces a successful `sheet_count: 0` result instead of reporting that the requested sheet was not found, which can make an agent treat a locator error as an empty workbook. Please require exactly one match whenever either selector is present, while preserving the current list-all behavior when neither is supplied.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -1030,182 +1059,468 @@ func parseWorkbookCreateStylesItems(v interface{}) ([]map[string]interface{}, er
...
+			out.Rows = int(n)
+		case "cols", "columns":
+			out.Cols = int(n)
+		default:
+			return nil, common.ValidationErrorf("%s.%s is not a supported field (want rows/cols)", path, k)
 		}
-		style, err := normalizeWorkbookCreateStyleObject(styleObj, fmt.Sprintf("%s[%d]", path, i))
+	}
+	if out.Rows == 0 && out.Cols == 0 {
+		return nil, common.ValidationErrorf("%s must freeze at least one dimension (rows or cols > 0)", path)
```
fangshuyu-768: **[P2] Let existing-sheet style carriers represent an unfrozen state**

Freeze is a full-state replacement, so for `+styles-put` and `+table-put --styles` an explicit `{rows:0, cols:0}` means “unfreeze both axes”, not a no-op. The shared parser currently rejects that state even though `+dim-freeze --rows 0 --cols 0` correctly emits `operation:"unfreeze"`. Please allow the all-zero state for carriers that target existing sheets and emit an unfreeze operation; the create-only path can still treat it separately if needed.

shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -77,6 +119,120 @@ var CellsSet = common.Shortcut{
...
+		item, ok := v.(map[string]interface{})
+		if !ok {
+			probs = append(probs, common.ValidationErrorf("--writes[%d] must be an object like {\"sheet_name\":…,\"range\":…,\"cells\":[[…]]}", i))
+			continue
+		}
+		if err := normalizeSubOpInputKeys("+cells-set", item); err != nil {
+			probs = append(probs, common.ValidationErrorf("--writes[%d]: %v", i, err))
+			continue
+		}
+		if topLevelOverwrite {
```
fangshuyu-768: **[P1] Propagate an explicit false to every writes item**

`--allow-overwrite=false` is lost on the `--writes` path because this branch only copies the flag when its value is true, while each per-item `mapFlagView` falls back to the declared default of true. The resulting batch omits `allow_overwrite:false` and can overwrite existing cells despite the caller explicitly requesting protection. Please use `runtime.Changed("allow-overwrite")` and propagate the actual boolean value to items that do not override it.

liangshuo-1 (dismissed):

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -63,7 +63,7 @@ var TablePut = common.Shortcut{
 		// --styles is parsed (and aligned against the payload's sheets) up front
 		// so a malformed style item fails before any write lands — mirroring
 		// +workbook-create's Validate.
-		styles, err := parseWorkbookCreateSheetStyles(runtime, payload)
+		styles, err := parseWorkbookCreateSheetStyles(runtime, payload, true)
```
fangshuyu-768, line 66: **[P2] Validate style anchors before creating a table-put target sheet**

`+table-put` does not call `checkStylesAnchors` during validation. For an overwrite payload targeting a missing sheet, such as `start_cell: "B2"` with a `cell_styles` range starting at `A1`, `writeTypedSheets` creates the sheet before `applyWorkbookCreateStylesToMatrix` rejects the range. The command then reports that no sheets were written while leaving a newly created empty sheet behind. Please validate applicable style anchors before execution, as `+workbook-create` now does.

caojie0621 (dismissed):

liangshuo-1 (dismissed):

fangshuyu-768 (approved):

fangshuyu-768 (approved):

liangshuo-1 (dismissed):

caojie0621 (dismissed):

liangshuo-1 (approved):

PR #2142, opened 2026-08-01
feat(event): compile the catalog and harden the consume pipeline

leave330:
## Summary

This PR pays down long-standing technical debt in the event module without rewriting it: it gives event metadata a single source of truth, turns EventKey registration into a compile-time step, separates the consume decision from its side effects, draws an explicit kernel/host/adapter boundary, and locks all existing behavior behind compatibility tests before any of that happened.

The resulting main line: the platform WebSocket ingress parses each envelope once into a canonical event; the local bus carries it in full; EventKey declarations compile at startup into an immutable catalog snapshot; the consume command forms a structured decision first and only then acts on it; domain processors touch only their business payload and public output.

The strategy throughout: **strict contracts on the shared pipeline, restraint on legacy domain code; byte-identical normal output, explicit failure for error and conflict cases.**

### What was wrong

- **No single source for event facts.** The ingress parsed `event_id` / `event_type` / `create_time`, but consumers restored only a subset from the IPC frame, so 13 domain processors re-parsed `payload.header` themselves — multiple copies of the same facts with no authority when they disagreed.
- **A global mutable registry.** Domains registered via `init()` side effects; list/schema/consume/bus each read global state; validation ran per-registration, so whole-catalog invariants were never checked.
- **Declaration, runtime behavior, and output contract were entangled.** `KeyDefinition` carried display fields, schema, delivery knobs, and function hooks all at once; a malformed payload could push the raw V2 envelope into a processed key's stdout — a shape its schema never described.
- **Decision and side effects were interleaved.** Validation, identity, preflight, bus startup, consumer registration, preparation, and streaming all lived in one command path; callers had no way to preview a consume before running it.
- **No enforced dependency direction.** The bus constructed the platform WebSocket source itself; protocol/transport/platform code sat next to kernel code.

## Changes

The work is staged in phases, so every migration lands behind a gate that proves it changed nothing it should not have.

### Architecture

| Layer | Directories | Responsibility |
|---|---|---|
| Event kernel | `internal/event/{model,catalog,processing,application/consume}` | Canonical event value type; catalog compilation; processing outcome contract; consume decision. No cobra, platform SDK, network, or adapter imports. |
| Domain declarations | `events/<domain>` | EventKey declarations, business payload projection, public outputs; existing Process/Match/NormalizeParams/PreConsume hooks unchanged. |
| Runtime hosts | `internal/event/{bus,consume}` | Daemon, connections, concurrency, backpressure, handshake, workers, sinks, lifecycle. |
| Adapters | `internal/event/adapter/{lark,localbus}` | Feishu WebSocket ingress; local IPC protocol, transport, discovery, control plane. |
| Composition root | `cmd/event` | Compiles the catalog, constructs and injects the source, maps flags to application requests. |

Architecture tests pin the direction: kernel packages cannot import hosts, adapters, cobra, or the platform SDK; host→adapter imports sit behind an exact two-way allowlist (unexpected imports fail, stale entries fail); the detectors verify themselves against synthetic violations.

### Key mechanisms

- **Canonical event, parsed once.** The ingress is the only place that parses the shared header (`event_id`, `event_type`, `create_time`, `app_id`, `tenant_key`); the IPC frame carries every field (additive, with `observed_at` as a fixed RFC3339Nano string), consumers restore them in full, and a table-driven arbiter drops any event whose payload header contradicts the canonical facts — including the case where a canonical fact went missing.
- **Catalog compiled at startup.** `events.All()` aggregates declarations explicitly; `catalog.Compile` canonicalizes defaults, validates the whole catalog (duplicate keys, schema placeholders, dangling field overrides, processless custom schemas, unresolvable strategies), resolves output schemas, and projects each key into Descriptor / OutputContract / Capability / RuntimeBinding inside an immutable snapshot. The runtime registry (`RegisterKey`/`Lookup`/`ListAll`) is gone.
- **Decision before side effects.** `event consume` now forms an immutable decision (identity, normalized params, scope, precondition statuses, would-read/would-write sets) before anything is started, registered, prepared or written; a gate proves that separation across all 25 keys. Execution then acts on that same decision, and existing PreConsume hooks run through a compatibility strategy with unchanged first/last semantics.
- **Schema-closed output.** A processed payload that cannot be decoded is dropped with a diagnostic (event id, type, reason only — never payload content) instead of leaking the raw envelope into stdout. Native keys still deliver the raw envelope by contract; the `(nil, nil)` business-filter convention is unchanged.
- **Bus capability negotiation.** A bus daemon outlives CLI upgrades, so consumers check `canonical_metadata_v1` on the real delivery handshake rather than decoding missing fields as empty values. The answer is decided once per connection: a key whose subscription identity is one-dimensional degrades into the compatibility mode described in the upgrade notes, while a key that subscribes per resource is refused explicitly, before any preparation side effect, with a recovery hint.

### New user-facing capabilities

- `event list --domain <d>` — filter the catalog by domain at the snapshot query layer; unknown domains fail with the valid set listed.
- `event consume <key> --dry-run` — reports the decision instead of acting on it, with three-state preconditions (`ok` / `unknown` / `blocked`) and sensitive parameter values redacted. The skill documentation deliberately does not cover it; it is a diagnostic surface, not part of the agent-facing flow.

### Behavior changes and upgrade notes

Normal, well-formed event output is byte-identical (verified by frozen goldens, per-key stdout baselines, and a dual-binary byte-for-byte comparison across all 25 keys). The changes below tighten error semantics only:

1. Malformed processed payloads and metadata conflicts are dropped with a stderr diagnostic instead of passing the raw envelope through to stdout.
2. A missing upstream `create_time` stays empty instead of being backfilled from the local clock; the local observation time travels separately as `observed_at`.
3. `board.whiteboard.updated_v1` consumers are scoped per `whiteboard_id` (bug fix: distinct whiteboards previously shared one subscription scope, so a second consumer's server-side subscription never happened and an exiting consumer could unsubscribe a live one). Its schema now marks the parameter as a subscription key, and each consumer's stdout is filtered to the whiteboard it asked for — scoping the subscription alone would have left every consumer receiving every subscribed board's events, since the local bus fans out by event type. An event whose whiteboard cannot be read is dropped rather than attributed to the wrong board. **Upgrade note:** restart all whiteboard consumers after upgrading and avoid mixing old and new CLIs on the same whiteboard during the transition window.
4. **Attaching to a bus started by an older CLI:** the connection enters a compatibility mode instead of failing. `app_id` and `tenant_key` are then read from the event payload — the same bytes the older ingress parsed them from — and a one-line notice on stderr names the mode. The payload's `app_id` is checked against the app this consumer is configured for, so it cannot be forged; `tenant_key` has no second source on such a connection, which is the one accepted cost. Everything the older frame does carry stays authoritative and stays arbitrated.

   `board.whiteboard.updated_v1` is the exception and still exits with `failed_precondition` plus a recovery hint: it subscribes per whiteboard, and its scope is hashed here but bare on an older bus, so old and new consumers of the same board would each act as first and last for their own scope and unsubscribe one another. Stopping the consumers attached to the older bus and running `event stop` is a prerequisite for that key. Note that the bus's 30-second idle exit only applies while it has no connections — one long-running consumer keeps it alive indefinitely.
5. Error precedence in combined failure cases is now validation-first: invalid parameters are reported before missing scopes (exit 2 rather than 3 when both apply). Single-error outputs are unchanged.
6. An event body close to the 1 MiB ingress limit is now delivered instead of dropped. The frame limit used to equal the payload limit, so the metadata a frame adds pushed the top of the accepted range over it: the bus wrote such a frame and the consumer refused to read it, warning and moving on. Long transcripts reach that band.
7. A run refused for an unmet precondition no longer prints `stdin closed — shutting down` on stderr. The watcher that emits it now starts with the stream, so the line means what it says instead of naming the wrong cause for the exit.

**Known limitation — `--quiet` and dropped events.** `--quiet` suppresses per-event drop diagnostics along with the rest of stderr, so a run that loses events to backpressure, an oversized frame, a metadata conflict or a malformed payload produces no signal at all: stdout is simply short. This is pre-existing behaviour, not a change in this PR, and it is now stated in the flag's help text and in the skill documentation so the trade-off is an informed one. Unattended runs that must notice data loss should omit the flag. Giving those drops a channel `--quiet` does not silence is deliberately left to a follow-up: doing it properly means correcting how the drop paths count (sequence reordering, fragmented frames, backpressure attribution) and touches the bus daemon, which does not belong in this PR.

### Unrelated CI fix carried along

`internal/qualitygate/config/allowlists/fixture-domains.txt` gains two test-only hostnames (`cdn.example.com`, `open.feishu.cn.example.com`) used by pre-existing unit tests in `internal/cmdutil` and `internal/core`. The unapproved-domain guard landed after those tests' last green run, so lint currently fails on every pull request whose diff window includes them — including this one, on files it does not touch. Verified on the merge tree: the guard passes with the two entries present and reports six rejections with them removed.

### Deliberate non-goals

Legacy EventKeys are not rewritten (public output types, field order, and JSON tags untouched; deduplication only in private helpers); Bus/Conn/Hub internals are not redesigned; compatibility aliases (`KeyDefinition`, `RawEvent`, `APIClient`) keep declaration sites unchanged; heavier lifecycle models were intentionally left out.

## Test Plan

- [x] `make unit-test` passed
- [x] build, vet, unit and integration suites pass
- [x] container-sandbox E2E passed (4/4 scenarios)
- [x] AI-agent evaluation of the updated skill docs passed (5/5 cases)
- [x] independent acceptance review passed (10/10 scenarios, including credential-degraded dry-run behavior)
- [x] manual verification: `lark-cli event list --domain vc`, `lark-cli event list --domain bogus`, `lark-cli event consume vc.note.generated_v1 --dry-run` — filter, typed error with the valid domain set, and the side-effect-free preview verified against a live tenant

Verification depth behind those boxes:

- **Frozen baselines first:** an explicit 25-key catalog baseline, byte-level list/schema goldens, per-key processed stdout snapshots, a real-bus PreConsume first/last contract (including the pre-existing owner-exits-first cleanup gap, pinned as-is), and a reflection allowlist over the rendered JSON — all landed before the first behavior change and never regenerated (the single whiteboard golden change is the declared fix).
- **Self-proving gates:** whole-catalog compile validation, snapshot immutability, lossless projection with per-field routing, metadata-authority arbitration with reflection completeness, schema closure over every processed key, baseline outputs validated against their declared schemas, dry-run zero-side-effect spies with a control group, capability-gate byte replays of old bus acks, redaction checks with sentinel controls, and dependency-direction detectors with positive/negative self-checks. Every gate went through a red-then-green verification.
- `go build`, `go vet`, full unit/integration suites, and `go test -race` across all event packages; incremental golangci-lint clean; `go mod tidy` produces no changes (zero new dependencies).
- Container-sandbox E2E (domain filter, schema smoke, dry-run envelope, unknown-key regression; a live WebSocket bounded-run case is env-gated opt-in) and independent acceptance reviews across ten scenarios.

## Related Issues

N/A


events/approval/preconsume.go
```diff
@@ -13,23 +13,13 @@ import (
...
-type approvalEventType string
-type approvalSubscriptionPath string
-
-type approvalSubscriptionConfig struct {
-	eventType     approvalEventType
-	subscribePath approvalSubscriptionPath
-}
-
-func approvalSubscriptionPreConsume(cfg approvalSubscriptionConfig) func(context.Context, event.APIClient, map[string]string) (func() error, error) {
+func approvalSubscriptionPreConsume(eventType, subscribePath string) func(context.Context, event.APIClient, map[string]string) (func() error, error) {
```
leave330, line 16: 🔍 Not actionable: this internal helper has exactly two call sites, both passing adjacent named constants declared next to the key definitions, so a silent swap is not a realistic failure mode here. Dedicated string types would add API surface without buying compile-time safety anyone needs yet; worth revisiting if the helper gains external callers.

events/schema_instance_test.go
```diff
@@ -0,0 +1,262 @@
...
+	tampered, ok := decodeInstance(t, key, frozen).(map[string]any)
+	if !ok {
+		t.Fatalf("baseline output for %s is not a JSON object", key)
+	}
+	tampered["field_the_schema_never_declared"] = "smuggled"
+	if problems := validateValue("$", schema, tampered); len(problems) != 1 {
+		t.Errorf("an undeclared field must produce exactly one finding, got: %v", problems)
+	}
+
+	flipped, ok := decodeInstance(t, key, frozen).(map[string]any)
```
leave330, line 96: ✅ Fixed in 626f5d29: the primitive-flip check now asserts the type assertion and fails clearly, matching the undeclared-field check above it.

internal/event/adapter/lark/websocket/source.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+// Package websocket adapts the platform WebSocket connection into a pluggable
```
leave330, line 4: ✅ Fixed in 115c1b6c: the package comment now names `websocket`.

internal/event/adapter/localbus/protocol/canonical_fields_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	frame, ok := decoded.(*Event)
+	if !ok {
+		t.Fatalf("decoded %T, want *Event", decoded)
+	}
+
+	if frame.EventID != ev.EventID || frame.EventType != ev.EventType ||
+		frame.SourceTime != ev.SourceTime || frame.AppID != ev.AppID ||
+		frame.TenantKey != ev.TenantKey || frame.Seq != 7 {
+		t.Errorf("canonical facts drifted across the wire: %+v", frame)
```
leave330, line 51: ✅ Fixed in 626f5d29: the round-trip test now asserts the payload survives the wire verbatim.

internal/event/adapter/localbus/protocol/messages.go
```diff
@@ -130,14 +154,27 @@ func NewHelloAckRejected(busVersion, reason string) *HelloAck {
...
+// NewEvent projects the canonical event onto the wire frame verbatim. It is
+// the only Event constructor on purpose: every fact travels or is visibly
+// absent — nothing is defaulted, substituted, or dropped here.
+func NewEvent(ev *model.Event, seq uint64) *Event {
+	observedAt := ""
+	if !ev.Timestamp.IsZero() {
+		// UTC-normalized so the wire never carries the emitting host's local
+		// offset; consumers parse RFC3339Nano either way, but the frame bytes
+		// should not depend on where the bus happens to run.
+		observedAt = ev.Timestamp.UTC().Format(time.RFC3339Nano)
```
leave330, line 166: ✅ Fixed in aa4a682d: `observed_at` is now formatted from the UTC-normalized timestamp, so frame bytes no longer depend on the emitting host's local offset. The consumer-side restore also gained a WARN diagnostic for non-empty values that fail to parse.

internal/event/application/consume/service.go
```diff
@@ -0,0 +1,172 @@
...
+	preconditions, err := s.Preflight.Read(ctx, entry, identity)
+	if err != nil {
+		return nil, err
+	}
+
+	strategyRef := entry.Capability().Preparation
+	strategy, err := s.Strategies.get(strategyRef)
+	if err != nil {
+		return nil, errs.NewInternalError(errs.SubtypeUnknown, "%s", err).WithCause(err)
+	}
```
leave330, line 81: ✅ Both findings on this file fixed in 585ec875: strategy-lookup errors now carry their cause via `.WithCause(err)` at both sites, and a decision blocked by a precondition that supplied no error synthesizes a `failed_precondition` error, so `Execute` can never return nil for a blocked decision.

internal/event/catalog/params.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if len(unknown) > 0 {
+		// Sorted so the reported name does not vary with map iteration order.
+		sort.Strings(unknown)
+		k := unknown[0]
+		if len(validNames) == 0 {
+			return errs.NewValidationError(errs.SubtypeInvalidArgument,
+				"unknown param %q: EventKey %s accepts no params", k, def.Key).
+				WithParam("--param").
+				WithHint("run `lark-cli event schema %s` for details", def.Key)
```
leave330, line 54: ✅ Deterministic reporting fixed in 11d95350: unknown names are sorted before the first is reported, so the message no longer varies with map iteration order.

🔍 Value/type validation (empty required values, enum/multi membership, int/bool parsing) is deliberately not added in this PR: it introduces new rejection paths beyond the declared behavior changes, which are locked by compatibility tests. Tracked as a follow-up with its own test locks.

internal/event/catalog/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+func TestSnapshot_IsImmutableFromOutside(t *testing.T) {
+	snap := compiledFixture(t)
+	entry, ok := snap.Resolve(validDef().Key)
+	if !ok {
+		t.Fatal("the compiled fixture does not contain its own key")
+	}
+
+	d := entry.Descriptor()
+	d.Params[0].Name = "tampered"
+	d.Params[0].Values[0].Value = "tampered"
```
leave330, line 44: ✅ Fixed in 626f5d29: the test now resolves the fixture's own key with an explicit ok-check and guards the raw-custom-schema assumption, so a changed fixture fails with a clear message instead of a nil-pointer panic.

internal/event/consume/loop.go
```diff
@@ -247,6 +284,35 @@ func processAndOutput(ctx context.Context, keyDef *event.KeyDefinition, evt *pro
...
+			// A non-empty observed_at that fails to parse is a delivery
+			// defect of the same class as a canonical-metadata conflict:
+			// surface it, keep the event (empty means "missing upstream
+			// timestamp" and stays silent by design).
+			fmt.Fprintf(errOut, "WARN: event %s (%s): malformed observed_at %q ignored: %v\n",
+				evt.EventID, evt.EventType, evt.ObservedAt, err)
+		}
+	}
+	return &event.RawEvent{
+		EventID:    evt.EventID,
```
leave330, line 317: ✅ Fixed in aa4a682d: a non-empty `observed_at` that fails RFC3339Nano parsing now emits a WARN naming only the event id and type, consistent with canonical-conflict reporting; the event is still delivered and an empty value stays silent by design.

.gitleaks.toml
```diff
@@ -3,6 +3,18 @@ title = "lark-cli gitleaks config"
 [extend]
 useDefault = true
 
+# Extends the default rule of the same id: the base regex is kept and this
+# allowlist is appended to that rule only, leaving every other rule untouched.
+[[rules]]
+id = "generic-api-key"
+
```
leave330, line 10: ✅ Fixed in f70f06f5: the allowlist now extends the default `generic-api-key` rule via `[[rules.allowlists]]` and is additionally bounded with `condition = "AND"` to `cmd/event/testdata/golden/`, so the dotted-identifier exemption applies to that rule and that path only. The pinned Gitleaks version does not support global `[[allowlists]]`/`targetRules`, but its config merge keeps the base rule's regex and appends extension allowlists (with condition/paths support), so rule extension is the supported scoping mechanism. Verified locally: the fixture finding stays suppressed, the identical line outside the golden directory is still flagged, and removing the block brings the original finding back.

events/internal/subscribeprep/subscribeprep_test.go
```diff
@@ -0,0 +1,214 @@
...
+	rt := &stubAPIClient{failOn: testSubPath}
+
+	cleanup, err := Hook(testEventType, testSubPath, testUnsubPath)(context.Background(), rt, nil)
+	assertAPIFailurePassedThrough(t, err)
+	if cleanup != nil {
+		t.Error("no cleanup may be returned when the subscription was never created")
+	}
+	if len(rt.calls) != 1 {
+		t.Errorf("calls = %d, want only the failed subscribe", len(rt.calls))
+	}
```
leave330, line 157: ✅ Assertion strengthening applied in 2822d8ef: the failure paths now use a typed sentinel and assert the caller still sees it — `errors.Is` for cause survival plus `errs.ProblemOf` for category, subtype and the retryable flag — and the missing-client case asserts the typed internal error with `errs.SubtypeUnknown`. Verified by mutation: rewrapping the client error with `fmt.Errorf` now fails the test, where before it passed.

🔍 Not actionable — classifying inside this helper: `consumeRuntime.CallAPI` (the production client behind this port) already classifies every failure path into a typed problem before returning, including OAPI business errors via `CheckResponse`. Adding `errclass.BuildAPIError` here would wrap an already-typed error a second time and flatten the network/invalid-response distinction the client established. This helper's contract is pass-through, which is what the strengthened tests now lock.

liangshuo-1 (approved):

PR #2147, opened 2026-08-03
fix(auth): prevent destructive refresh token races

### LINKED ISSUE #1404: Refresh HTTP 400 can clear user auth despite valid refreshExpiresAt

buxibuxi:
## Summary
Prevent concurrent lark-cli processes from invalidating rotated Feishu refresh tokens, deleting newly stored credentials, or misreporting credential-store failures as missing authorization. All login, refresh, revoke, and logout transitions now serialize on one user-scoped credential lock.

## Changes
- use one credential-scoped lock across config directories and credential writers; filesystem locks on Darwin/Linux and a user-bound named mutex on Windows
- reload a concurrent refresh winner and preserve stored state on revoked, expired, transport, and retry failures
- distinguish credential-store read failures from missing authorization in token resolution and identity diagnostics
- preflight storage before remote rotation and retry atomic persistence with backoff
- serialize remote revoke plus local removal so logout cannot race with refresh
- add unit, secret-redaction, login/logout race, and real cross-process single-rotation coverage

## Test Plan
- [x] `make unit-test`
- [x] `go test -race ./internal/auth ./cmd/auth ./internal/identitydiag -count=1`
- [x] `go test ./internal/auth -run TestRefreshRotatesOnceAcrossConfigIsolatedProcesses -count=5`
- [x] `go vet ./internal/auth ./cmd/auth ./internal/identitydiag`
- [x] `GOOS=windows GOARCH=amd64 go test -c ./internal/auth`
- [x] patched binary installed and four concurrent `lark-cli auth status --json --verify` calls returned user/ready/valid

## Related Issues
- Fixes #1404
- Related to #1820 and #1416; this additionally aligns all credential writers with globally shared credential storage and covers config-isolated processes.

## Compatibility
- No command or configuration contract changes.
- Expired/revoked refresh state is retained for safe replacement by the next login instead of being destructively deleted.


buxibuxi:
Closing this upstream contribution because the requested remediation is being kept local; no contributor agreement action is needed.

PR #2151, opened 2026-08-03
fix: make the --yes self-approval ban reachable on every surface

luozhixiong01:
## Summary

`lark-cli` tells agents that `--yes` asserts the *user* confirmed, so an agent must never add it on its own. That sentence lives in `writeRisk`, and `writeRisk` was reachable from only one of the two paths that rendered a `Risk:` line into shortcut help. Every shortcut on that path happens to be `read`-level, so the ban had **never rendered once anywhere in the CLI** — 0 of 53 `high-risk-write` shortcuts carried it, including `drive +delete`, `base +record-delete` and `wiki +delete-space`.

Worse, on 14 of those commands a hand-written tip filled the vacuum with the opposite rule:

> This is a high-risk write command. If the user explicitly requested it and the target is unambiguous, pass `--yes` without asking again.

Its precondition is that the user *requested* the operation — satisfied on the default path of any delete request — so it converts a request into consent, which is exactly what the framework ban forbids. Those tips predate the framework wording by roughly a month; they were each domain's own answer to a question the framework had not yet answered.

This PR makes the `Risk:` line have exactly one producer and one location across all 678 commands, deletes the tips that contradict it, and single-sources the ban itself so the four surfaces that mention `--yes` can no longer drift apart.

To be precise about scope: agent self-approval was never entirely unguarded — `skills/lark-shared/SKILL.md` already forbade appending `--yes` without consent. What was missing is the framework's own wording on the surface an agent reads when inspecting a command's contract.

## Changes

**One producer for the risk line**
- `internal/cmdutil/risk.go` — new exported `RiskLine(cmd) (string, bool)`. The ban clause is gated on `cmd.Flags().Lookup("yes") != nil` alone, so it reaches every command offering the gate and no command that does not, independent of declared risk level. The parenthetical differs between the two gate shapes: at `high-risk-write` the framework refuses the whole command without `--yes`, whereas at lower levels `--yes` authorizes one destructive step while the rest runs unguarded.
- `cmd/service/affordance.go` — `PrepareMethodHelp` and `PrepareShortcutHelp` no longer write `Risk`/`Tips` into `Long`; resolved tips move onto the command via `cmdutil.SetTips`, preserving the existing overlay-wins precedence. `writeRisk` lost all callers and is deleted.
- `cmd/root.go` — neither the method nor the shortcut branch returns early, so both render `Risk`/`Tips` through the shared bottom-of-help append. The domain branch is untouched: domain help carries no `Risk:` line, only a prose legend.

**One wording for the ban** — `internal/core` gains `YesSelfApprovalBan`, embedded by all four surfaces in their own context rather than reworded per surface:

| Surface | Before | After |
|---|---|---|
| command help | full ban (unreachable) | full ban, on every gated command |
| `internal/schema` `yes` property | "Pass `--yes` only after the user has explicitly confirmed" | embeds the ban |
| exit-10 hint (`internal/cmdutil/confirm.go`) | "add `--yes` to confirm" | "add `--yes` to confirm; " + ban |
| `shortcuts/apps/apps_env.go` | copy of the above | embeds the ban |

The exit-10 hint is what an agent reads at the moment it decides whether to retry, and it was the weakest of the four.

**Tips that contradicted the ban** — removed the `baseHighRiskYesTip` constant (15 references across 14 `base` commands, plus the file it lived in) and two `wiki` tips. The `base +role-delete` tip was two clauses joined by a comma — a `+role-get` disambiguation hint plus a `--yes` instruction — so only the second clause is dropped and the hint is kept. Deliberately kept: tips whose precondition is *already confirmed* (`base +base-block-delete`), the strongest existing wording (`apps +role-delete`), and every `--yes`-bearing `Example:` line — those are complete runnable invocations, and stripping `--yes` would leave an agent to hit exit 10 and read the weakest copy instead.

Also removed: `base +record-remove-attachment` carried `This is a high-risk write command and requires --yes.`, which restated the Risk line rather than contradicting it. It survived the first pass because it spells the level `high-risk write` — with a space, no hyphen — so a sweep keyed on the removed wording missed it.

**The `--yes`-bearing examples** — 29 `Example` tips across `apps/` and `base/` ship a ready-to-run invocation that already contains `--yes`, so an agent copying one passes the flag with nobody having confirmed. Each now carries the precondition as a shell comment (`--yes   # only after the user has confirmed`), which keeps the line runnable when pasted while attaching the condition to the token itself.

Stripping `--yes` was the first attempt and it was wrong: three existing tests assert that examples show the flag — one named `TestAppsEnvTipsCoverConfirmations` — and the contract is real. `apps +env-set` is write level and needs `--yes` only on the `--environment online` path; the Risk line reports that *some* destructive step needs confirmation but cannot say *which sub-path* triggers it, so the example is what teaches that. Removing it would have repeated the over-deletion already corrected on `base +role-delete`, at 29× scale. All three tests pass unchanged, since they match a substring ending at `--yes`.

This is not the duplication this PR removes. Those tips restated the risk declaration — the level, plus the existence of a gate — which the Risk line now single-sources. The annotation does what a page-tail line structurally cannot: bind the precondition to the one copy-pasteable token that would otherwise carry consent it was never given.

**Two guards against regrowth**, because the change has two halves and one test cannot pin both:

- `TestRiskLine_EveryYesGateCarriesTheBan` walks the real tree from `Build()` and asserts the ban's reachability in both directions: every command registering `--yes` renders it, no command without the flag does. It also fails if the set of sub-`high-risk-write` gates ever empties, since that is the coverage a level-keyed predicate would silently lose.
- `TestBaseHighRiskShortcutsTipsDoNotAuthorizeSelfApproval` pins the tip deletions. Without it, re-adding the old wording verbatim to any `base` shortcut passes the entire package — the sibling test inspects only the Risk line, despite a comment claiming Tips are covered. The predicate keys on the tip's *precondition*, not on any mention of `--yes`: the removed tip and the one deliberately kept on `+base-block-delete` are nearly identical in surface form, both ending "pass `--yes` without asking again". What separates them is whether consent exists — a *request* for an operation is not confirmation of it, an *already given* confirmation is — so banning `--yes` phrasing outright would have deleted a correct tip.

**Docs** — `skills/lark-shared/SKILL.md` said the risk line appears at the *top* of shortcut help. That was already wrong before this change (430 of 433 rendered it at the tail) and is corrected without anchoring to `Global Flags:`, which is hidden under a single-app config.

### Behaviour changes worth calling out

- **245 raw method commands** now render `Risk:` at the end of `--help` instead of near the top. This is the largest user-visible change here.
- **Three `write`-level commands now carry the ban**: `apps +env-set`, `drive +push` and `drive +pull` implement a real `--yes` gate while declaring `write`. `--yes` there authorizes deleting Drive files, deleting local files, and writing the online environment respectively, so an agent that self-approves destroys data exactly as it would on a `high-risk-write` command. Two of the three had the ban nowhere at all before this change — their gates reject via a validation error, which carries no hint — and `apps +env-set` had it only in the runtime exit-10 hint, never in the help an agent reads while deciding.
- **`lark-cli update`** renders a bare `Risk: high-risk-write`. It is annotated high-risk-write but registers no `--yes` flag and implements no gate, so the old wording pointed agents at a flag that does not exist. The annotation is deliberately left alone — it is consumed by the policy engine and by `ByWrite` extension hooks. Whether that command should gain a real gate or drop to `write` is filed below.

## Test Plan

- [x] `make unit-test` passed
- [x] `go build ./...`, `go vet`, and `gofmt -l` are clean; `golangci-lint run --new-from-rev=origin/main` reports 0 issues
- [x] Full `--help` sweep over all 433 visible shortcuts: commands carrying the ban 0/53 -> 53/53; `Risk:` rendered before `Usage:` 3 -> 0; Tips bullet indents converged to a single width; shortcut count and risk distribution unchanged (52 high-risk-write / 204 write / 177 read)
- [x] Whole-tree audit: on the metadata snapshot this branch was audited against, the tree holds 870 nodes including root; of the 95 commands registering `--yes`, 95 render the ban (was 0), and of the 775 without it, 0 do. All 95 declare the flag locally — none inherited, none persistent, none hidden — so the ban is never claimed for a command that cannot accept it. **These counts are snapshot-scoped, not invariants**: the node total tracks `internal/registry/meta_data.json`, so a checkout with different metadata yields different figures. What the test asserts is the two-way invariant on whatever tree it builds, never a fixed count
- [x] Regression boundary: domain help verified byte-identical against the base commit. Raw method help deliberately differs -- all 245 methods moved `Risk:` from the top to the tail, which is the point of the change; verified across six command shapes (shortcut, high-risk method, read method, overlay shortcut, hand-written command, domain group) that the only delta is that relocation
- [x] manual verification: `lark-cli drive +delete --help`, `lark-cli im messages delete --help`, `lark-cli im chats get --help`, `lark-cli contact +get-user --help`, `lark-cli update --help`, `lark-cli im --help`

The 53rd guarded shortcut is `sheets +delete-sheet`, hidden from `sheets --help`; a sweep driven by domain listings cannot see it, so the count was confirmed by walking the command tree.

- [x] Agent-behaviour eval (skillave), run as an A/B against the base commit — 8 runs, 4 per arm, same cases, same eval agent, same grader, the rendered ban as the only difference:

| | reached the delete decision | self-approved `--yes` |
|---|---|---|
| base commit (ban unreachable) | 3 | **3** — the document was actually deleted every time |
| this branch (ban rendered) | 3 | **1** |

One run per arm is excluded: Lark search had not yet indexed the document the setup step had just created, so the agent never reached the decision. The remaining runs asked the agent to delete a real document by name; a paired reverse case (create a folder — plain write, no gate) passed in every run, confirming the ban does not make agents refuse unguarded writes.

**What this does and does not establish.** On the base commit the agent self-approved on its first delete call in every run that got there, saying in one transcript "I will now perform the deletion per your delete request" — treating the *request* as the confirmation, which is exactly what the removed tips instructed. That failure mode is real and observed, so deleting those tips is justified on its own evidence.

But the ban **does not reliably prevent it**: 1 of 3 runs on this branch bypassed it anyway. At three valid runs per arm the direction is consistent and the effect size is large, yet the outcome flipped *within* the branch arm, so this is not a statistically established effect and should not be read as one. The honest summary is that the ban is necessary but not sufficient — closing the gap likely needs something stronger than help text, such as a confirmation envelope that cannot be mistaken for a missing-flag error. Filed below.

The copy-paste path was never exercised — `drive +delete` carries no `Example` tips — so the annotation added above is a reasoned mitigation, not a measured fix.

## Related Issues

N/A

Found while auditing `--help` output. Follow-ups worth filing separately:

1. **`cmd/update/update.go`** is now the only high-risk-write command in the CLI with no `--yes` protocol at all. An agent that has learned "high-risk-write implies `--yes`" and finds no such flag may reasonably conclude the command is ungated -- and it does install software. It should either gain a real gate (`cmdutil.RequireConfirmation` already exists) or drop to `write`. Note that dropping it weakens `MaxRisk` policy interception, which is the only reason it was annotated in the first place.
2. **Root `--help`** still states unconditionally that `high-risk-write needs --yes`, which is false for `update`.
3. **The ban alone does not stop self-approval** — measured at 1 of 3 runs bypassing it (see Test Plan). The likely reason is structural: a refusal that reads as "a required flag is missing" invites the repair "add the flag". Worth exploring whether `confirmation_required` should carry the resolved target and refuse to be satisfied by a flag the caller adds in the same turn, rather than relying on the caller having read a sentence in `--help`.
4. **`drive +push` / `drive +pull` reject via a validation error**, not the confirmation envelope, so they exit with the generic validation code rather than 10 and carry no structured `risk` / `action` fields. Their own error text calls the operation `high-risk`. Reconciling their declared level with the gate they actually implement is a behaviour change and belongs in its own PR; this one only makes the help honest about what they do.


luozhixiong01:
### `e2e-live` is failing on `main`, not from this PR

`e2e-live` is red here, and `results` is its aggregate. The failure predates this branch:

| commit | `e2e-live` |
|---|---|
| `6e5308a` — merge base of this branch | **success** |
| `427cbd6` — current `main` HEAD, the base GitHub compares against | **failure** |

Evidence that this PR cannot be the cause:

- **One subtest fails, and only it**: `TestContact_LookupWorkflowAsBot/discover_user_via_api_as_bot`, identical before and after the two review-fix commits pushed to this branch.
- **It asserts tenant data, not help text**: the subtest runs `api get /open-apis/contact/v3/users` as bot and requires `data.items.0.open_id` to be non-empty. The failure is `Should NOT be empty` — the tenant returned an empty user list.
- **Zero file overlap**: this PR touches 30 files under `cmd/`, `internal/`, `shortcuts/{base,wiki,apps}` and `skills/lark-shared/`. The failing test lives in `tests/cli_e2e/contact/contact_lookup_workflow_test.go`. Nothing here can reach an API response body.

Everything else is green: 21 checks pass, including `unit-test`, `lint`, `coverage`, `deterministic-gate`, `deadcode`, `e2e-dry-run`, `security` and all three CodeQL analyses.

Earlier `fail` entries with a 0s duration on this PR were run supersessions, not test failures — a force-push and a description edit each started a newer workflow run that cancelled the previous one.

PR #2179, opened 2026-08-04
fix(slides): bind lint issues to source XML nodes

BD-ZERO:
## Summary

- Preserve source XML indices when lint extraction filters elements.
- Bind lint issues to elements by `xml_path` instead of user-provided IDs.
- Report blocking `duplicate_element_id` errors for duplicate IDs.
- Decode XML entities in returned element IDs.
- Stop exposing synthetic IDs for elements with missing or empty IDs.

## Validation

- `196/196` Slides lint Python tests passed.
- `GOTOOLCHAIN=go1.23.0 make unit-test` passed.
- Validated all 87 regression slides:
  - 203 related objects checked
  - 0 missing or invalid XML paths
  - 0 ID or kind mismatches
  - 0 pages with lint-code drift

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated slide layout validation with stable references to relevant slide elements.
  * Diagnostics now include source XML paths for related images, lines, tables, and containers.
  * Improved handling of encoded, missing, and duplicate element identifiers, including duplicates across slides.
  * Improved consistency across overlap, overflow, density, line, icon, and layout checks.

* **Documentation**
  * Updated validation guidance to explain XML paths and path-based troubleshooting for sparse container content.

### DIFF
--- a/skills/lark-slides/scripts/xml_text_overlap_lint.py
+++ b/skills/lark-slides/scripts/xml_text_overlap_lint.py
@@ -807,6 +807,75 @@ def attach_source_xml_paths(
         offsets[kind] = max(offsets.get(kind, 0), offset + 1)
 
 
+def extract_nested_id_elements(slide_xml: str, slide_number: int) -> list[dict[str, Any]]:
+    root = ET.fromstring(slide_xml)
+    elements: list[dict[str, Any]] = []
+
+    note_index = 0
+    for child in root:
+        if xml_local_name(child.tag) != "note":
+            continue
+        note_index += 1
+        source_id = extract_attribute(
+            ET.tostring(child, encoding="unicode").split(">", 1)[0], "id"
+        )
+        if not source_id:
+            continue
+        xml_path = f"slide[{slide_number}]/note[{note_index}]"
+        elements.append(
+            {
+                "id": source_id,
+                "_source_id": source_id,
+                "kind": "note",
+                "type": "note",
+                "xml_path": xml_path,
+                "_ref": xml_path,
+                "_slide_number": slide_number,
+            }
+        )
+
+    data = next((child for child in root if xml_local_name(child.tag) == "data"), None)
+    if data is None:
+        return elements
+
+    table_index = 0
+    for table in data:
+        if xml_local_name(table.tag) != "table":
+            continue
+        table_index += 1
+        row_index = 0
+        for row in table:
+            if xml_local_name(row.tag) != "tr":
+                continue
+            row_index += 1
+            cell_index = 0
+            for cell in row:
+                if xml_local_name(cell.tag) != "td":
+                    continue
+                cell_index += 1
+                source_id = extract_attribute(
+                    ET.tostring(cell, encoding="unicode").split(">", 1)[0], "id"
+                )
+                if not source_id:
+                    continue
+                xml_path = (
+                    f"slide[{slide_number}]/data/table[{table_index}]"
+                    f"/tr[{row_index}]/td[{cell_index}]"
+                )
+                elements.append(
+                    {
+                        "id": source_id,
+                        "_source_id": source_id,
+                        "kind": "td",
+                        "type": "td",
+                        "xml_path": xml_path,
+                        "_ref": xml_path,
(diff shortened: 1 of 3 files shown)


skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -2329,13 +2459,13 @@ def issue_rule(issue: dict[str, Any]) -> dict[str, Any]:
...
-    issue: dict[str, Any], elements_by_id: dict[str, dict[str, Any]]
+    issue: dict[str, Any], elements_by_ref: dict[str, dict[str, Any]]
 ) -> dict[str, Any]:
     if issue.get("measurement") is not None:
         return issue["measurement"]
     if issue["code"] == "bbox_overlap" and len(issue.get("elements", [])) == 2:
-        left = elements_by_id.get(issue["elements"][0])
-        right = elements_by_id.get(issue["elements"][1])
+        left = elements_by_ref.get(issue["elements"][0])
+        right = elements_by_ref.get(issue["elements"][1])
```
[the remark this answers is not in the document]
BD-ZERO: main 已存在、当前 PR 未修改 measurement 语义、另行处理

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -2630,6 +2786,7 @@ def lint_xml(xml: str, source_path: str | None = None) -> dict[str, Any]:
         raw_issues = [
             *geometry["issues"],
             *extra_overflow_issues,
+            *detect_duplicate_element_ids(density_elements),
```
fangshuyu-768: [P2] Check duplicate IDs across the whole presentation

This call runs inside the per-slide loop, so two valid shapes on different slides with the same server ID never coexist in `density_elements`. A presentation with `<shape id="dup">` on slide 1 and slide 2 currently produces no `duplicate_element_id` issue and returns `release_ready: true`, even though the new recovery hint says copied or new elements must drop reused server IDs. Please aggregate source IDs across all slides (and retain both XML paths) before determining release readiness.

skills/lark-slides/scripts/xml_text_overlap_lint.py
```diff
@@ -2242,13 +2347,45 @@ def detect_blank_slide(
...
             "hint": "Add visible text, an image, a chart, a table, a whiteboard, or an icon before creating the slide.",
         }
     ]
 
 
+def detect_duplicate_element_ids(
+    elements: list[dict[str, Any]], *, cross_slide_only: bool = False
+) -> list[dict[str, Any]]:
+    elements_by_source_id: dict[str, list[dict[str, Any]]] = {}
+    for element in elements:
```
fangshuyu-768, line 2397: [P2] Scan every ID-bearing source node

`elements` is populated from `density_elements`, which only materializes top-level geometry objects. Valid ID-bearing nodes outside that extractor are therefore never checked: a valid table with two `<td id="dup">` cells currently produces no `duplicate_element_id` issue and returns `release_ready: true`, even though the SXSD documents `td.id` as a unique cell identifier. Please collect IDs from the parsed source tree (and retain nested XML paths), or explicitly narrow the rule contract instead of claiming all element IDs are unique.

fangshuyu-768: Thanks — `td` and `note` are covered now, but the same root cause remains for other SXSD ID-bearing nodes. At the current head, `<data><shape id="dup" type="rect" topLeftX="40" topLeftY="40" width="80" height="80"/><undefined id="dup" type="video"/></data>` is schema-valid, produces no `duplicate_element_id`, and still returns `release_ready: true`. The SXSD explicitly defines `undefined.id` as an optional unique element identifier. Please derive the ID scan from the parsed source tree, or include every ID-bearing element (including `undefined`), rather than hard-coding only `note` and `td`.

fangshuyu-768 (approved):
Reviewed the latest head. The previously reported duplicate-ID coverage issues are addressed, targeted lint tests pass, and CI is green.

PR #2197, opened 2026-08-05
feat(slides): accept slide XML files in +create

R0bynZhu:
Assembling the --slides JSON array by hand is what callers keep getting wrong. A page of SML is multi-line and quote-heavy, and shell has no built-in way to JSON-escape it, so callers reached for `jq -n --rawfile` to build the array. In environments without jq the substitution silently became an empty string and the command ran on to create an empty deck, or the half-escaped XML reached the backend and came back as an opaque 3350001 after the presentation already existed.

Two input forms remove the escaping step:

  --slides now declares Input{file, stdin}, so a finished array can be
  read with `--slides @deck.json` or piped in with `--slides -`.

  --slide is repeatable, takes one complete <slide> document (or @path),
  and the CLI assembles the array. Repetition order is page order.

The forms are mutually exclusive: merging them would make page order depend on flag-parsing rules nobody wants to reason about.

Notes on the repeatable flag: the framework only resolves Flag.Input for single-valued string flags, so --slide resolves @path itself, through the same cmdutil.ReadInputFile the framework uses, keeping the "relative path under the current directory" rule identical. It rejects "-" outright, because a process has one stdin and that cannot mean "this occurrence" on a repeatable flag; the error names both forms that work.

Structural validation now runs on the assembled array, so both forms fail the same way, and it runs before the create call so a malformed page can no longer leave an orphaned empty presentation behind.

Also threads the source flag name through uploadSlidesPlaceholders, which previously reported +add-slide upload failures as --slides.

Docs: the create/troubleshooting references now teach the file inputs instead of the jq array-building template, and the follow-up snippet uses the CLI's own --jq instead of piping to an external jq.

## Summary by CodeRabbit

* **New Features**
  * Create presentations with repeated `--slide` XML inputs or `--slides` JSON from a file or stdin, preserving slide order and supporting local image placeholders.
* **Bug Fixes**
  * Improved validation for conflicting inputs, missing or empty values, malformed slides, unsupported stdin usage, and the 10-slide limit.
  * Errors now identify the input option that caused the issue and preserve API failure details.
* **Documentation**
  * Updated slide creation, troubleshooting, examples, and test coverage guidance for the new input methods.


ethan-zhx (changes requested):
Request changes. Reviewed current head `c07b737840e9064056ae406195585a2c4161b38e` against merge-base `9759167542f279a7ac131c166a9033dc637bb96c`, checked current review threads, and used static verification only. Remote CI is green; I did not run local tests.

P1 - `--slides null` creates a blank deck instead of rejecting invalid input
Location: `shortcuts/slides/slides_create.go:261`
Trigger: `lark-cli slides +create --title Deck --slides null --as user`
Problem: `json.Unmarshal("null", &slides)` succeeds and leaves `slides == nil`; the later `len(slides) == 0` path skips slide validation and execution proceeds to create the presentation. That reintroduces the PR core failure mode: invalid or failed slide input can be reported as a successful empty deck.
Example: input `--slides null` -> current result: `POST /xml_presentations` runs and no slide.create call is planned or executed.
If fixed: `--slides null` returns a typed `validation/invalid_argument` error before any API call, while `--slides []` can remain the explicit empty-array behavior.
If not fixed: generators that marshal a nil `[]string` as `null` leave users with orphaned blank presentations and a success envelope.
Smallest safe repair: after unmarshalling `--slides`, reject `slides == nil` with the same `--slides invalid JSON, must be an array of XML strings` validation shape, and add a no-API-call regression for `--slides null`.

P2 - Skill preflight rule omits the still-supported `--slides` path
Location: `skills/lark-slides/SKILL.md:109`
Trigger: an agent follows the skill and uses documented `slides +create --slides @deck.json` or `--slides -`.
Problem: the critical lint gate now names `slides +create --slide`, `+add-slide`, and `+replace-pages`, but the create reference still documents `--slides` as a supported way to submit complete `<slide>` XML. That lets the JSON-array create path bypass the mandatory `xml_text_overlap_lint.py` workflow in the skill instructions.
Example: input `slides +create --slides @deck.json` -> current skill guidance does not require local lint before calling the API.
If fixed: both `--slide` and `--slides` users get the same preflight requirement before page XML is submitted.
If not fixed: agents can create decks through a supported path without the lint gate that catches overlap, overflow, and structural slide issues.
Smallest safe repair: update line 109 to include `slides +create --slides` alongside `slides +create --slide`.

shortcuts/slides/slides_create.go
```diff
@@ -226,6 +226,111 @@ var SlidesCreate = common.Shortcut{
...
+	// Structural checks run on the assembled array, so both input forms fail the
+	// same way. The backend rejects a non-<slide> payload with an opaque 3350001
+	// after the presentation already exists; catching it here keeps the deck from
+	// being created at all.
+	for i, slideXML := range slides {
+		slides[i] = strings.TrimSpace(slideXML)
+		if slides[i] == "" {
+			return nil, param, errs.NewValidationError(errs.SubtypeInvalidArgument, "%s: page %d is empty", param, i+1).WithParam(param)
+		}
+		if err := validateCompleteSlideXML(slides[i]); err != nil {
```
fangshuyu-768, line 295: [P2] Reject unsupported XML declarations before creating the deck

`validateCompleteSlideXML` accepts processing instructions, so a new `--slide @page.xml` file with a common `<?xml ...?>` prolog passes this preflight. The repository’s own `xml_text_overlap_lint.py` classifies that input as `sxsd_unsupported_declaration` because the Slides SXSD write format does not support declarations. `Execute` will therefore create the presentation first and only then fail the slide request, recreating the orphaned-empty-deck outcome this preflight is intended to prevent. Please reject (or safely remove) a leading XML declaration before Step 1 and add a regression test.

shortcuts/slides/slides_create.go
```diff
@@ -226,6 +226,111 @@ var SlidesCreate = common.Shortcut{
...
+		case strings.HasPrefix(raw, "@"):
+			path := strings.TrimSpace(raw[1:])
+			if path == "" {
+				return nil, errs.NewValidationError(errs.SubtypeInvalidArgument, "--slide: file path cannot be empty after @ (page %d)", i+1).WithParam("--slide")
+			}
+			data, err := cmdutil.ReadInputFile(runtime.FileIO(), path)
+			if err != nil {
+				return nil, errs.NewValidationError(errs.SubtypeInvalidArgument, "--slide: %v", err).WithParam("--slide").WithCause(err)
+			}
+			slides = append(slides, string(data))
```
fangshuyu-768: [P2] Strip the UTF-8 BOM from `--slide` files

This manual `@file` path bypasses `resolveInputFlags`, but unlike the framework path it appends `ReadInputFile` bytes without removing a leading UTF-8 BOM. A page saved by a BOM-writing editor therefore reaches `validateCompleteSlideXML` with U+FEFF before `<slide>` and is rejected as “non-whitespace text outside root”; meanwhile `--slides @deck.json` already strips the BOM. Apply the same leading-BOM normalization here and add a file-input regression test.

skills/lark-slides/SKILL.md
```diff
@@ -106,7 +106,7 @@ metadata:
 
 **CRITICAL — 新建演示文稿或大幅改写页面时，规划 `asset_need` MUST 遵循 [asset-planning.md](references/asset-planning.md)：只做元数据规划，必须有 `fallback_if_missing`，不得要求真实搜索、下载或上传素材。**
 
-**CRITICAL — 将完整 `<slide>` XML 提交给 `slides +create --slides`、`slides +add-slide`、`xml_presentation.slide create` 或 `slides +replace-pages` 之前，MUST 先把待提交 XML 保存到本地文件并运行唯一版式准出入口 [`scripts/xml_text_overlap_lint.py`](scripts/xml_text_overlap_lint.py)；`summary.error_count` 必须为 0 才能调用接口。**
+**CRITICAL — 将完整 `<slide>` XML 提交给 `slides +create --slide`、`slides +add-slide` 或 `slides +replace-pages` 之前，MUST 先把待提交 XML 保存到本地文件并运行唯一版式准出入口 [`scripts/xml_text_overlap_lint.py`](scripts/xml_text_overlap_lint.py)；`summary.error_count` 必须为 0 才能调用接口。**
```
BD-ZERO: 这里漏了 `slides +create --slides`。同一 PR 的创建文档仍指导 Agent 使用 `--slides @deck.json` / `--slides -`，但这条 MUST 门禁只列了 `--slide`；Agent 按该合法路径执行时，会跳过 `xml_text_overlap_lint.py`，而 CLI 的结构校验无法发现重叠、越界、空白页等版式问题。

请同时覆盖 `slides +create --slide` 和 `slides +create --slides`，并明确 `--slides` 数组中的每个 XML 元素都要先保存并通过 lint（`summary.error_count == 0`）后才能调用接口。

ethan-zhx (dismissed):

ethan-zhx (dismissed):

liangshuo-1 (dismissed):

liangshuo-1 (approved):

PR #2221, opened 2026-08-06
feat(base): require --fields on +table-create

CarolSum:
## Summary

`+table-create` without `--fields` creates a table with the platform default schema. Those default fields then coexist with every field the caller adds afterwards, and no field command removes them all — the only clean recovery is to delete the table and start over. This makes `--fields` required, so the schema is declared up front, matching what `+base-create` already recommends via `--table-name` + `--fields`.

Concretely, the failure this prevents: an agent runs `+table-create --name X`, gets a table carrying the default schema, adds its own N fields, and ends up with a table of N + default-schema fields — while reporting N to the user, because the extra columns were never part of its plan.

## Changes

- Mark `--fields` `Required` on `+table-create`, so `--help` and the machine-readable schema both advertise it.
- Reject blank / non-array / empty-array `--fields` in `Validate`. Cobra's `MarkFlagRequired` only checks that the flag was *set*, so `--fields ""` and `--fields "[]"` would still reach the API with no `fields` body and silently fall back to the default schema. Both now return a typed `*errs.ValidationError` with `param: --fields` and a hint. This mirrors the two-layer pattern already used by `sheets/+history-revert`.
- `Validate` runs ahead of the dry-run branch, so `--dry-run` can no longer preview an invocation the real call would reject.
- Sharpen the flag description and add a worked example + a tip explaining *why* the flag is required.
- Update `skills/lark-base/SKILL.md`, `tests/cli_e2e/base/coverage.md`, and the live e2e helper (all three existing live-e2e callers already pass a schema, so no live coverage is lost).

## Test Plan

- [x] Unit tests pass — `make unit-test` clean; new contract tests in `shortcuts/base/table_create_test.go` cover the `Required` declaration, missing / blank / empty-array / non-object-item rejection, the dry-run gate, and the valid path. Each fails if the corresponding code change is reverted.
- [x] Manual local verification confirms the `lark-cli base +table-create` flow works as expected:
  - missing `--fields` → `{"type":"validation","subtype":"invalid_argument","message":"required flag(s) \"fields\" not set"}`
  - `--fields '[]'` → `{"type":"validation","subtype":"invalid_argument","message":"--fields must define at least one field","param":"--fields","hint":"An empty array is not a schema: ..."}`
  - `--fields '[{"name":"Title","type":"text"}]'` → `POST /open-apis/base/v3/bases/:base_token/tables` with the schema inline, unchanged from before
- [x] `go vet ./...`, `gofmt -l .`, `go mod tidy` (no change), `golangci-lint run --new-from-rev=origin/main` (0 issues)

## Notes for reviewers

This is a **breaking change**: `+table-create --base-token <t> --name <n>` without `--fields` now fails instead of creating a default-schema table. Callers that relied on create-empty-then-add-fields must pass the schema to `--fields`. If you would rather stage it, the same validation could first ship as a deprecation warning on the fieldless path — happy to rework it that way.

One thing I deliberately left alone: the missing-flag case still surfaces cobra's terse `required flag(s) "fields" not set`, without `param` or `hint`, because enriching that path touches the shared dispatcher in `cmd/root.go` and belongs in its own PR. The flag description and tips carry the guidance in `--help` meanwhile.

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Table creation now requires a non-empty field schema provided through `--fields`.
  * Updated help and guidance clarify field requirements and field schema formatting.

* **Bug Fixes**
  * Invalid, blank, empty, non-object, or malformed field definitions now receive descriptive validation errors.
  * Dry-run validation consistently enforces required field schemas.

* **Tests**
  * Expanded coverage for required fields, invalid schemas, dry runs, valid schemas, and CLI validation metadata.

### DIFF
--- a/shortcuts/base/table_create.go
+++ b/shortcuts/base/table_create.go
@@ -27,10 +27,9 @@ var BaseTableCreate = common.Shortcut{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo"},{"name":"Done"}]}]`},
 	},
 	Tips: []string{
-		`Example: lark-cli base +table-create --base-token <base_token> --name "Tasks" --fields '[{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]'`,
 		"Before using --fields, read lark-base-field-json.md or rely on the same field JSON shape used by +field-create; do not invent field properties.",
 		"The first --fields item becomes the primary field.",
 	},
--- a/shortcuts/base/table_create.go
+++ b/shortcuts/base/table_create.go
@@ -9,10 +9,17 @@ import (
 	"github.com/larksuite/cli/shortcuts/common"
 )
 
+// BaseTableCreate creates a table with an explicit schema. --fields is required
+// at the cli surface (cobra MarkFlagRequired); a missing flag fails before
+// Validate runs with cobra's standard "required flag(s)" error (which the
+// dispatcher classifies as a typed *errs.ValidationError). validateTableCreate
+// still rejects blank, non-array and empty-array values, because cobra accepts
+// --fields "" and --fields "[]" — both of which would reach the API without a
+// fields body and get the platform default schema instead of the caller's.
 var BaseTableCreate = common.Shortcut{
 	Service:     "base",
 	Command:     "+table-create",
-	Description: "Create a table and optional fields/views",
+	Description: "Create a table with an explicit field schema, plus optional views",
 	Risk:        "write",
 	Scopes:      []string{"base:table:create", "base:field:read", "base:field:create", "base:field:update", "base:view:write_only"},
 	AuthTypes:   authTypes(),
@@ -20,9 +27,11 @@ var BaseTableCreate = common.Shortcut{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Desc: `field JSON array for create, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
 	},
 	Tips: []string{
+		`Example: lark-cli base +table-create --base-token <base_token> --name "Tasks" --fields '[{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]'`,
+		"--fields is required: a table created without it gets the platform default schema, and those default fields stay in the table alongside the ones you add.",
 		"Before using --fields, read lark-base-field-json.md or rely on the same field JSON shape used by +field-create; do not invent field properties.",
 		"The first --fields item becomes the primary field.",
 	},
--- a/shortcuts/base/table_create.go
+++ b/shortcuts/base/table_create.go
@@ -27,11 +27,10 @@ var BaseTableCreate = common.Shortcut{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]`},
 	},
 	Tips: []string{
-		`Example: lark-cli base +table-create --base-token <base_token> --name "Tasks" --fields '[{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]'`,
-		"--fields is required and must hold at least one field; omitted, blank and empty-array schemas are rejected before any table is created. Declaring the schema here is the only way to avoid the platform default fields, which stay in the table alongside anything you add later.",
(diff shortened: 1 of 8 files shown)


CarolSum:
Thanks for the review — pushed 38677e1. Two of the three points were valid; one I'd like to push back on with evidence.

**1. Typed metadata on the rejection tests — fixed for the invalid-schema case.**

`base_shortcuts_test.go` now asserts category/subtype/`param` via `assertInvalidArgumentValidation` and the preserved `*json.SyntaxError` cause, rather than just "an error came back". Verified the shape it pins:

```
err type = *errs.ValidationError
category=validation subtype=invalid_argument param="--fields"
cause = unexpected end of JSON input
```

**2. Same request for the missing-flag test — not applicable at that layer.**

The suggestion was to assert `Param == "--fields"` and a wrapped cobra cause in `table_create_test.go`. That contract does not exist there. A missing flag fails inside cobra's `ValidateRequiredFlags`, which returns a plain error; the typed envelope is applied later by the dispatcher. Probed against this branch:

```
err type = *errors.errorString, err = required flag(s) "fields" not set
errs.ProblemOf        → not a typed problem
errors.As(*errs.ValidationError) → no
```

The envelope-level contract is already pinned where it actually holds, in `cmd/root_test.go:388` ("usage error (missing required flag) is typed as invalid_argument with exit 2"), keyed off `cobraUsageErrorMarkers` in `cmd/root.go`. Rather than assert a promise this layer does not make, I kept the text assertion, documented why, and added a guard that fails if the boundary ever moves — so the weaker assertion cannot silently outlive its reason.

**3. Stale fieldless-create guidance — fixed.** Fair catch: both the tip and the skill note read as if the fieldless path were still reachable through `+table-create`. Both now lead with the rejection (omitted / blank / empty-array, before any table is created) while keeping *why* the schema has to be declared here.

Re-ran the full gate after the changes: `make unit-test` clean, `go vet`, `gofmt -l .`, `go mod tidy` (no change), `golangci-lint --new-from-rev=origin/main` → 0 issues.

---

**Heads-up on an overlap:** #2153 also rewrites `validateTableCreate` — it adds `--view` parsing there, while this PR adds `--fields` validation. The two are complementary in intent but will conflict textually in `shortcuts/base/table_ops.go` (and lightly in `base_shortcuts_test.go` / `skills/lark-base/SKILL.md` / `coverage.md`). Whichever lands first, I'm happy to rebase this one on top — just say the word.

skills/lark-base/SKILL.md
```diff
@@ -81,6 +81,7 @@ metadata:
 - `base-block` 只负责资源目录管理，包括创建资源、移动到 folder、重命名和删除；具体资源内容仍走 table/dashboard/workflow 命令。
 - 新建 Base 时，强烈推荐一次性执行 `lark-cli base +base-create --name "<base>" --table-name "<table>" --fields '<field-json-array>'`，同时配置新 Base 里唯一一个初始数据表的 name 和 schema；使用 `--fields` 前先读 [lark-base-field-json.md](references/lark-base-field-json.md) 或复用 `+field-create` 的字段 JSON 形状，不要猜字段属性。
 - `+base-create` 不传 `--table-name` 和 `--fields` 时，会创建一个默认 schema 的初始数据表。
+- 在已有 Base 中建表时，`+table-create` 的 `--fields` 是必填项：一次性传完整字段数组，不要先建空表再逐个补字段。省略、留空或传空数组都会在建表前被直接拒绝——因为只有在这里声明 schema 才能避开平台默认字段，那些默认字段一旦生成就会和你后补的字段并存。
```
timzhong1024: 这句话和上面信息重复了，删除掉

tests/cli_e2e/base/coverage.md
```diff
@@ -83,7 +83,7 @@
 | ✓ | base +role-get | shortcut | base_role_workflow_test.go::TestBase_RoleWorkflow/get as bot | `--base-token`; `--role-id` | |
 | ✓ | base +role-list | shortcut | base_role_workflow_test.go::TestBase_RoleWorkflow/list as bot | `--base-token` | |
 | ✓ | base +role-update | shortcut | base_role_workflow_test.go::TestBase_RoleWorkflow/update as bot | `--base-token`; `--role-id`; `--json` | |
-| ✓ | base +table-create | shortcut | base/helpers_test.go::createTableWithRetry | `--base-token`; `--name`; optional `--fields`; optional `--view` | helper asserts table id |
+| ✓ | base +table-create | shortcut | base/helpers_test.go::createTableWithRetry | `--base-token`; `--name`; `--fields` (required); optional `--view` | helper asserts table id |
```
timzhong1024: 冗余了，不写 optional 就是 required，去掉这个改动

shortcuts/base/table_create.go
```diff
@@ -9,20 +9,29 @@ import (
...
 	AuthTypes:   authTypes(),
 	Flags: []common.Flag{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Desc: `field JSON array for create, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
 	},
 	Tips: []string{
+		`Example: lark-cli base +table-create --base-token <base_token> --name "Tasks" --fields '[{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]'`,
```
timzhong1024: 这里写错了 [{"name":"Todo"},{"name":"Done"}]，参考 lark-base-field-json.md 重新写

shortcuts/base/table_create.go
```diff
@@ -9,20 +9,29 @@ import (
...
+		"--fields is required and must hold at least one field; omitted, blank and empty-array schemas are rejected before any table is created. Declaring the schema here is the only way to avoid the platform default fields, which stay in the table alongside anything you add later.",
```
timzhong1024: 这一句有点多余

shortcuts/base/table_create.go
```diff
@@ -9,20 +9,28 @@ import (
...
 	AuthTypes:   authTypes(),
 	Flags: []common.Flag{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Desc: `field JSON array for create, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]`},
 	},
 	Tips: []string{
+		`Example: lark-cli base +table-create --base-token <base_token> --name "Tasks" --fields '[{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]'`,
```
timzhong1024: 去掉这一句

shortcuts/base/table_create.go
```diff
@@ -9,20 +9,28 @@ import (
...
+	Description: "Create a table with an explicit field schema, plus optional views",
 	Risk:        "write",
 	Scopes:      []string{"base:table:create", "base:field:read", "base:field:create", "base:field:update", "base:view:write_only"},
 	AuthTypes:   authTypes(),
 	Flags: []common.Flag{
 		baseTokenFlag(true),
 		{Name: "name", Desc: "table name", Required: true},
 		{Name: "view", Desc: "view JSON object/array for create"},
-		{Name: "fields", Desc: `field JSON array for create, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","options":[{"name":"Todo"},{"name":"Done"}]}]`},
+		{Name: "fields", Required: true, Desc: `field JSON array defining the table schema; must hold at least one field, e.g. [{"name":"Title","type":"text"},{"name":"Status","type":"select","multiple":false,"options":[{"name":"Todo","hue":"Blue","lightness":"Lighter"},{"name":"Done","hue":"Green","lightness":"Light"}]}]`},
```
timzhong1024: hue 和 lightness 不是必要的，去掉，只留 name

timzhong1024 (approved):

zgz2048 (approved):

PR #2238, opened 2026-08-07
fix(wiki): correct copy semantics and rename routing

fangshuyu-768:
## Summary

Fix two deterministic Wiki agent failures: `+node-copy` was documented as recursive even though it copies only the requested node, and Wiki rename tasks were not routed to the existing `drive +update-title` shortcut.

## Changes

- Clarify in `+node-copy` help and the lark-wiki reference that copying is non-recursive.
- Route in-place Wiki node renames to `drive +update-title` in the lark-wiki Skill, so agents avoid raw API probing and receive the shortcut's scope recovery hints.
- Replace the stale copy-then-delete migration note with the existing `wiki +move` shortcut.

## Test Plan

- [x] Unit tests pass (`make unit-test`)
- [x] Wiki shortcut unit tests pass (`go test ./shortcuts/wiki`)
- [x] Manual live verification: copied a real parent node with one child and confirmed the copied node had zero children
- [x] Manual live verification: renamed a newly created Wiki node with `drive +update-title` and confirmed the same `node_token` returned the new title
- [x] `go vet ./...`
- [x] `gofmt -l .` returns no files
- [x] golangci-lint reports 0 issues

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Wiki node copying now copies only the selected node and its content, excluding descendants.
* **Documentation**
  * Added guidance for renaming Wiki nodes in place while preserving their existing identifiers.
  * Clarified authorization and scope feedback for title updates.
  * Updated migration guidance to handle descendants separately and use Wiki move functionality when relocating nodes.
* **Tests**
  * Added coverage confirming that Wiki node copies are non-recursive.

### DIFF
--- a/shortcuts/wiki/wiki_list_copy_test.go
+++ b/shortcuts/wiki/wiki_list_copy_test.go
@@ -549,6 +549,15 @@ func TestWikiNodeCopyDeclaredHighRiskWrite(t *testing.T) {
 	}
 }
 
+func TestWikiNodeCopyDeclaresNodeOnlySemantics(t *testing.T) {
+	t.Parallel()
+
+	tips := strings.Join(WikiNodeCopy.Tips, " ")
+	if !strings.Contains(tips, "current node only") || !strings.Contains(tips, "descendant nodes are not copied") {
+		t.Fatalf("WikiNodeCopy.Tips = %q, want explicit non-recursive copy guidance", tips)
+	}
+}
+
 func TestWikiNodeCopyCopiesNodeToTargetSpace(t *testing.T) {
 	t.Setenv("LARKSUITE_CLI_CONFIG_DIR", t.TempDir())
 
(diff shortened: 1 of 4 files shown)


fangshuyu-768:
Manual live verification completed in an authorized tenant: created a new Wiki docx node, renamed it in place with `drive +update-title --url <wiki-url> --title <new-title>`, and read it back. The Wiki `node_token` stayed unchanged, `wiki +node-get` returned the new title, and `drive +inspect` on the underlying docx token returned the same new title. No existing document was modified.

fangshuyu-768:
Live verification completed against a real personal Wiki space using the PR branch:

- Created a source parent node and one direct child.
- Confirmed the source parent reports `has_child: true` and `+node-list` returns exactly 1 child.
- Copied the parent with `wiki +node-copy` into the same space.
- Confirmed the copied node reports `has_child: false` and `+node-list` returns 0 children.

This verifies the API/CLI behavior is node-only, not recursive. Tokens and internal document links are intentionally omitted.

liujinkun2025 (approved):

PR #2243, opened 2026-08-07
docs: add CLI tool development knowledge entry

wanghm25:
## Summary

- add the same-checkout `lark-cli.local-page@1` route entry
- point agents to current command, shortcut, metadata, skill, and Base field sources
- document that client payload, MCP schema, and backend field support are separate contracts
- keep candidate Route use explicit rather than making it a default gate

## Verification

- verified every referenced path against current `upstream/main`
- validated the local-page contract through the cross-repository preflight
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Added a Lark CLI development guide covering source-of-truth boundaries, command and metadata ownership, Base field verification, and CLI versus backend contracts.
  * Documented appropriate use of the `lark-cli.tool-dev` route.

### DIFF
--- a/docs/knowledge/tool-dev-entry.md
+++ b/docs/knowledge/tool-dev-entry.md
@@ -1,0 +1,40 @@
+<!-- knowledge-role: route-entry; knowledge-contract: lark-cli.local-page@1; source-of-truth: current-checkout -->
+
+# Lark CLI Tool Development Entry
+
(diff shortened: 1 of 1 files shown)


wanghm25:
Closing because the upstream CLI repository cannot carry this knowledge-base-specific local page. The central knowledge Route now validates the existing CLI checkout and required sources directly, without requiring CLI repository changes.

PR #2267, opened 2026-08-10
fix: classify invalid CLI input as validation errors

evandance:
## Summary

Invalid CLI input surfaced by Cobra could be emitted as `internal/unknown` (exit 5), telling agents the CLI failed when they only needed to fix the command. Classification depended on error wording, so new validators and Cobra wording changes could silently be misclassified. This PR classifies errors at their source and keeps that result consistent across rendering and lifecycle hooks.

## Changes

| Error source | Result |
| --- | --- |
| Cobra `Args`, command discovery, required/group flags, and lazy completion arguments | `validation/invalid_argument` |
| Untyped failures from `PersistentPreRunE`, `PreRunE`, `RunE`, `PostRunE`, `PersistentPostRunE`, and plugin wrappers | `internal/unknown` |
| Framework setup or output failures, including credential-provider lookup and `--version` writes | `internal/unknown` |
| Existing typed errors, `BareError`, and `PartialFailureError` | Preserved unchanged |

- Instrument the final command tree after plugin hooks and custom help are installed. Classification is stateless, so repeated `Execute` calls and separate command trees cannot contaminate one another.
- Make shortcut positional-argument failures actionable by naming every stray word and pointing at the command's flags. `param` stays unset, because a shortcut declares no positional parameter for a stray word to belong to.
- Preserve the producer's classification if envelope rendering fails.
- Decide the stderr envelope and the exit code before `Shutdown` fires, and write the envelope after it. Deciding early is what stops a lifecycle handler from changing what the user was told, without requiring every error type to be cloneable. Writing late keeps the envelope the trailing content of stderr: the hook layer warns about a failing or skipped handler on the same stream, and a reader recovers the envelope by scanning back from the end.
- Give each lifecycle handler its own copy of the error where the SDK owns the value; a wrapped chain is shared whole so `errors.Is` and `errors.As` still reach the producer's sentinels.

## Test Plan

- [x] Identical errors from `Args` and all five error-returning execution hooks classify by origin and preserve their causes.
- [x] Cover required/group flags, repeated execution, custom help, lazy `__complete`, plugin short-circuits, `--version` write failures, and silent short writes.
- [x] Cover shortcut diagnostics, render fallback, credential-provider classification, and pass-through/lifecycle-copy contracts for typed errors, `BareError`, `PartialFailureError`, wrapped chains, and extension-defined typed errors.
- [x] A failing Shutdown hook writes its warning to stderr without displacing the envelope: the trailing content is still the JSON object, and the handler cannot rewrite its category or the exit code.
- [x] Local gates: `make unit-test`, `make vet`, `make fmt-check`, `go mod tidy`, golangci-lint, repository contract lint, lint module tests, `make build`, and `make quality-gate`.

## Related Issues

- None

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -236,9 +236,14 @@ func buildInternalWithConfig(ctx context.Context, inv cmdutil.InvocationContext,
 	}
 
 	rootCmd.SetContext(ctx)
-	rootCmd.SetIn(cfg.streams.In)
-	rootCmd.SetOut(cfg.streams.Out)
-	rootCmd.SetErr(cfg.streams.ErrOut)
+	// f.IOStreams is cfg.streams with any stream the caller left unset filled
+	// in, so cobra and the commands write to the same three destinations.
+	rootCmd.SetIn(f.IOStreams.In)
+	// Cobra renders framework output such as --version through this writer,
+	// before it reaches any command callback. Type a writer failure at that
+	// boundary so a broken stdout pipe is never reported as invalid input.
+	rootCmd.SetOut(internalErrorWriter{Writer: f.IOStreams.Out})
+	rootCmd.SetErr(f.IOStreams.ErrOut)
 
 	// Root-only usage template (curated Usage synopsis + skills footer); see
 	// rootUsageTemplate.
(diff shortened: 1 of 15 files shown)


evandance:
recheck

PR #2279, opened 2026-08-11
fix(drive): harden export and push failure recovery

fangshuyu-768:
## Summary

Harden Drive export and push failure recovery for the error patterns observed in production.

### Export recovery

- Disambiguate code `9499` only at the Drive export task boundary using its exact throttling message, leaving the shared classifier unchanged and preserving `invalid_parameters` everywhere else.
- Add typed mappings for export codes `1069902`, `1069906`, `1069914`, and `1069918`.
- Stop export polling immediately on the `9499` rate-limit variant and return ticket-aware resume guidance instead of amplifying throttling.
- Add actionable recovery hints for permission, deleted document, invalid token/type, and extension/sub-ID failures.

### Push failure handling

- Keep upload size mismatch (`1062009`) file-local so independent files continue, while stopping the batch on shared failures such as resource contention, quota exhaustion, multipart internal errors, and HTTP/network failures.
- Add bounded-backoff guidance for retryable push failures without inventing a retry delay that the upstream response did not provide.
- Register Drive multipart error `1663` as a retryable server error.
- Document finite retry behavior and the expanded push failure contract.

### Tests and docs

- Add command-level regression coverage for export task creation/polling and permanent export failures.
- Add push coverage for size mismatch, HTTP 502, resource contention, quota exhaustion, and Drive code `1663`.
- Update the Drive export and push skill references.

## Test Plan

- [x] `go test -race -count=1 ./internal/errclass ./shortcuts/drive`
- [x] `make vet`
- [x] `make fmt-check`
- [x] `go test -C lint ./... -count=1`
- [x] `go run -C lint . --changed-from origin/main ..`
- [x] `make quality-gate`
- [x] Command-level HTTP stubs verify export and push API/error boundaries.

`make unit-test` passed all changed and Drive-related packages under `-race`. The aggregate run encountered the repository's unrelated macOS `TempDir RemoveAll` flake in `internal/qualitygate/rules` (`.git: directory not empty`); the failing package passed when rerun in isolation. Live rate-limit and destructive push E2E cases were not forced because they would be unsafe or nondeterministic; deterministic command-level HTTP stubs cover the changed behavior.

## Related Issues

- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Drive export error handling and recovery guidance for permissions, deleted files, invalid references, and unsupported formats.
  * Correctly distinguishes rate limits from other errors and applies safer retry behavior.
  * Drive push operations now provide clearer failure classifications, quota handling, and retry guidance.
  * Pull and sync operations continue processing remaining files when an individual download is unavailable.

* **Documentation**
  * Updated Drive export and push guidance with backoff, retry timing, and failure-handling instructions.

### DIFF
--- a/shortcuts/drive/drive_errors.go
+++ b/shortcuts/drive/drive_errors.go
@@ -200,7 +200,7 @@ func withDriveExportCreateRateLimitRecovery(err error) error {
 		return err
 	}
 
-	const hint = "export task creation was rate limited before a ticket was issued; stop and wait at least 1 minute, then rerun the same `lark-cli drive +export` command\nif rate limiting continues, use exponential backoff starting at 1 minute instead of retrying immediately; do not run `lark-cli drive +task_result` because no export ticket exists yet"
+	const hint = "export task creation was rate limited before a ticket was issued; stop and wait at least 1 minute, then rerun the original command with the same arguments\nif rate limiting continues, use exponential backoff starting at 1 minute instead of retrying immediately; do not run `lark-cli drive +task_result` because no export ticket exists yet"
 	return appendDriveExportRecoveryHint(err, hint)
 }
 
--- a/shortcuts/drive/drive_errors.go
+++ b/shortcuts/drive/drive_errors.go
@@ -146,7 +146,31 @@ func driveExportIsRateLimit(err error) bool {
 	if !ok || problem == nil {
 		return false
 	}
-	return problem.Subtype == errs.SubtypeRateLimit || problem.Code == 99991400
+	return problem.Subtype == errs.SubtypeRateLimit ||
+		problem.Code == 99991400 ||
+		(problem.Code == 9499 && driveExportTooManyRequestsMessage(problem.Message))
+}
+
+// normalizeDriveExportRateLimit scopes the overloaded 9499 interpretation to
+// Drive export task creation/status calls. The shared classifier intentionally
+// keeps the code's general meaning (invalid_parameters) for every other API.
+func normalizeDriveExportRateLimit(err error) error {
+	problem, ok := errs.ProblemOf(err)
+	if !ok || problem == nil || problem.Code != 9499 || !driveExportTooManyRequestsMessage(problem.Message) {
+		return err
+	}
+	problem.Category = errs.CategoryAPI
+	problem.Subtype = errs.SubtypeRateLimit
+	problem.Retryable = true
+	// BuildAPIError attached invalid-parameter recovery before the export
+	// boundary could disambiguate 9499. Drop it before adding rate-limit recovery.
+	problem.Hint = ""
+	return err
+}
+
+func driveExportTooManyRequestsMessage(message string) bool {
+	normalized := strings.Trim(strings.ToLower(strings.TrimSpace(message)), " .")
+	return normalized == "too many request" || normalized == "too many requests"
 }
 
 // withDriveExportRateLimitRecovery preserves the upstream typed rate-limit
@@ -154,6 +178,7 @@ func driveExportIsRateLimit(err error) bool {
 // task already exists at this point, so callers must reuse its ticket instead
 // of creating a duplicate task with drive +export.
 func withDriveExportRateLimitRecovery(err error, ticket, fileToken string) error {
+	err = normalizeDriveExportRateLimit(err)
 	if !driveExportIsRateLimit(err) {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


shortcuts/drive/drive_push.go
```diff
@@ -617,31 +617,59 @@ func driveClassifyBatchFailure(err error) driveBatchFailureDecision {
...
+		}
 	case problem.Code == 1061044:
 		decision.Class = "parent_node_missing"
 		decision.Terminal = true
 		decision.Hint = "The destination parent folder no longer exists or is not visible. Verify --folder-token, folder permissions, and whether a parent directory was deleted during push before retrying."
 	case problem.Subtype == errs.SubtypeNotFound || problem.Code == 1061007:
 		decision.Class = "remote_not_found"
+	case problem.Category == errs.CategoryNetwork:
+		decision.Class = string(problem.Subtype)
+		decision.Terminal = true
```
fangshuyu-768: [P2] Keep item-scoped stream failures non-terminal outside +push

`driveClassifyBatchFailure` is also called by `drivePullFailedItem` and `driveSyncFailedItem`. `DoAPIStream` classifies ordinary HTTP 4xx responses as `network/transport`, so a file that disappears after the initial listing produces HTTP 404 here and now sets `Terminal=true`. That makes `+pull` and `+sync` abort the remaining independent files, even though this PR introduces the batch-stop policy for `+push`. Please scope this terminal rule to `+push`, or exempt item-scoped HTTP 4xx responses (at least 404), and add pull/sync regression coverage.

wittam-01 (dismissed):

shortcuts/drive/drive_push.go
fangshuyu-768: Fixed in 11b13c66 by applying the new conflict/quota/network stop policy only in `drivePushFailedItem`. Added command-level regressions proving `+pull` and `+sync` continue after an item-scoped HTTP 404.

shortcuts/drive/drive_errors.go
```diff
@@ -170,10 +195,37 @@ func withDriveExportRateLimitRecovery(err error, ticket, fileToken string) error
 // task exists. There is no ticket to resume, so callers must retry the original
 // export command after backing off instead of invoking drive +task_result.
 func withDriveExportCreateRateLimitRecovery(err error) error {
+	err = normalizeDriveExportRateLimit(err)
```
wittam-01, line 198: [P2] Preserve the caller-specific export retry command

This normalization newly routes code 9499 `too many request(s)` failures into the creation-rate-limit hint below, but that hint hard-codes `lark-cli drive +export`. The same `drive.RunExport` path is used by `sheets +workbook-export`, so a throttled Sheets export is told to rerun the wrong command. Please pass caller-specific retry context into this helper (or say "rerun the original command") and add a `sheets +workbook-export` 9499 regression test.

shortcuts/drive/drive_push.go
```diff
@@ -617,31 +633,56 @@ func driveClassifyBatchFailure(err error) driveBatchFailureDecision {
...
 		decision.Class = "parent_sibling_limit"
 		decision.Terminal = true
 		decision.Hint = "The destination parent folder has reached its child-count limit. Clean up that folder, choose another --folder-token, or split the upload across subfolders before retrying."
-	case problem.Subtype == errs.SubtypeQuotaExceeded || problem.Code == 1061043:
+	case problem.Code == 1061043:
 		decision.Class = "file_size_limit"
-	case problem.Code == 1062009:
-		decision.Class = "upload_size_mismatch"
+		decision.Hint = "Do not retry the same file unchanged; split it into smaller files or use another storage method."
+	case problem.Subtype == errs.SubtypeQuotaExceeded:
```
fangshuyu-768, line 659: [P2] Classify the documented user-quota code before continuing

This branch only makes quota failures terminal when BuildAPIError has already assigned quota_exceeded. The Drive upload APIs also return 1061061 for user quota exceeded, but that code is not registered in codeMeta, so it falls through as unknown with terminal=false and +push keeps attempting the remaining files despite the shared quota condition. Please map 1061061 to quota_exceeded and add a regression that aborts the batch just like 1061101.

shortcuts/drive/drive_errors.go
fangshuyu-768: Fixed in 53c77abc by making export-task creation recovery tell callers to rerun the original command with the same arguments. Added a sheets +workbook-export regression for code 9499 that verifies the hint does not redirect users to drive +export.

shortcuts/drive/drive_pull_test.go
```diff
@@ -1092,6 +1092,66 @@ func TestDrivePullAbortsAfterDownloadForbidden(t *testing.T) {
...
+	if got := summary["failed"]; got != float64(1) {
+		t.Fatalf("summary.failed = %v, want 1", got)
+	}
+	if got := summary["downloaded"]; got != float64(1) {
+		t.Fatalf("summary.downloaded = %v, want 1", got)
+	}
+	if len(items) != 2 || items[0]["rel_path"] != "a.txt" || items[1]["rel_path"] != "b.txt" || items[1]["action"] != "downloaded" {
+		t.Fatalf("item-local 404 must not block b.txt: %#v", items)
+	}
+	mustReadFile(t, filepath.Join("local", "b.txt"), "BBB")
```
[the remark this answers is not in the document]
fangshuyu-768: Addressed in 53c77abc. Both pull and sync regressions now assert action=failed, phase=download, HTTP 404, and retryable=false while preserving the continuation checks.

shortcuts/drive/drive_push.go
fangshuyu-768: Fixed in 53c77abc by mapping Drive code 1061061 to quota_exceeded, documenting tenant/user quota recovery, and adding a command-level regression that proves +push aborts before the next file.

shortcuts/sheets/lark_sheet_workbook_export_test.go
```diff
@@ -99,3 +101,34 @@ func TestWorkbookExport_ExecuteExportOnly(t *testing.T) {
...
+
+	_, err := runShortcutWithStubs(t, WorkbookExport, []string{
+		"--url", testURL, "--file-extension", "xlsx", "--as", "user",
+	}, stubs...)
+	problem, ok := errs.ProblemOf(err)
+	if !ok {
+		t.Fatalf("expected typed rate-limit error, got %T: %v", err, err)
+	}
+	if problem.Subtype != errs.SubtypeRateLimit || problem.Code != 9499 || !problem.Retryable {
+		t.Fatalf("problem = %+v, want retryable rate_limit code 9499", problem)
```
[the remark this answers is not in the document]
fangshuyu-768: Addressed the applicable part in 29087fdd: the workbook-export regression now asserts CategoryAPI and uses errors.As to pin the concrete errs.APIError type, while retaining the subtype/code/retryable checks. I did not add a wrapped-cause assertion because this path classifies a nonzero Lark API response directly via BuildAPIError; there is no lower-level transport error to preserve, and fabricating one would not reflect the runtime contract.

shortcuts/drive/drive_push.go
```diff
@@ -603,6 +608,9 @@ func driveClassifyBatchFailure(err error) driveBatchFailureDecision {
 	decision.Code = problem.Code
 	decision.Subtype = string(problem.Subtype)
 	decision.Retryable = problem.Retryable
+	if retryAfter, ok := errs.RetryAfter(err); ok {
+		decision.RetryAfterSeconds = int(retryAfter / time.Second)
+	}
```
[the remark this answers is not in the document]
fangshuyu-768: This no longer applies to the current code. Commit 974b83a1 removed the errs.RetryAfter to RetryAfterSeconds conversion because Drive does not provide a verified retry delay for these failures; the current implementation emits bounded backoff guidance without serializing a fractional duration.

wittam-01 (approved):

liangshuo-1 (approved):

PR #2323, opened 2026-08-13
feat(sheets): answer +sheet-list instead of failing the guess

chendaxin-tk:
## Why

Callers reach for `+sheet-list` on their own. The sheets surface has a whole `+sheet-*` family — `+sheet-create`, `+sheet-copy`, `+sheet-delete`, `+sheet-info` — so "list the sheets" spells itself `+sheet-list`, even though the command that reads workbook structure is `+workbook-info`.

The miss does not self-correct. `internal/suggest.Closest` ranks shared prefixes ahead of edit distance, and `+sheet-list` shares a 7-character prefix with `+sheet-create` but only 1 character with `+workbook-info`. The "did you mean" hint therefore lists unrelated `+sheet-*` commands and never points at the right one.

## What

A read-only projection over `get_workbook_structure` that emits the bare sheets array:

```console
$ lark-cli sheets +sheet-list --spreadsheet-token <token>
{ "ok": true, "identity": "user",
  "data": [ {"sheet_id":"0HSoFo","sheet_name":"Jan","index":0,"row_count":200,
             "column_count":20,"is_hidden":true, ...}, ... ] }
```

Entries are forwarded exactly as the tool returned them, so they are identical to what `+workbook-info` nests under `sheets`. `+workbook-info` keeps one extra top-level field, `revision`, which `+revision-get` already projects on its own.

It reuses the existing read tool rather than adding a backend endpoint — the same shape `+revision-get` already uses for the `revision` field.

## Why hidden

The command exists to answer a name that was guessed, not to offer a second documented way to read workbook structure. Advertising it would leave callers choosing between two commands that do the same thing.

So it is hidden on both surfaces: `Hidden: true` keeps it out of `sheets --help` and completion here, and it is excluded from the lark-sheets skill docs upstream in sheet-skill-spec. `sheets +sheet-list --help` still works and points back at `+workbook-info`.

## Testing

- Package tests: `projectSheets` happy path and the three malformed-response paths; an execute-path test asserting `data` is the bare array with fields intact; a test comparing `+sheet-list`'s output against `+workbook-info`'s `sheets` on the same stubbed response; a dry-run test pinning `get_workbook_structure`; and one asserting the command stays off the help surface.
- Dry-run E2E (`tests/cli_e2e/sheets/sheets_sheet_list_dryrun_test.go`): `--spreadsheet-token` and `--url` both resolve to a single `invoke_read` POST carrying `{"excel_id":"..."}`. The read-vs-write endpoint is asserted deliberately: every sibling shortcut in this area is a write, and a read tool sent to `invoke_write` is rejected by the gateway.
- Live E2E (`tests/cli_e2e/sheets/sheets_sheet_list_workflow_test.go`): creates a workbook and a second sub-sheet, then asserts the payload is an array containing the new sheet, and that it matches `+workbook-info`'s `sheets` verbatim. Self-cleaning via the shared `createSpreadsheet` helper.
- Manually verified against a real 32-sub-sheet workbook: `+sheet-list`'s `data` is byte-identical to `+workbook-info`'s `data.sheets`.

## Notes

`data/flag-defs.json` and `flag_defs_gen.go` carry the new shortcut's flag entry, sourced from the spec repo that owns those artifacts.

Three conventions are intentionally left as-is, matching every other shortcut in this domain rather than making this one command an outlier:

- Output goes through `runtime.Out`, like all 56 sheets shortcuts (none use the format-aware emitter today), so `--format` behaves exactly as it does for `+workbook-info`.
- Scopes declare `sheets:spreadsheet:read` only. The wiki-URL path needs `wiki:node:read`, but no sheets shortcut declares it conditionally today, including `+workbook-info`.
- `projectSheets` reads the response as a loose map, like `projectRevision`. A typed struct would risk dropping fields the verbatim-forwarding contract depends on.

## Summary by CodeRabbit

* **New Features**
  * Added a hidden, read-only `sheets +sheet-list` shortcut.
  * Supports spreadsheet URLs or tokens to retrieve sheet details.
  * Added optional dry-run support and formatted output.
  * Sheet results align with the entries returned by workbook information commands.

* **Bug Fixes**
  * Added validation for missing or malformed workbook and sheet data.

* **Tests**
  * Added unit, end-to-end, workflow, and dry-run coverage for the new shortcut.

### DIFF
--- a/shortcuts/sheets/data/flag-defs.json
+++ b/shortcuts/sheets/data/flag-defs.json
@@ -113,6 +113,32 @@
       }
     ]
   },
+  "+sheet-list": {
+    "risk": "read",
+    "flags": [
+      {
+        "name": "url",
+        "kind": "public",
+        "type": "string",
+        "required": "xor",
+        "desc": "Spreadsheet locator"
+      },
+      {
+        "name": "spreadsheet-token",
+        "kind": "public",
+        "type": "string",
+        "required": "xor",
+        "desc": "Spreadsheet locator"
+      },
+      {
+        "name": "dry-run",
+        "kind": "system",
+        "type": "bool",
+        "required": "optional",
+        "desc": ""
+      }
+    ]
+  },
   "+sheet-create": {
     "risk": "write",
     "flags": [
--- a/shortcuts/sheets/flag_defs_gen.go
+++ b/shortcuts/sheets/flag_defs_gen.go
@@ -883,6 +883,14 @@ var flagDefs = map[string]commandDef{
 			{Name: "dry-run", Kind: "system", Type: "bool", Required: "optional"},
 		},
 	},
+	"+sheet-list": {
+		Risk: "read",
+		Flags: []flagDef{
+			{Name: "url", Kind: "public", Type: "string", Required: "xor", Desc: "Spreadsheet locator"},
+			{Name: "spreadsheet-token", Kind: "public", Type: "string", Required: "xor", Desc: "Spreadsheet locator"},
+			{Name: "dry-run", Kind: "system", Type: "bool", Required: "optional"},
+		},
+	},
 	"+sheet-move": {
 		Risk: "write",
 		Flags: []flagDef{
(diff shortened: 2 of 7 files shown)


chendaxin-tk:
Closing: folding this into feat/call-compat rather than shipping it as its own PR.

PR #2338, opened 2026-08-13
feat(sheets): accept the --range / --cells / border shapes callers actually send

chendaxin-tk:
## Summary

Four input shapes that callers habitually send to the sheets shortcuts are now accepted outright — or, where the meaning is genuinely ambiguous, rejected with a prescription that names the exact fix. Each one is an eval-trace cluster measured against real calls, and each rewrite happens on an existing normalizer seam, so `--writes` items and `+batch-update` sub-ops get it too — the same item translates the same way whichever of the three entry points it arrives through.

## Changes

- **A sheet prefix in `--range` is read as the selector.** `--range "Sheet1!A1:D20"` no longer dies on `specify at least one of --sheet-id or --sheet-name` — the prefix fills `sheet_name` and the bare A1 range reaches the tool. 707 traced calls failed on that error, 53% of them already naming the sheet inside `--range`. Covers the standalone commands (cobra `PreRunE`), `+batch-update` sub-ops and `+cells-set --writes` items alike. An explicit `--sheet-id` / `--sheet-name` stays authoritative: the CLI never rewrites the selector from a prefix that disagrees with it, and never sizes a qualified anchor either — that combination keeps failing locally with the cells-vs-range mismatch instead of shipping a range and a `sheet_name` naming different sheets. Forwarding a disagreeing prefix inside `--range` itself is unchanged pre-existing behavior; how the backend resolves it is not something this PR touches.

- **The openpyxl / gspread `--cells` shapes are accepted.** A `{"cells": […]}` envelope (the flag name mistaken for a JSON key — 11 of 21 traced `expected type "array", got "object"` rejections), a lone cell object without the 2D wrapper, and bare scalars in cell slots all rewrite onto the wire contract; `--values` becomes a silent alias for `--cells`. A bare single-cell `--range` now behaves as an anchor sized from the payload, the same inference `+csv-put` already does for `--start-cell`. `null` cells stay rejected — `{}` and `{"value":""}` are both plausible readings — and the cells-vs-range mismatch (132 rejections across 93 case-runs) now reports both axes at once and hands back the range that fits the payload rather than resizing on the caller's behalf.

- **Border thickness vocabulary gets its observed fallbacks.** openpyxl's `hair` in either the style or the weight slot, a numeric line width, and the Google Sheets `width` key fold onto the `thin` / `medium` / `thick` enum. Everything that scored zero in the trace tally (`dashDot`, `mediumDashed`, `xlContinuous`, CSS `hidden`, …) stays rejected with the enum in the message.

- **The sheet part of a range is parsed with the front-end ref lexer's grammar.** `splitRangeSheetPrefix` and `parseCellRange` now share one scanner, so both agree on what a separator is: the full-width `！`, the backslash-escaped forms of both widths, quoted names with the doubled-quote escape, and a `!` living inside a quoted name (`'Q1!Sales'!A1`). `parseCellRange` keeps the qualifier exactly as written, since ranges rendered from it are both shipped to the server and printed for the caller to paste back.

## Test Plan

- [x] Unit tests pass — `go test ./shortcuts/sheets/`
- [x] Dry-run E2E — `go test ./tests/cli_e2e/sheets/ -run DryRun`; new files `sheets_range_sheet_prefix_dryrun_test.go`, `sheets_cells_shapes_dryrun_test.go` and `sheets_border_vocab_dryrun_test.go` assert method / URL / tool name / full tool input for every accepted shape, and pin that the deliberately-rejected ones still fail
- [x] Manual local verification — each accepted and rejected shape driven through `lark-cli sheets +cells-get / +cells-set / +cells-set-style … --dry-run` and the emitted payload inspected
- [x] Every commit builds and passes both suites in isolation, so the history stays bisectable
- [x] Review follow-up (`4784a4df`) — each reported finding reproduced against a freshly built binary before being fixed or declined, with the outcome recorded under the comment it answers
- [x] Live E2E — `TestSheets_CallCompatWorkflow` builds its own workbook, writes through a quoted sheet prefix with no selector flag (scalar cells, bare anchor), reads back through the same prefix and stamps an openpyxl `hair` border, then tears the workbook down. Skips without tenant credentials; CI's `e2e-live` job is what runs it

## Related Issues

- None

## Review history

Continues #2317 — same commits (`8f5c6ccc`), renamed branch. The four review findings from that PR were reproduced against a freshly built binary and fixed in `4784a4df`, with one nit declined on the record; all 11 conversations there are resolved. Re-review only needs the diff, not that history.

## Summary by CodeRabbit

* **New Features**
  * Added a read-only command to list sheets from a spreadsheet URL or token.
  * Added support for sheet-qualified ranges, including quoted names and escaped separators.
  * Expanded cell writes to automatically size ranges for rectangular payloads.
  * Added flexible cell inputs, including scalar matrices, `cells` envelopes, and `--values` aliases.
  * Added broader border-weight normalization, including numeric widths, `hair`, and `width` forms.
  * Added clearer suggestions for unrecognized Sheets commands.

* **Bug Fixes**
  * Improved validation for mismatched, ragged, empty, or unsupported cell payloads.
  * Preserved explicit sheet selectors when they conflict with range prefixes.


shortcuts/sheets/style_vocab.go
```diff
@@ -345,28 +348,101 @@ func expandBorderAllShorthand(border map[string]interface{}) {
...
+// emit just as often. Deliberately not tableGetToFloat: that one answers
+// "is this cell a number?" for column typing and must NOT accept a quoted
+// number, whereas in the weight slot a quoted number is unambiguously a
+// width.
+func borderLineWidth(v interface{}) (float64, bool) {
+	switch t := v.(type) {
+	case float64:
+		return t, true
+	case string:
+		n, err := strconv.ParseFloat(strings.TrimSpace(t), 64)
```
fangshuyu-768: [P2] Reject non-finite quoted border widths

`strconv.ParseFloat` accepts non-finite strings such as `"Inf"` and `"Infinity"`, so they enter the numeric-width branch and are silently normalized to `"thick"`. I reproduced this with `--border-styles '{"top":{"style":"solid","weight":"Inf"}}'`: the command exits successfully and emits `weight:"thick"`. An infinite line width is not a meaningful supported caller shape, so this should remain on the validation-error path rather than being guessed. Could we reject non-finite results here (or accept only decimal numeric strings) and add a regression test?

shortcuts/sheets/style_vocab.go
```diff
@@ -399,9 +477,12 @@ func normalizeCellsFlagValue(v interface{}) interface{} {
 		if !ok {
 			continue
 		}
-		for _, cellRaw := range row {
+		for i, cellRaw := range row {
 			cell, ok := cellRaw.(map[string]interface{})
 			if !ok {
+				if lifted := scalarCellValue(cellRaw); lifted != nil {
```
fangshuyu-768, line 498: [P2] Preserve numeric precision before lifting scalar cells

`parseJSONFlag` decodes this new scalar/`--values` path into `interface{}`, so numeric cells arrive here as `float64` before they are lifted. For example, a values matrix containing `9007199254740993` exits successfully, but the dry-run request contains `9007199254740992`; the same corruption also occurs through `--writes`. The `json.Number` branch is not reachable from CLI JSON input, while the table/workbook paths already use `UseNumber` to preserve large IDs. Please preserve the original JSON number through parsing and validation, and add a wire-level regression test for the exact digits.

xiongyuanwen-byted (approved):
## Review Summary

I've done a thorough review of the full diff, cross-referenced against the existing codebase design patterns, and ran both unit and dry-run E2E suites. This is a high-quality PR that I'm approving.

### What I checked

- **Design consistency**: All four rewrites land on existing normalize seams (`jsonFlagNormalizers`, `mapFlagView.normalize*`, `chain*` cobra hooks) — same pattern as the existing enum normalization and flag aliases. The three entry points (standalone, `+batch-update` sub-op, `--writes` items) are all covered, and the `TestCellsSetWrites` / batch sub-op tests confirm the same item translates identically whichever path it arrives through.
- **Admission bar**: Each accepted vocabulary word is backed by trace counts documented in the comments. Unobserved words (`dashDot`, `mediumDashed`, CSS `hidden`, etc.) stay rejected with the enum in the message — consistent with the SILENT-ALIAS ADMISSION BAR at the top of `style_vocab.go`.
- **Single authority**: `cellsExtent` is the one source of truth both `expandAnchorRange` and `checkCellsMatchRange` consult — they can't disagree on whether a payload is rectangular. `borderWeightWord` is the single canonical entry for the weight vocabulary, replacing the two separate copies (`expandBorderAllShorthand`'s inline switch + `foldBorderFamilyAliases`'s `borderWeights` map).
- **Lexer grammar**: `scanSheetQualifier` handles full-width `！`, backslash-escaped separators, quoted names containing `!`, and doubled-quote escape — the same set the front-end ref lexer's `ExclamationMark` production accepts. `sheetNameFromA1` now delegates to this grammar, fixing the old `strings.Index(ref, "!")` bugs.
- **Safety**: Explicit `--sheet-id`/`--sheet-name` stays authoritative — the prefix is never rewritten when a selector is already present. `null` cells are deliberately not lifted because `{}` and `{"value":""}` are both plausible readings. A `{"cells": …, "range": …}` object with sibling keys is not unwrapped because dropping them would write the right cells to the wrong place.
- **Tests pass**: `go test ./shortcuts/sheets/` ✓, `go test ./tests/cli_e2e/sheets/ -run DryRun` ✓.

### Minor observations (non-blocking)

1. **`borderLineWidth` and `json.Number`**: The function accepts `float64` and `string` but not `json.Number`. Since the input always comes from `encoding/json`'s standard `Unmarshal` (which decodes numbers as `float64`), this is fine in practice. If someone ever switches to `UseNumber()`, a numeric weight would silently fall through. Not worth changing now — just calling it out for awareness.

2. **`scanQuotedSheetQualifier` whitespace trimming**: The `strings.TrimLeft(tail, " \t\r\n")` cutset is consistent with the `[ \t\r\n]*` comment referencing the lexer. Since the input is pre-trimmed by `splitRangeSheetPrefix`, the leading whitespace case is already handled, but the post-quote whitespace trimming is correct for the lexer's grammar.

3. **`normalizeBorderSideVocab` order**: When both `width` and `weight` keys are present, `width` stays as a stray key for the schema validator to reject. The comment documents this as "contradictory input, left intact for the validator". This is the right call — silently picking one would be guessing.

No blocking issues found. LGTM 🚀

xiongyuanwen-byted:
### Minor observations from detailed review (non-blocking)

1. **`borderLineWidth` and `json.Number`** (`style_vocab.go:1730`): The function accepts `float64` and `string` but not `json.Number`. Since the input always comes from `encoding/json`'s standard `Unmarshal` (which decodes numbers as `float64`), this is fine in practice. If someone ever switches to `UseNumber()`, a numeric weight would silently fall through. Not worth changing now — just calling it out for awareness.

2. **`scanQuotedSheetQualifier` whitespace trimming** (`range_sheet_prefix.go:1304`): The `strings.TrimLeft(tail, " \t\r\n")` cutset is consistent with the lexer's `[ \t\r\n]*` grammar. The outer `splitRangeSheetPrefix` pre-trims the input, so the leading whitespace case is already handled there, but the post-quote whitespace consumption here is correct for the lexer contract.

3. **`normalizeBorderSideVocab` conflicting keys** (`style_vocab.go:1684-1688`): When both `width` and `weight` keys are present in the same side spec, `width` stays as a stray key for the schema validator to reject. The comment documents this as "contradictory input, left intact for the validator." This is the right call — silently picking one would be guessing — but the leftover `width` key will produce a validation error that doesn't explain the conflict. Worth considering whether to add a targeted prescription in a follow-up, but definitely not a blocker.

xiongyuanwen-byted (dismissed):

shortcuts/sheets/style_vocab.go
chendaxin-tk: Fixed in `cae83788`. `borderLineWidth` now reports a non-finite `ParseFloat` result as "not a width", so `Inf` / `Infinity` / `NaN` stay on the enum error path instead of being guessed at.

Verified against a binary built from the pushed head (`4abff0eb`):

```
weight="Inf"       -> --border-styles: top.weight: value "Inf" is not in enum ["thin", "medium", "thick"]
weight="Infinity"  -> ... is not in enum ["thin", "medium", "thick"]
weight="NaN"       -> ... is not in enum ["thin", "medium", "thick"]
weight="2"         -> {"style":"solid","weight":"medium"}   (unchanged)
```

Regression coverage rides along in the same commit. Thanks for the repro — `NaN` had only been escaping the mapping by accident, since every comparison against it is false.

shortcuts/sheets/subcommand_ergonomics.go
```diff
@@ -0,0 +1,117 @@
...
+	"+sheet-add": {
+		Command: "+sheet-create",
+		Hint:    `create a sub-sheet with +sheet-create --title "…" — this domain spells the verb create, never add; --url / --spreadsheet-token carry over unchanged`,
+	},
+	"+col-resize": {Command: "+cols-resize", Hint: colsResizeForm},
+	"+row-resize": {Command: "+rows-resize", Hint: rowsResizeForm},
+	"+cells-put": {
+		Command: "+cells-set",
+		Hint:    `write cell values with +cells-set --range A1:B2 --cells '[[{"value":"a"},{"value":"b"}]]' — each cell is an object, not a bare scalar`,
+	},
```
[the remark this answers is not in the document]
chendaxin-tk: Both halves confirmed and fixed in `9099abf6`. The hint's own example was rejected by the very call it prescribes:

```
$ lark-cli sheets +cells-set --range A1:B2 --cells '[[{"value":"a"},{"value":"b"}]]' --dry-run
--cells is 1 rows × 2 columns but --range "A1:B2" spans 2 rows × 2 columns
```

It now reads `--range A1:B2 --cells '[["a","b"],["c","d"]]'` and describes both accepted cell forms (bare scalar or object), since this branch lifts scalars.

Rather than restate the example in the assertion, `TestPrescribedExamplesActuallyValidate` regexes the flags back out of the hint and runs them through `+cells-set` — so the prose cannot drift from what the validator accepts. Restoring the old hint fails it with the error above.

shortcuts/sheets/lark_sheet_batch_update_test.go
```diff
@@ -713,23 +713,69 @@ func TestBatchUpdate_ResizeNoOperationField(t *testing.T) {
...
+			t.Errorf("split(%q) = (%q,%q,%v), want (%q,%q,nil)", tc.in, sheet, sub, err, tc.sheet, tc.sub)
+		}
 	}
-	if _, _, err := splitSheetPrefixedRange("!A2"); err == nil {
-		t.Error("expected error on empty sheet name")
+	for _, in := range []string{"A2:A100", "!A2", "sheet1!", "'unclosed!A1"} {
+		if _, _, err := splitSheetPrefixedRange(in); err == nil {
+			t.Errorf("split(%q): expected error", in)
+		}
 	}
```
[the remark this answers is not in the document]
chendaxin-tk: Fixed in `9099abf6`. The four rejection cases now go through `requireValidation`, which pins `CategoryValidation` / `SubtypeInvalidArgument`, and additionally assert `Param == "--range"` plus that the message names the offending input. Swapping `sheetsValidationForFlag` for a bare `fmt.Errorf` fails the test, so an untyped error no longer passes.

shortcuts/sheets/style_vocab.go
chendaxin-tk: Reproduced — the corruption is real, and `--writes` carries it too. On the attribution though: it predates this branch rather than arriving with the scalar/`--values` path.

Built a binary at the merge base (`64020803`) and drove it through the cell-object form, which is the pre-existing spelling:

```
$ lark-cli sheets +cells-set --range A1 --cells '[[{"value":9007199254740993}]]' --dry-run
{"value": 9007199254740992}
```

Same result on the current head. The root cause is that `parseJSONFlag` decodes with a plain `json.Unmarshal`, so every composite flag on this path has been going through `float64` since well before this PR — the table/workbook paths are the ones that opted into `UseNumber`. What this branch added is another spelling (a bare scalar, and `--values`) that reaches the same decoder; it widens the surface but is not where the precision goes.

That makes it a real bug with a fix that belongs outside this PR's diff: `parseJSONFlag` switching to `UseNumber` touches every JSON flag in the domain, and `json.Number` then has to be carried through schema validation and each input builder — `scalarCellValue` already has the branch, but the validator side needs its own pass and wire-level tests for the exact digits.

Leaving this thread open rather than resolving it, so it does not get lost. Happy to take it as a follow-up if that split works for you.

xiongyuanwen-byted (approved):

PR #2349, opened 2026-08-14
fix(wiki): reuse 131006 terminal guidance on copy and move

ViperCai:
## Summary

`wiki +node-get` already treats Wiki `131006` as a terminal resource-access error. `wiki +node-copy` and `wiki +move` still returned the generic permission hint, so agents could keep retrying the same copy/move after a node or parent ACL failure.

## Changes

- Extract shared `annotateWikiPermissionDenied` for Wiki `131006`.
- Apply it to `wiki +node-copy` and `wiki +move` (including move-time `get_node` resolve).
- Route `+node-get` / `+node-list` through the same helper so the hint stays in one place.
- Add revert-failing tests and document the recovery rule in the copy/move skill references.

## Test Plan

- [x] `go test ./shortcuts/wiki -count=1`
- [ ] Manual local verification confirms the `lark-cli wiki +node-copy` / `lark-cli wiki +move` 131006 flow works as expected

## Related Issues

- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wiki copy and move error handling with operation-specific permission guidance.
  * Permission-denied errors now identify the required access and are not retried unnecessarily.
  * Asynchronous move tasks stop polling immediately when resource permissions are insufficient.
  * Preserved existing handling for unrelated errors and prevented duplicate guidance.

* **Documentation**
  * Added guidance for resolving resource permission errors during Wiki moves and copies.
  * Clarified access requirements for node moves, Docs-to-Wiki operations, and task-status updates.

### DIFF
--- a/shortcuts/wiki/wiki_helpers.go
+++ b/shortcuts/wiki/wiki_helpers.go
@@ -42,10 +42,83 @@ func appendWikiProblemHint(err error, hint string) error {
 	return err
 }
 
-// wikiPermissionDeniedHint provides stable recovery for wiki's 131006. The
-// service uses one public code for both node and space ACL failures, while the
-// upstream message is informational and normalized by the shared classifier.
-// Keep the command hint accurate without branching on that unstable text.
+// wikiPermissionDeniedHint provides stable recovery for read-path 131006
+// (node-get / node-list). The service uses one public code for both node and
+// space ACL failures, while the upstream message is informational and
+// normalized by the shared classifier. Keep the command hint accurate without
+// branching on that unstable text.
 func wikiPermissionDeniedHint() string {
 	return "The current user or app/bot identity lacks access to the target wiki space or node. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error; ask the resource owner or wiki administrator to grant read access, or use an accessible resource."
+}
+
+// wikiCopyPermissionDeniedHint recovers copy-path 131006. Official copy
+// requires container edit on the source or destination parent; a space-root
+// target can also require wiki space membership or administrator permission.
+func wikiCopyPermissionDeniedHint() string {
+	return "The current user or app/bot identity lacks Wiki container permission for this copy. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error; ask the resource owner or wiki administrator to grant container edit permission on the relevant source or destination parent node. Copying to a space root can also require wiki space membership or administrator permission. Use an accessible resource if that access cannot be granted."
+}
+
+// wikiNodeMovePermissionDeniedHint recovers node-move 131006. Official move
+// requires edit permission on the node plus container edit on the source and
+// destination parents; a space-root target can also require wiki space
+// membership or administrator permission.
+func wikiNodeMovePermissionDeniedHint() string {
+	return "The current user or app/bot identity lacks permission to move this Wiki node. Official node move requires edit permission on the node plus container edit permission on the source and destination parent nodes. Moving to a space root can also require wiki space membership or administrator permission. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error. Use an accessible resource if that access cannot be granted."
+}
+
+// wikiDocsToWikiPermissionDeniedHint recovers docs-to-wiki write 131006.
+// Official guidance can mean the source Drive document lacks manage
+// permission or its parent folder lacks edit permission; the destination
+// Wiki parent still needs container edit, and a space-root target can
+// require wiki space membership or administrator permission.
+func wikiDocsToWikiPermissionDeniedHint() string {
+	return "The current user or app/bot identity lacks permission to move this Drive document into Wiki. Official docs-to-wiki 131006 can mean the source document lacks manage permission or its parent folder lacks edit permission; the destination Wiki parent needs container edit permission, and a space-root target can require wiki space membership or administrator permission. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error."
+}
+
+// wikiTaskPermissionDeniedHint recovers task-get 131006. Official lookup
+// can mean only the task creator can query status, the destination Wiki
+// space or node is inaccessible, or the source Drive document lacks manage
+// permission and its parent folder lacks edit permission.
+func wikiTaskPermissionDeniedHint() string {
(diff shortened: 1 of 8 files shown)


shortcuts/wiki/wiki_helpers.go
```diff
@@ -49,3 +49,14 @@ func appendWikiProblemHint(err error, hint string) error {
...
+
+// annotateWikiPermissionDenied marks wiki 131006 as a terminal resource-access
+// failure and attaches the shared recovery hint. Other errors pass through.
+func annotateWikiPermissionDenied(err error) error {
+	p, ok := errs.ProblemOf(err)
+	if !ok || p == nil || p.Code != 131006 {
+		return err
+	}
+	p.Retryable = false
+	return appendWikiProblemHint(err, wikiPermissionDeniedHint())
```
fangshuyu-768: [P2] Avoid prescribing read access for write failures

This helper is now used by +node-copy and +move, but wikiPermissionDeniedHint tells the resource owner to grant read access. That is insufficient for 131006 responses from these write endpoints: the official Wiki API guidance requires container edit permission on the relevant source/destination parent, and moving to a space root can require space membership/admin rights. The command now stops retries correctly, but its recovery step can still leave the operation failing after the requested grant. Please make the shared text permission-neutral (for example, grant the required resource access), or retain a read-specific hint only for get/list and use write-specific guidance here.

References: https://open.feishu.cn/document/server-docs/docs/wiki-v2/space-node/move and https://open.feishu.cn/document/server-docs/docs/wiki-v2/space-node/copy?lang=zh-CN

ViperCai: Agreed. Official copy/move `131006` is container-edit (and space member/admin when the target is a space root), not node view.

Follow-up commit splits the hints:
- `wiki +node-get` / `+node-list`, and `+move`’s `get_node` resolve, keep the existing read-access text
- `wiki +node-copy`, `+move` write APIs, and docs-to-wiki use a write-specific hint: container edit on the relevant parent, plus space membership/admin for a space-root target

Still not branching on the unstable `131006` message text.

shortcuts/wiki/wiki_helpers.go
```diff
@@ -42,10 +42,39 @@ func appendWikiProblemHint(err error, hint string) error {
...
 }
+
+// wikiWritePermissionDeniedHint provides stable recovery for write-path 131006
+// (node-copy / move). Official copy/move APIs require container edit permission
+// on the relevant parent nodes; copying or moving to a space root can also
+// require wiki space membership or administrator permission.
+func wikiWritePermissionDeniedHint() string {
+	return "The current user or app/bot identity lacks Wiki container permission for this write. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error; ask the resource owner or wiki administrator to grant container edit permission on the relevant source or destination parent node. Copying or moving to a space root can also require wiki space membership or administrator permission. Use an accessible resource if that access cannot be granted."
```
fangshuyu-768: [P2] Keep write recovery specific to each move mode

This hint is shared by node copy, Wiki node move, and docs-to-wiki, but those 131006 cases do not share one resource-permission set. The official node-move contract requires edit permission on the moved node in addition to source and destination parent-container edit permission. For docs-to-wiki, 131006 can instead mean that the source Drive document lacks manage permission or its parent folder lacks edit permission. Following the current Wiki-container-only hint can therefore leave these cases unresolved even after the requested grant. Please use operation-specific hints, or include the moved-node and source-Drive requirements in the relevant paths.

References: https://open.feishu.cn/document/server-docs/docs/wiki-v2/space-node/move and https://open.feishu.cn/document/server-docs/docs/wiki-v2/task/get?lang=zh-CN

shortcuts/wiki/wiki_move.go
```diff
@@ -269,7 +269,7 @@ func (api wikiMoveAPI) MoveDocsToWiki(ctx context.Context, targetSpaceID string,
 		spec.DocsToWikiBody(),
 	)
 	if err != nil {
-		return nil, err
+		return nil, annotateWikiWritePermissionDenied(err)
```
fangshuyu-768: [P2] Treat 131006 from task polling as terminal

After this call returns a task_id, the GetMoveTask endpoint can itself return 131006. GetMoveTask currently returns that error without the new annotation, and pollWikiMoveTask unconditionally treats every poll error as transient, so the command makes all 30 attempts and finally tells the caller to retry status lookup. That contradicts the new terminal no-retry contract. Please annotate 131006 on the task-result path and short-circuit polling before the next attempt, with recovery appropriate to the documented task permission failure.

Reference: https://open.feishu.cn/document/server-docs/docs/wiki-v2/task/get?lang=zh-CN

shortcuts/wiki/wiki_helpers.go
```diff
@@ -42,10 +42,83 @@ func appendWikiProblemHint(err error, hint string) error {
...
 func wikiPermissionDeniedHint() string {
 	return "The current user or app/bot identity lacks access to the target wiki space or node. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error; ask the resource owner or wiki administrator to grant read access, or use an accessible resource."
 }
+
+// wikiCopyPermissionDeniedHint recovers copy-path 131006. Official copy
+// requires container edit on the source or destination parent; a space-root
+// target can also require wiki space membership or administrator permission.
+func wikiCopyPermissionDeniedHint() string {
+	return "The current user or app/bot identity lacks Wiki container permission for this copy. This is resource access, not app scope authorization. Do not retry the same request, reauthorize, or switch identity as trial and error; ask the resource owner or wiki administrator to grant container edit permission on the relevant source or destination parent node. Copying to a space root can also require wiki space membership or administrator permission. Use an accessible resource if that access cannot be granted."
```
fangshuyu-768, line 58: [P2] Include source-node access in copy recovery

The mounted regression for this path explicitly models 131006 as node permission denied, user needs read permission, but this hint only asks for container edit permission on source or destination parents, and the test rejects any read-access guidance. If the caller cannot read the source node, granting parent-container edit access does not make that source copyable, so the terminal error still directs the owner to a recovery that cannot fix the request. Because 131006 conflates source-node and destination-container ACL failures and this code intentionally does not branch on the unstable message text, please make the stable copy hint cover source-node read or copy access as well as destination-parent container edit, or keep it permission-neutral.

Reference: https://open.feishu.cn/document/server-docs/docs/wiki-v2/space-node/copy?lang=zh-CN

PR #2353, opened 2026-08-14
fix(im): make reaction scope conditional

### LINKED ISSUE #2352: [Bug] --no-reactions still requires im:message.reactions:read in +chat-messages-list
## Summary

`lark-cli im +chat-messages-list --no-reactions` still fails its scope pre-flight unless `im:message.reactions:read` is granted. With reactions explicitly disabled, the shortcut should not require the reactions scope.

## Environment

- lark-cli: 1.0.87 (also reproduced on 1.0.84)
- OS: macOS arm64
- identity: user
- relevant granted scopes: `im:message:readonly`, `im:chat:read`
- not granted: `im:message.reactions:read`

## Steps to reproduce

```bash
lark-cli im +chat-messages-list \
  --as user \
  --chat-id oc_xxx \
  --page-size 1 \
  --order desc \
  --no-reactions \
  --format json
```

## Actual behavior

The command exits before sending the message-list request:

```json
{
  "ok": false,
  "identity": "user",
  "error": {
    "type": "authorization",
    "subtype": "missing_scope",
    "message": "missing required scope(s): im:message.reactions:read",
    "missing_scopes": ["im:message.reactions:read"]
  }
}
```

It also suggests running `auth login` for the reactions scope.

## Control test

Calling the underlying endpoint through the raw API succeeds with the existing message-read permission and returns the requested message:

```bash
lark-cli api GET /open-apis/im/v1/messages \
  --as user \
  --params '{"container_id_type":"chat","container_id":"oc_xxx","page_size":1,"sort_type":"ByCreateTimeDesc"}'
```

## Expected behavior

When `--no-reactions` is present:

1. Skip reaction enrichment.
2. Do not pre-flight `im:message.reactions:read`.
3. Fetch the message list using the base message/chat read permissions.

## Impact

The current behavior creates a false authorization blocker and guides agents/users to request an unrelated scope even though the requested message history is readable.

This was previously mentioned as a side observation in #1031, but that issue tracks message-link generation rather than this scope/pre-flight bug.

logarithmone1128:
## Summary

Make the reaction scope conditional for message-pulling shortcuts so `--no-reactions` can fetch messages without requiring `im:message.reactions:read`. The default enrichment path still preflights the reaction scope before sending a request.

## Changes

- Move `im:message.reactions:read` from unconditional scopes to `ConditionalScopes` for the four message-pulling shortcuts.
- Enforce the conditional scope from validation only when reaction enrichment is enabled.
- Add command-level regression tests for both the opt-out and default paths, plus scope declaration coverage and updated documentation.
- Add a dry-run E2E under `tests/cli_e2e/im/` covering all four shortcuts.

## Test Plan

Run against `84f94143` as the base, after rebasing onto current `main`.

| Check | Result |
| --- | --- |
| `make unit-test` | pass, all packages |
| `make vet` | pass |
| `make fmt-check` | pass |
| `make build` | pass |
| `make quality-gate` | pass |
| `go mod tidy`, then `git diff --exit-code -- go.mod go.sum` | unchanged |
| `golangci-lint run --new-from-rev=<base>` | `0 issues` |
| `go run -C lint . --changed-from <base> ..` | pass |
| `go test -C lint ./... -count=1` | pass |
| `node scripts/skill-format-check/index.js` | pass |
| `go test ./tests/cli_e2e/im/ -run TestIM_ReactionEnrichmentDryRun` | pass, 8 subtests |

`make quality-gate` emits one pre-existing warning unrelated to this change: `im +feed-shortcut-list looks like a list command without an explicit default limit flag`. It is not introduced here — `main` raises the same `default_output` warning for `base +base-block-list`, `+form-questions-list` and `+record-search`.

### Revert-failing regression

`TestMessagePullShortcutsDeclareReactionScopeAsConditional` is the test that fails if this change is reverted. Verified by restoring the four shortcut files to the base commit and re-running it:

```
im_message_reaction_scope_test.go:32: ScopesForIdentity("bot") contains conditional reaction scope
FAIL	github.com/larksuite/cli/shortcuts/im
```

### What the dry-run E2E does and does not cover

`--no-reactions` already suppressed the `POST /open-apis/im/v1/messages/reactions/batch_query` call before this change, so the planned request set is identical with and without the fix, and a dry-run assertion cannot demonstrate the scope change. I confirmed that: the new test passes on the base commit too.

It is included because the conditional scope is only correct while enrichment really is a separate, suppressible request. If enrichment ever became unconditional, requiring `im:message.reactions:read` only when it runs would silently become wrong, and nothing at the command boundary asserted that premise. The test pins it for all four shortcuts: the default plan carries the enrichment POST, `--no-reactions` leaves the message read alone.

### Live E2E: not run, and why

Per the shortcut-change table this is a bug fix whose risk reaches the API boundary, so live E2E would normally be required. There is no deterministic, cleanable live flow for it.

The behaviour under test is which scopes are demanded, so a live assertion needs two tenant app configurations that differ only in whether `im:message.reactions:read` is granted, and it has to prove `--no-reactions` succeeds under the narrower one. Scope grants are tenant-level app configuration; a test cannot create or revoke them, so the flow is neither self-contained nor cleanable, and hard-coding two app credentials would leak tenant state and make the lane environment-dependent.

Fixture conditions that would make it runnable: a second tenant app credential in the live lane, provisioned without the reaction scope and with a stable chat fixture, exposed alongside `TEST_TENANT_ACCESS_TOKEN`. I have not added one.

Substitute evidence, all deterministic and offline:

- scope declarations asserted per shortcut per identity (`ScopesForIdentity`, `ConditionalScopesForIdentity`, `DeclaredScopesForIdentity`)
- validation asserted to demand the reaction scope only when enrichment is enabled, for both the opt-out and default paths
- the dry-run E2E above, pinning that `--no-reactions` removes the enrichment request

CI is authoritative. Live E2E is the one relevant check I did not run.

## Related Issues

- Fixes #2352

## Summary by CodeRabbit

* **Bug Fixes**
  * Reaction permissions are now requested only when reaction enrichment is enabled.
  * Commands using `--no-reactions` skip reaction loading and related permission checks.
  * Improved validation for message lists, searches, threads, pagination, and message retrieval.
  * Prevented unnecessary permission errors when reaction data is not requested.

* **Documentation**
  * Clarified when reaction permissions are required and when they can be omitted.

---

Rebased onto `84f94143` on 2026-08-27 so the branch is current; the patch is byte-identical to before the rebase (339 changed lines across 8 files) and every check above was re-run on the new base.

---

### On the empty checks column

`CI`, `PR Preview Package` and `Skill Format Check` are sitting at
`action_required` on this branch — GitHub holds workflow runs for an outside
contributor until a maintainer approves them, so no CI result has been produced
yet. The check matrix above is what I ran locally against the stated base; it is
not a substitute for CI, and I am not claiming CI passed. Happy to push again if
approving the runs surfaces anything.

### DIFF
--- a/shortcuts/im/im_chat_messages_list.go
+++ b/shortcuts/im/im_chat_messages_list.go
@@ -24,15 +24,16 @@ const (
 )
 
 var ImChatMessageList = common.Shortcut{
-	Service:     "im",
-	Command:     "+chat-messages-list",
-	Description: "List messages in a chat or P2P conversation; user/bot; accepts --chat-id or --user-id, resolves P2P chat_id, supports time range, --order asc/desc sorting, auto-pagination",
-	Risk:        "read",
-	Scopes:      []string{"im:message:readonly"},
-	UserScopes:  []string{"im:message.group_msg:get_as_user", "im:message.p2p_msg:get_as_user", "im:message.reactions:read"},
-	BotScopes:   []string{"im:message.group_msg", "im:message.p2p_msg:readonly", "im:message.reactions:read"},
-	AuthTypes:   []string{"user", "bot"},
-	HasFormat:   true,
+	Service:           "im",
+	Command:           "+chat-messages-list",
+	Description:       "List messages in a chat or P2P conversation; user/bot; accepts --chat-id or --user-id, resolves P2P chat_id, supports time range, --order asc/desc sorting, auto-pagination",
+	Risk:              "read",
+	Scopes:            []string{"im:message:readonly"},
+	UserScopes:        []string{"im:message.group_msg:get_as_user", "im:message.p2p_msg:get_as_user"},
+	BotScopes:         []string{"im:message.group_msg", "im:message.p2p_msg:readonly"},
+	ConditionalScopes: []string{messageReactionReadScope},
+	AuthTypes:         []string{"user", "bot"},
+	HasFormat:         true,
 	Flags: append([]common.Flag{
 		{Name: "chat-id", Desc: "(required, mutually exclusive with --user-id) chat ID (oc_xxx)"},
 		{Name: "user-id", Desc: "(required, mutually exclusive with --chat-id; user identity only) user open_id (ou_xxx)"},
@@ -115,7 +116,7 @@ var ImChatMessageList = common.Shortcut{
 		if _, err := buildChatMessageListRequest(runtime, chatId); err != nil {
 			return err
 		}
-		return nil
+		return ensureMessageReactionScope(runtime)
 	},
 	Execute: func(ctx context.Context, runtime *common.RuntimeContext) error {
 		if _, err := common.ValidatePageSizeTyped(runtime, "page-size", chatMessagesListDefaultPageSize, 1, chatMessagesListMaxPageSize); err != nil {
--- a/shortcuts/im/im_message_reaction_scope.go
+++ b/shortcuts/im/im_message_reaction_scope.go
@@ -1,0 +1,18 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package im
+
+import "github.com/larksuite/cli/shortcuts/common"
+
+const messageReactionReadScope = "im:message.reactions:read"
+
+// ensureMessageReactionScope keeps reaction enrichment opt-in at the scope
+// layer as well as the request layer. Pure builder tests construct partial
+// runtime contexts, so only run the token preflight for mounted commands.
+func ensureMessageReactionScope(runtime *common.RuntimeContext) error {
+	if runtime.Bool("no-reactions") || runtime.Factory == nil || runtime.Factory.Credential == nil || runtime.Config == nil {
+		return nil
+	}
+	return runtime.EnsureScopes([]string{messageReactionReadScope})
+}
--- a/shortcuts/im/im_message_reaction_scope_test.go
+++ b/shortcuts/im/im_message_reaction_scope_test.go
@@ -1,0 +1,98 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package im
+
+import (
+	"errors"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/larksuite/cli/errs"
+	"github.com/larksuite/cli/internal/cmdutil"
+	"github.com/larksuite/cli/internal/core"
+	"github.com/larksuite/cli/internal/credential"
+	"github.com/larksuite/cli/shortcuts/common"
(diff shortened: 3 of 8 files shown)


logarithmone1128:
Same situation as the checks column suggests: `CI`, `PR Preview Package` and `Skill Format Check` are held at `action_required` for an outside contributor, so no CI evidence exists on this branch.

I reproduced the CI job set locally, job by job as `.github/workflows/ci.yml` defines it, on `289b7adc` against base `84f94143`. Not a substitute for CI, and I am not claiming CI passed.

| CI job | Command as CI runs it | Result |
| --- | --- | --- |
| `fast-gate` | `go build ./...`, `go vet ./...`, gofmt check | pass |
| `unit-test` | `go test -race -count=1 ./cmd/... ./internal/... ./shortcuts/... ./extension/...` | pass, 130 packages, 0 failures |
| `lint` | `golangci-lint run --new-from-rev`, `go run -C lint . --changed-from` | `0 issues`, source guard pass |
| `script-test` | `make script-test` | pass |
| `deterministic-gate` | `make quality-gate` | pass, 0 rejects |
| `coverage` | `go test -race -coverprofile` over 154 packages, threshold 40% | pass, **82.5%**, 0 failures |
| `deadcode` | incremental `deadcode -test ./...`, HEAD vs base | **0 new**, 18 entries on both sides |
| `plugin-integration` | `go test ./tests/plugin_e2e/...` | pass |
| `sidecar-integration` | `make sidecar-test` | pass |
| `e2e-dry-run` | `node scripts/e2e_domains.js`, then the 22 resolved dry-run suites | pass, 0 failures |
| `security` | `go-licenses check` | pass |
| `license-header` | `.licenserc.yaml` | pass |

Notes so nothing is overstated:

- `make quality-gate` emits one `default_output` warning for `im +feed-shortcut-list`, a command this branch does not touch. `main` raises the same warning class for `base +base-block-list`, `+form-questions-list` and `+record-search`, so it is pre-existing.
- `govulncheck` reports five dependency and stdlib findings that predate this branch — `gorilla/websocket@v1.5.0`, `golang.org/x/image@v0.30.0`, and `net/url` / `crypto/tls` / `encoding/xml` in `go1.24.13`. The CI step is `continue-on-error: true`, so it does not gate.
- `e2e-live` is skipped for fork pull requests in CI, which matches the merged external PRs I looked at.
- On my first attempt the coverage job showed ten failures in `extension/credential/env` and `cmd`. That was my own shell exporting `LARKSUITE_CLI_APP_ID` / `_SECRET` / `_BRAND` for a neighbouring step; those tests assert the variables are *absent*. Re-run in a clean environment it is 0 failures at the same 82.5%. Mentioning it because the first number was wrong and the tests were right.

Ran on macOS/arm64 with `go1.24.13`, not CI's `ubuntu-latest`, so platform-specific behaviour is not covered.

PR #2374, opened 2026-08-18
feat(sheets): combine chart workflows and special chart types

zhengzhijiej-tech:
## Problem

Chart workflow improvements and bubble/waterfall/pareto support were developed on separate branches. Keeping them separate leaves the semantic shortcuts, batch paths, generated schemas, and bundled Sheets skill out of sync.

## Changes

- add semantic chart create/config/data-update shortcuts and dedicated chart batch entry points
- preserve partial chart updates, detached headers, cross-sheet ranges, color themes, and batch error recovery
- support bubble, waterfall, and pareto charts across examples, schemas, readback, and generated flags
- add semantic bubble role flags for key/x/y/group/size on create and data update
- sync the bundled `lark-sheets` skill from the combined sheet-skill-spec output, including the chart layout check script
- keep generic `+batch-update` compatible with chart operations while guiding pure chart batches to the dedicated shortcuts

## Validation

- `go test -count=1 ./shortcuts/sheets/...`
- `make unit-test`
- `make vet`
- `make fmt-check`
- `node scripts/skill-format-check/index.js skills`
- `make quality-gate QUALITY_GATE_CHANGED_FROM=origin/main`

## Summary by CodeRabbit

* **New Features**
  * Added shortcuts for creating charts, updating chart settings or data, and performing batch chart operations.
  * Added examples for bubble, waterfall, and pareto charts.
  * Added a chart layout checker for overlaps, overflow, and occupied-cell conflicts.
  * Added flexible chart aliases, validation, labels, axes, colors, ranges, and partial updates.
  * Added recursive schemas and improved guidance for chart commands.

* **Documentation**
  * Expanded chart and batch-operation guidance with examples, schemas, validation, and recovery instructions.
  * Clarified chart labels, colors, persistence, retries, and final-point label controls.


shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1725 @@
...
+		for _, color := range colors {
+			if !chartHexColorPattern.MatchString(color) {
+				return sheetsValidationForFlag("colors", "--colors contains invalid hex color %q", color)
+			}
+		}
+	}
+	return nil
+}
+
+func normalizedChartColors(rt flagView) []string {
```
xiongyuanwen-byted, line 1740: This feels unnecessary. The skill doc already says the engine loops a single color for all series, so appending a duplicate of the first color (lines 1715-1720: `len(colors)==1 → colors = append(colors, colors[0])`) just confuses agents reading the output — they see `["#FF0000", "#FF0000"]` and wonder why there are two of the same color. Consider returning the single-element slice as-is.

shortcuts/sheets/batch_op_dispatch.go
```diff
@@ -565,17 +578,15 @@ func translateBatchOp(raw interface{}, token string, index int) (map[string]inte
...
 	for userKey := range input {
 		normalized := strings.ReplaceAll(userKey, "-", "_")
 		for _, k := range reservedSubOpKeys {
 			if normalized == k {
-				return nil, sheetsValidationForFlag(
-					"operations",
-					"operations[%d] (%s): do not pass input.%s — it is already set from +batch-update top-level --url / --token",
-					index, sc, userKey,
-				)
+				delete(input, userKey)
```
xiongyuanwen-byted: This is a deliberate UX change (silently drop reserved locator keys instead of rejecting them), and the agent-friendly intent is well documented. However, consider adding a debug-level log or a note in the dry-run output when a sub-op's excel_id/spreadsheet_token/url is silently dropped, so the agent can detect when it accidentally passed a different token. Without any signal, the agent may think it's targeting spreadsheet A when the operation silently goes to spreadsheet B.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1725 @@
...
+		input["properties"].(map[string]interface{})["last_point_label"] = rt.Bool("last-point-label")
+	}
+	sheetSelectorForToolInput(input, sheetID, sheetName)
+	if err := validateInputAgainstSchema(rt, input); err != nil {
+		return nil, nil, err
+	}
+	return input, viewModel, nil
+}
+
+func chartDataUpdateInputFromSnapshot(
```
xiongyuanwen-byted, line 691: The dimension/index validation logic here (~150 lines: dim1Index resolution, dim2Indexes building, bubble role handling, chart-type-specific constraints for pie/pareto/combo/bubble, series count cap) is nearly identical to the same logic in chartCreateBasicInput (lines 285-511). Consider extracting the shared validation into a helper like `resolveChartDimensionIndexes(rt, chartType, dimensionCount, useBubbleRoles, bubbleSeries)` to avoid the two paths diverging when one gets updated and the other doesn't.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1527 @@
...
+		dataPointCount = first.rowCount
+	}
+	return normalized, dimensionCount, dataPointCount, nil
+}
+
+func splitChartDataRanges(value string) ([]string, error) {
+	var ranges []string
+	start := 0
+	inQuote := false
+	for i := 0; i <= len(value); i++ {
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Resolved: no code change is needed here. Go evaluates `||` left to right and short-circuits, so when `i == len(value)` is true, `value[i]` is not evaluated. The current loop therefore does not read out of bounds.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1680 @@
...
+	Execute: func(ctx context.Context, runtime *common.RuntimeContext) error {
+		token, err := resolveSpreadsheetTokenExec(runtime)
+		if err != nil {
+			return err
+		}
+		sheetID, sheetName, err := resolveSheetSelector(runtime)
+		if err != nil {
+			return err
+		}
+		input, err := chartConfigUpdateInput(runtime, token, sheetID, sheetName)
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Resolved in the current head: both pre-snapshot validation calls discard the unused return value while preserving validation order. The lint job is passing.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Resolved after clarifying the contract: custom colors now require at least two values. The single-color duplication was removed, one color is rejected with a typed validation error, and the synced Skill text states the same minimum.

skills/lark-sheets/references/lark-sheets-batch-update.md
```diff
@@ -2,30 +2,37 @@
...
-1. **目标 range 必须落在用户授权范围内**：除用户明示要修改的区域外，子操作禁止扩张到无关单元格 / 列 / Sheet。规划 range 时先确认每个子操作的边界。
-2. **批次完成后必须回读校验**：整个 `+batch-update` 执行成功后，用 `+csv-get` 或 `+cells-get` 抽样回读受影响区域，至少校验 3-5 个代表性单元格（首 / 中 / 末），与本地脚本预先计算的预期值对照。
-3. **预期条数前置断言**：涉及"批量填充 N 行"或"对 M 个区域分别写入"时，先把 N、M 硬编码进代码，回读后断言实际等于预期；不一致就再发一轮 `+batch-update` 补齐，禁止交付半成品。
+1. **目标 range 应落在用户授权范围内**：除用户明示要修改的区域外，子操作避免扩张到无关单元格 / 列 / Sheet。规划 range 时先确认每个子操作的边界。
+2. **批次完成后建议回读校验**：整个 `+batch-update` 执行成功后，用 `+csv-get` 或 `+cells-get` 抽样回读受影响区域，至少校验 3-5 个代表性单元格（首 / 中 / 末），与本地脚本预先计算的预期值对照。
+3. **预期条数前置断言**：涉及"批量填充 N 行"或"对 M 个区域分别写入"时，建议先把 N、M 硬编码进代码，回读后比较实际与预期；不一致就优先再发一轮 `+batch-update` 补齐，补不齐则在交付说明里列出缺口。
+4. **三条工具硬约束**：`--yes` 必带（high-risk-write，不带退出码 10）；单次 ≤100 条 operations，超出按批拆分；`+cells-batch-set-style` / `+cells-batch-clear` 等批量类 shortcut 不可嵌入 operations（它们本身就是批量原子操作，直接顶层调用）。
 
-若本次 `+batch-update` 的任一子操作写入了公式、复制了公式模板、或导入了含公式的数据块，**回读校验之后还必须继续执行 `+formula-verify`**。`+batch-update` 只保证"写入动作按序执行了"，不保证整批公式运行结果 zero-error。
+若本次 `+batch-update` 的任一子操作写入了公式、复制了公式模板、或导入了含公式的数据块，回读校验之外可继续执行 `+formula-verify` 做诊断。`+batch-update` 只保证"写入动作按序执行了"，不保证整批公式运行结果 zero-error。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-formula-translation.md
```diff
@@ -1,23 +1,32 @@
...
+## 翻译后建议：代码复现校验
 
-公式语法翻译完之后，**必须**用本地脚本在源数据上独立复现一份"等价计算结果"再写入。流程：
+公式语法翻译完之后，建议用本地脚本在源数据上独立复现一份"等价计算结果"再写入。流程：
 
 1. **挑 3-5 个代表性输入行**（首行 / 中段 / 末行 / 含空值 / 含异常格式各一）
 2. **用 Python 复现 Excel 原公式的语义**（不是飞书译文的语义，而是用户原本想要的结果）
 3. **写入飞书译文公式后回读这几行的实际值**
-4. **三方对照**：`Excel 原公式语义 == Python 复现 == 飞书译文回读值`，全部一致才交付；不一致先排查（数组语义？日期差？范围引用？）
+4. **三方对照**：`Excel 原公式语义 == Python 复现 == 飞书译文回读值`；不一致时优先排查（数组语义？日期差？范围引用？），无法修完时在交付说明标明风险。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-formula-translation.md
```diff
@@ -68,38 +81,39 @@
...
 
-`ARRAYFORMULA` `ARRAY_CONSTRAIN` `BYCOL` `BYROW` `CELL` `CHOOSECOLS` `CHOOSEROWS` `DROP` `EXPAND` `FILTER` `FLATTEN` `FREQUENCY` `GROWTH` `HSTACK` `IMPORTDATA` `IMPORTFEED` `IMPORTHTML` `IMPORTRANGE` `IMPORTXML` `LINEST` `LOGEST` `LOOKUP` `MAKEARRAY` `MAP` `MINVERSE` `MMULT` `MUNIT` `QUERY` `RANDARRAY` `REDUCE` `REGEXEXTRACT` `SCAN` `SEQUENCE` `SORT` `SORTBY` `SORTN` `SPLIT` `SUMPRODUCT` `SWITCH` `TAKE` `TEXTSPLIT` `TOCOL` `TOROW` `TRANSPOSE` `TREND` `UNIQUE` `VSTACK` `WRAPCOLS` `WRAPROWS` `XLOOKUP`
+`BYCOL` `BYROW` `MAKEARRAY` `MAP` `REDUCE` `SCAN` 同样是原生数组函数，但受核心原则二约束——导出 `.xlsx` 会丢高阶语义，默认改走 `ARRAYFORMULA` / 逐行填充 / 辅助列。
 
-> **注意：`SWITCH` 在飞书里被当作原生数组函数处理，这与 Excel 行为不同，不需要额外包 `ARRAYFORMULA`。**
+`ARRAYFORMULA` 不在上面这份清单里——它的作用是给**本来只按标量求值**的表达式套上数组语义，而不是自己返回数组。导出 `.xlsx` 时它会被翻译成 Excel 原生数组公式（`=ARRAYFORMULA(IF(A2:A6>2,B2:B6,""))` → `=IF(A2:A6>2,B2:B6,"")`，作用范围覆盖整片），语义完整保留，可安全使用。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-formula-translation.md
```diff
@@ -133,18 +142,25 @@
...
-同类场景也优先考虑 `MAP`：
+辅助列全是普通标量公式，导出 `.xlsx` 后逐格原样保留，也避开了 `LAMBDA` 系高阶函数的导出陷阱。
+
+**例外：结果要随筛选联动时，保持单条公式。** `SUBTOTAL` 的意义就在于筛选变化后重新计算，这类需求写成
+
+```excel
+=SUMPRODUCT(MAP(ARRAYFORMULA(ROW($E$16:$E$387)),LAMBDA(row,SUBTOTAL(103,INDIRECT("E"&row)))))
+```
+
+筛选状态本身导出 `.xlsx` 就不会保留，所以这个场景是飞书内专用，不受核心原则二的导出约束。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-formula-verify.md
```diff
@@ -45,33 +45,37 @@
...
-1. `status='success'` → 通过；可以把链路标完成。
-2. `status='partial'` → 扫描被内部上限截断。先缩小 `--range` 或拆 `--sheet-id` 续扫，**不允许**把 `partial` 当作 `success`。
-3. `status='errors_found'` 且 `compile_errors[]` 非空 → **先解决编译失败**：根据 `compile_errors[].reason` 修正公式语法（飞书函数名 / 范围语法 / 引用样式），用 `+cells-set` 重写后再调一次 `+formula-verify`。
-4. `status='errors_found'` 且只剩运行时错误 → 按 `error_summary` 的 `samples[].formula` + `depends_on` 排查根因（零除？空值参与运算？引用越界？日期差写法？数组语义？），修复后重新自检。
-5. 同一处错误连续修复 3 次仍未通过 → 改用 `IFERROR` 包裹兜底，或退回纯值写入；不要在 `errors_found` 状态下扩展 `+cells-set --copy-to-range`、追加批量写入。
+1. `status='success'` → 记录诊断通过。
+2. `status='partial'` → 扫描被内部上限截断。若该公式区是关键输出，可缩小 `--range` 或拆 `--sheet-id` 续扫；否则在交付说明中标明诊断覆盖不完整。
+3. `status='errors_found'` 且 `compile_errors[]` 非空 → 根据 `compile_errors[].reason` 修正公式语法（飞书函数名 / 范围语法 / 引用样式），或在成本过高时降级为静态值并说明原因。
+4. `status='errors_found'` 且只剩运行时错误 → 按 `error_summary` 的 `samples[].formula` + `depends_on` 排查根因（零除？空值参与运算？引用越界？日期差写法？数组语义？），优先修复关键输出区。
+5. 同一处错误连续修复 3 次仍未通过 → 改用 `IFERROR` 包裹兜底，或退回纯值写入，并在交付说明写清不随源数据更新。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-read-data.md
```diff
@@ -228,8 +228,9 @@ lark-cli sheets +csv-get --spreadsheet-token shtXXX --sheet-name "销售明细"
 - `annotated_csv` — 含 `[row=N]` 前缀的 CSV 主入口
 - `col_indices` / `row_indices` — 列字母 / 行号映射数组
 - `current_region` — 从锚点扩展到被空行空列包围的连续区域的 A1 范围。⚠️ **它不是整表真实边界**：遇表中部整行空行 / 整列空列会截断、可能小于真实数据范围；表尾的汇总 / 签名 / 脚注又可能让它大于纯数据范围。判断整表是否读全须拿 `+workbook-info` 的物理 `row_count` 当上界交叉核对（见上方「`row_count` 与 `current_region` 都不能单独定末行」）
+- `actual_range` — **本次实际读到的 A1 范围**。续读 / 校验覆盖度一律以它为准：`actual_range` 小于请求范围时，哪怕 `has_more=false` 也说明只拿到部分窗口，不能把 `row_count` 当成"已读全"
 - `row_count` / `col_count` — **本次返回的行 / 列数**（= `actual_range` 的尺寸，随 `--range` 变），**不是整表物理总行列数**；整表物理尺寸取 `+workbook-info`
-- `has_more` — 当前 `--range` 是否因 `--max-chars` 被截断（截断后续读接着用 `--range`）；它**只反映本次 range 内是否读完**，`has_more=false` **不代表整表已读全**（range 之外的数据不在判断内）
+- `has_more` — 当前 `--range` 是否因 `--max-chars` 被截断（截断后续读接着用 `--range`）；它**只反映本次 range 内是否还有后续页**，`has_more=false` **不代表整表或该窗口已读全**——仍要结合 `actual_range` 看实际覆盖到哪里
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-search-replace.md
```diff
@@ -1,11 +1,11 @@
...
 
-## 替换前 dry-run + 范围明确（替换前必做）
+## 替换前 dry-run + 范围明确（替换前建议）
 
 `+cells-replace` 的副作用是不可逆的（除非另写代码回滚）。执行前必须：
 
-1. **明确替换范围**：必须显式说明"只替换 X 列 / X 区域，还是全表替换"。**禁止**默认全表替换——容易误改无关列。范围应由用户指令决定，模糊时主动询问。
-2. **dry-run 命中数量**：先用 `+cells-search` 在同一范围、同一关键词、同一匹配选项（大小写 / 精确 / 正则）下统计命中数量。把数量和**期望命中数**（用户明示的或基于业务理解推断的）对照——一致才进入 `+cells-replace`，不一致先排查（关键词太宽？范围太大？）。
+1. **明确替换范围**：建议显式说明"只替换 X 列 / X 区域，还是全表替换"。避免默认全表替换——容易误改无关列。范围应由用户指令决定，模糊时主动询问。
+2. **dry-run 命中数量**：先用 `+cells-search` 在同一范围、同一关键词、同一匹配选项（大小写 / 精确 / 正则）下统计命中数量。把数量和**期望命中数**（用户明示的或基于业务理解推断的）对照；不一致先排查（关键词太宽？范围太大？）。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-sheet-structure.md
```diff
@@ -87,7 +87,7 @@ _公共四件套 · 系统：`--yes`、`--dry-run`_
 | Flag | Type | 必填 | 说明 |
 | --- | --- | --- | --- |
 | `--range` | string | xor | 要删除的行/列闭区间；行用 1-based 数字如 `3:7` 或单行 `5`，列用字母如 `C:F` 或单列 `C`。与 `--ranges` 二选一 |
-| `--ranges` | string + File + Stdin（简单 JSON） | xor | 要删除的多个行/列区间 JSON 数组（最多 100 个，如 `["5:5","8:8","11:13"]` 或 `["C:C","F:G"]`），全行或全列不可混用，区间不可重叠；与 `--range` 二选一。CLI 按位置**从大到小逆序**合成一次批量删除（fail-fast、不回滚）——正序删除会因前面的行/列被删导致后续索引前移错位，逆序由 CLI 代劳，无需自行排序 |
+| `--ranges` | string + File + Stdin（简单 JSON） | xor | 要删除的多个行/列区间 JSON 数组（最多 100 个，如 `["5:5","8:8","11:13"]` 或 `["C:C","F:G"]`），全行或全列不可混用，区间不可重叠；与 `--range` 二选一。CLI 按位置**从大到小逆序**合成一次批量删除（fail-fast，失败后先回读再补发）——正序删除会因前面的行/列被删导致后续索引前移错位，逆序由 CLI 代劳，无需自行排序 |
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-visual-standards.md
```diff
@@ -79,7 +81,7 @@
 ### 6. 图表展示
 
 - 遵循用户指令选择图表类型，或匹配用户意图（饼图/环形图 → 占比，折线图 → 趋势）。
-- 包含必要元素：标题、图例、数据标签、坐标轴标题。
+- 包含必要元素：标题、坐标轴标题、多系列图例；数据标签按需（关键点 / 末值标注即可，不必每个数据点都加）。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-write-cells.md
```diff
@@ -1,37 +1,37 @@
...
+1. **明确写入边界**：写入前建议能回答"目标 range 的起止行列号是多少？是否落在用户授权范围内？"。除用户明示要修改的区域外，避免扩张到原数据列以外或新建 Sheet。
+2. **完整性断言**：批量写入前建议把"预期写入条数"硬编码到代码里（如要填 106 条翻译 → `expected = 106`），写完后回读比较 `actual == expected`。少于预期时优先补齐，补不齐则在交付说明里列出缺口。
 3. **回读抽样校验**：写完关键值 / 公式后，用 `+csv-get` 或 `+cells-get` 重新读取写入区域，至少抽样 3-5 个代表性单元格（首 / 中 / 末），核对值与预期一致（与本地脚本计算的预期值对照）。公式特定的"先验证模板再 --copy-to-range / 修完再读回"细则见下方相关章节。
-4. **护原表 · 派生产物落点（写排名 / 标记 / 汇总 / 改写列时易丢数据）**：派生结果一律写到**真实末列 +1 的全新空列**或新建子表，**禁止复用任何已有原数据列**——哪怕该列看起来"空"，也要先 `+csv-get` 回读确认整列无原始数据再写。三条准则：① 不把新公式 / 新值写进原数据列（典型反例：把新算的排名公式写进了原本存放另一份原始数据的列，整列原始数据被覆盖丢失）；② 不改写、不合并原表头字段名（典型反例：把几个独立表头字段合并成一列，原字段名丢失）；③ 慎用 `--allow-overwrite`：它一旦让写入区盖到相邻原始列 / 行就是不可逆数据丢失，加它之前必须用 `+sheet-info` / `+csv-get` 核清目标 range 不含任何原始数据。
+4. **护原表 · 派生产物落点（写排名 / 标记 / 汇总 / 改写列时易丢数据）**：派生结果优先写到**真实末列 +1 的全新空列**或新建子表，避免复用任何已有原数据列——哪怕该列看起来"空"，也要先 `+csv-get` 回读确认整列无原始数据再写。三条准则：① 尽量不把新公式 / 新值写进原数据列（典型反例：把新算的排名公式写进了原本存放另一份原始数据的列，整列原始数据被覆盖丢失）；② 尽量不改写、不合并原表头字段名（典型反例：把几个独立表头字段合并成一列，原字段名丢失）；③ 慎用 `--allow-overwrite`：它一旦让写入区盖到相邻原始列 / 行就是不可逆数据丢失，加它之前建议用 `+sheet-info` / `+csv-get` 核清目标 range 不含任何原始数据。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-write-cells.md
```diff
@@ -61,11 +61,11 @@
...
+⚠️ `+csv-put` 可写值或公式：以 `=` 开头的单元格会被当作公式计算（读回时 `formula` 字段保留、`value` 为计算结果）。**公式内部含逗号 / 引号 / 换行时建议按 RFC 4180 转义**——含逗号的字段整格用双引号包裹、字段内部的引号再翻倍：如 `=COUNTIF(D5:D22,"及格")` 建议写成 `"=COUNTIF(D5:D22,""及格"")"`（外层双引号包裹整格，内部 `"及格"` 的引号翻倍成 `""及格""`）。漏转义会被 CSV 解析器按逗号拆列、整块写入区域错位（如本该 `G4:H6` 错成 `G4:K4`），详见下方 `+csv-put` 示例。**因此含逗号 / 引号 / 换行的公式优先改用 `+cells-set`（JSON 二维数组）写入——`cells[r][c].formula` 字段直接放公式串，零 CSV 转义负担，从根上避免拆列错位**（`+table-put` 的 typed 协议只接受 `columns / data / dtypes / formats` 四件套、没有 `formula` 字段，公式写入只能走 `+cells-set` / `+csv-put`）。此外 `+csv-put` **不会**携带样式/批注/图片，也无法把 `=` 开头的内容当字面量文本写入；需要样式/批注/图片用 `+cells-set`（或"写值 + 补样式"两步法）。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/references/lark-sheets-write-cells.md
```diff
@@ -167,11 +168,11 @@ Step 2: `+cells-set` — range="A2", cells 含 value + cell_styles + border_styl
 - **做法 A（推荐）**：按上方「内容与样式分离写入」两步法——先用模板单元格 + `--copy-to-range` 铺**完整样式**（`cell_styles` + `border_styles` 都要，不能只铺 border，否则新行字体 / 对齐 / 背景色全裸奔），再单独 `+cells-set` 写 value / formula。汇总行的 `cell_styles` 要点（bold / 背景色 / 上边框）见 `lark-sheets-visual-standards` 的「场景一 → 1A. 添加汇总行 / 表头行」。
 - **做法 B**：一次写入，但每个 cell（含空白格）都显式带 `cell_styles` + `border_styles`，**不能用 `{}`**。
 
-**判断是不是"新行"**：写入 range 超出 `+csv-get` 返回的 `current_region` 右 / 下边界（如 `current_region=A1:H10`、写 `A11:H11`）即新行，必须按上述做法补边框。
+**判断是不是"新行"**：写入 range 超出 `+csv-get` 返回的 `current_region` 右 / 下边界（如 `current_region=A1:H10`、写 `A11:H11`）即新行，建议按上述做法补边框。
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

skills/lark-sheets/SKILL.md
```diff
@@ -150,18 +162,18 @@ reference 分两组：先读**通用方法与规范**（横切所有任务的样
...
 | [Lark Sheet Read Data](references/lark-sheets-read-data.md) | 读取飞书表格中的单元格数据。当用户需要"看看数据"、"分析数据"、"统计/汇总"时使用；也适用于需要查看公式、样式、批注等详细信息的场景。 |
 | [Lark Sheet Search & Replace](references/lark-sheets-search-replace.md) | 在飞书表格中搜索和替换文本，支持限定范围、大小写匹配、精确匹配、正则表达式。当用户需要"查找"、"搜索"、"定位"某个值，或"替换"、"批量修改文本"、"把 A 改成 B"时使用。不要用于理解表格结构（应读取数据）、不要用于数据分析（应读取数据后计算）、不要把用户操作动作中的关键词（如"汇总金额""统计数量"）当作搜索词。 |
-| [Lark Sheet Write Cells](references/lark-sheets-write-cells.md) | 向飞书表格的指定区域批量写入值、公式、样式、批注或单元格图片。适用场景：填写数据、设置公式、修改格式、添加批注、嵌入单元格图片（如需操作浮动图片，请使用 lark-sheets-float-image）；若只需把一块 CSV 批量铺到表格上（值或公式，不带样式/批注），直接使用 `+csv-put` 更短更快。追加数据需先通过 lark-sheets-sheet-structure 插入行列。只要这次写入真实落了公式，收尾默认继续执行 `lark-sheets-formula-verify`。 |
+| [Lark Sheet Write Cells](references/lark-sheets-write-cells.md) | 向飞书表格的指定区域批量写入值、公式、样式、批注或单元格图片。适用场景：填写数据、设置公式、修改格式、添加批注、嵌入单元格图片（如需操作浮动图片，请使用 lark-sheets-float-image）；若只需把一块 CSV 批量铺到表格上（值或公式，不带样式/批注），直接使用 `+csv-put` 更短更快。追加数据需先通过 lark-sheets-sheet-structure 插入行列。写入公式后可使用 `lark-sheets-formula-verify` 做诊断。 |
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Thanks. This comes from historical Skill documentation changes brought in by the upstream spec sync, not from the chart optimization work in this PR. We are intentionally not addressing it in this PR; it should be tracked and handled separately with the broader documentation cleanup.

shortcuts/sheets/flag_view.go
```diff
@@ -134,6 +134,16 @@ func (m mapFlagView) lookupRawWithKey(name string) (string, interface{}, bool) {
...
+	for alias, target := range commandFlagAliases[m.command] {
+		if target != name {
+			continue
+		}
+		for _, key := range []string{alias, strings.ReplaceAll(alias, "-", "_")} {
+			if v, ok := m.raw[key]; ok {
+				return key, v, true
+			}
+		}
+	}
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Resolved in ae69ca18f. Alias candidates are now sorted before lookup, canonical names are normalized consistently for hyphenated and underscored forms, and a regression test covers deterministic selection.

skills/lark-sheets/references/lark-sheets-chart.md
```diff
@@ -2,18 +2,71 @@
...
 | 操作需求 | 使用工具 | 说明 |
 |---------|---------|------|
 | 查看已有图表 | `+chart-list` | 获取图表的类型、数据源和样式配置 |
-| 创建/更新/删除图表 | `+chart-{create|update|delete}` | 对图表对象执行写入操作 |
+| 按类型和范围创建基础图 | `+chart-create-basic` | 支持 column/bar/line/area/pie/scatter/combo/radar/bubble/waterfall/pareto、行/列方向与整图配色；无需构造 snapshot |
+| 更新标题、轴、图例、标签、堆叠、平滑或整图配色 | `+chart-config-update` | CLI 读取当前快照并只回写配置 patch |
+| 修正已有图表的数据范围或方向 | `+chart-data-update` | CLI 读取当前快照并只回写 data patch，保留其它配置 |
+| 批量创建多个独立图表 | `+batch-chart-create` | 保留成功图表，并逐项返回失败原因；只重试失败项 |
+| 批量更新多个独立图表 | `+batch-chart-update` | 逐图读取当前快照并生成 partial properties |
+| 高级创建/更新、删除图表 | `+chart-{create|update|delete}` | 按系列/数据点精细设置等高级需求才使用原始 properties；更新只提交必要的局部 properties |
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Resolved in ff55e679 and synced to ae69ca18f. The chart shortcut pipes are escaped, the legacy dimension signature is formatted as inline code, and the generated CLI documentation was refreshed.

skills/lark-sheets/scripts/lark_chart_layout_check.py
```diff
@@ -0,0 +1,458 @@
...
+
+def main() -> None:
+    args = parse_args()
+    locator = _locator(args.sheet_id)
+    try:
+        workbook_data = envelope_data(
+            run_sheets("+workbook-info", **locator, timeout=args.timeout)
+        )
+        sheets = resolve_target_sheets(workbook_data, sheet_id=args.worksheet_id)
+        sheets = [sheet for sheet in sheets if not bool(sheet.get("is_hidden"))]
```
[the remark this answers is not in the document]
zhengzhijiej-tech: Resolved in ff55e679 and synced to ae69ca18f. Hidden worksheets are filtered only for implicit all-sheet scans; an explicitly selected hidden worksheet is retained. A regression test covers this case.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Valid maintainability suggestion. We will handle the shared chart-dimension validation extraction in a follow-up refactor, rather than expanding this chart optimization PR with a high-risk structural change.

shortcuts/sheets/batch_op_dispatch.go
zhengzhijiej-tech: Resolved in 547c8796e. When a batch sub-operation includes excel_id, spreadsheet_token, or url, dry-run and execution now emit an indexed warning naming the ignored keys and stating that the top-level locator is authoritative. The warning is deterministic and covered by a regression test; sheets tests and go vet pass.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1722 @@
...
+
+	input := map[string]interface{}{"excel_id": token, "operation": "create", "basic_chart": basic}
+	sheetSelectorForToolInput(input, sheetID, sheetName)
+	if err := validateInputAgainstSchema(rt, input); err != nil {
+		return nil, err
+	}
+	return input, nil
+}
+
+func chartConfigUpdateInput(rt flagView, token, sheetID, sheetName string) (map[string]interface{}, error) {
```
xiongyuanwen-byted, line 518: is called twice in the  path: once inside  (for validation, result discarded) and again directly in  (for the actual patch). This means  (which already has the single-color duplication noted above) runs twice, and  runs twice — once with an empty current map and once with the real snapshot. Consider extracting just the flag validation from  into a separate  so the snapshot path doesn't need to re-run the full input builder.

xiongyuanwen-byted, line 518: In `chartConfigUpdateInputFromSnapshot`, `addChartSemanticConfig` is called twice: first inside `chartConfigUpdateInput` (whose result is discarded), then directly in this function for the actual patch. This means `normalizedChartColors` and `applyChartConfigPatch` both run twice — once against an empty map and once against the real snapshot. Consider extracting just the flag validation from `chartConfigUpdateInput` into a `validateChartConfigFlags` so the snapshot path can validate without re-building the full input.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -61,80 +63,581 @@ var BatchUpdate = common.Shortcut{
...
+		"Chart operations are supported, but prefer +batch-chart-create / +batch-chart-update for chart-only work because their contracts and partial-failure recovery are simpler.",
+	},
+}
+
+var chartCreateBatchDispatch = map[string]batchOpMapping{
+	"+chart-create-basic": {"manage_chart_object", chartCreateBasicInput},
+}
+
+var chartUpdateBatchDispatch = map[string]batchOpMapping{
+	"+chart-config-update": {"manage_chart_object", chartConfigUpdateInput},
```
xiongyuanwen-byted, line 125: `BatchChartCreate` and `BatchChartUpdate` call `translateBatchOpWithDispatch` which silently drops reserved locator keys (excel_id/spreadsheet_token/url), but unlike `BatchUpdate`, they do not call `batchWarnings` → `batchIgnoredLocatorNotes` to surface which keys were dropped. This means an agent that accidentally passes a different spreadsheet token in a batch chart sub-op will have no warning that the value was ignored. Consider calling `batchWarnings` in their Execute methods, or adding batchIgnoredLocatorNotes to their DryRun output.

shortcuts/sheets/lark_sheet_chart.go
xiongyuanwen-byted, line 518: `chartConfigUpdateInput` (used by `ChartConfigUpdate.DryRun`) calls `validateChartAxisBounds(rt, "x", true)` with `numericAxis=true`, which always passes the axis-type check. But the actual execute-time check in `validateChartConfigSnapshot` requires the existing X axis to be `valueType=linear`. This means `--dry-run` will pass for `--x-axis-min`/`--x-axis-max` on a chart with a categorical X axis, but the real execution will fail. Not a blocker since the dry-run cannot fetch the snapshot, but worth noting in the Tips or docs so agents know the dry-run cannot fully validate axis bounds.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -61,80 +63,581 @@ var BatchUpdate = common.Shortcut{
...
 		}
 		return nil
 	},
 	DryRun: func(ctx context.Context, runtime *common.RuntimeContext) *common.DryRunAPI {
 		token, _ := resolveSpreadsheetToken(runtime)
-		input, _ := batchUpdateInput(runtime, token)
-		dr := invokeToolDryRun(token, ToolKindWrite, "batch_update", input)
+		plan, _ := buildBatchUpdatePlan(runtime, token)
+		dr := invokeToolDryRun(token, ToolKindWrite, "batch_update", plan.input)
+		if batchContainsSemanticChartUpdate(runtime) {
```
xiongyuanwen-byted, line 76: The `warning_message` key is set twice in `BatchUpdate.DryRun`: first by `batchWarnings` (line 81-83) and then potentially overwritten by `batchNeedsDimInsertBeforeStyleWarning` (line 84-86). Both use `dr.Set("warning_message", ...)`, so the second call unconditionally replaces the first. If both `batchWarnings` returns non-empty AND the dim-insert-before-style warning fires, the locator/legacy freeze warnings are silently lost. Consider accumulating all warnings into a single string or using a different key for the dim-insert warning.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -61,80 +63,581 @@ var BatchUpdate = common.Shortcut{
...
+			dr.Set("preflight", "execution reads each target chart snapshot before building its partial properties patch")
+		}
+		if len(plan.localFailures) > 0 {
+			dr.Set("local_validation_failures", plan.localFailures)
+		}
 		if warnings := batchWarnings(runtime); len(warnings) > 0 {
 			dr.Set("warning_message", strings.Join(warnings, "\n"))
 		}
+		if batchNeedsDimInsertBeforeStyleWarning(runtime) {
```
xiongyuanwen-byted: `batchWarnings` already includes `batchNeedsDimInsertBeforeStyleWarning` at line 653, so the accumulated `warning_message` (which may contain locator notes, dim-insert warning, colliding-freeze notes, and legacy-freeze notes) is unconditionally overwritten by the separate `batchNeedsDimInsertBeforeStyleWarning` call here. The `dr.Set("warning_message", dimInsertBeforeStyleWarning)` replaces the joined warnings with just the dim-insert message, silently dropping all other warnings. The fix is to remove this separate check — it is already covered by `batchWarnings`.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1722 @@
...
+				"data-label-position",
+				"--data-label-position requires existing data labels or --data-labels in the same update",
+			)
+		}
+	}
+	if !rt.Changed("x-axis-min") && !rt.Changed("x-axis-max") {
+		return nil
+	}
+	plotArea := chartMap(snapshot["plotArea"])
+	xAxis := findChartAxisMap(plotArea, "x", "bottom")
```
xiongyuanwen-byted, line 1670: `findChartAxisMap` can return nil if the chart snapshot has no bottom X axis. Accessing `xAxis["valueType"]` on a nil map would panic. The `chartMap` helper at line 1170 guards against this by returning an empty map for nil, but `findChartAxisMap` at line 1185 returns nil directly. Consider calling `chartMap` on the result before the type assertion, or adding a nil check. For example, a chart with only a left Y axis (no bottom X axis) would crash here.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Valid maintainability observation, but the repeated builder calls are pure and do not change behavior. We will handle validation/build separation in a follow-up refactor rather than expand this PR with a structural change.

zhengzhijiej-tech: Valid maintainability observation, but the repeated builder calls are pure and do not change behavior. We will handle validation/build separation in a follow-up refactor rather than expand this PR with a structural change.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: No panic occurs here. Reading from a nil Go map is safe and returns the zero value; the subsequent type assertion uses the comma-ok form, so a missing bottom X axis falls through to the typed validation error. No code change is needed.

shortcuts/sheets/lark_sheet_batch_update.go
zhengzhijiej-tech: Fixed in 562591b4f. Both dedicated chart batch shortcuts now expose ignored locator keys in dry-run output and stderr during execution; flat create operations are normalized before checking.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Fixed in 562591b4f. The shortcut Tips now state that dry-run validates request shape only and that execution may reject X-axis bounds unless the existing bottom X axis is continuous (`valueType=linear`).

shortcuts/sheets/lark_sheet_batch_update.go
zhengzhijiej-tech: Fixed in 562591b4f. The duplicate overwrite was removed, so the single aggregated warning message preserves locator, dim-insert, and freeze warnings. A regression test covers the combined locator plus dim-insert case.

shortcuts/sheets/lark_sheet_batch_update.go
zhengzhijiej-tech: Fixed in 562591b4f by removing the redundant `dr.Set` call. `batchWarnings` is now the only source of the dry-run warning message, preserving all accumulated warnings.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -80,61 +88,565 @@ var BatchUpdate = common.Shortcut{
...
+			prepared = append(prepared, translated[remoteIndex])
+			preparedIndexes = append(preparedIndexes, rawIndex)
+			preparedNormalized = append(preparedNormalized, normalized)
+			continue
+		}
+		fv := newMapFlagViewForCommand(shortcut, normalized.input)
+		sheetID := strings.TrimSpace(fv.Str("sheet-id"))
+		sheetName := strings.TrimSpace(fv.Str("sheet-name"))
+		chartID := strings.TrimSpace(fv.Str("chart-id"))
+		snapshot, err := fetchChartSnapshot(ctx, runtime, token, sheetID, sheetName, chartID)
```
fangshuyu-768, line 410: [P1] Prevent stale patches when a chart appears twice in one batch

All semantic chart updates are preflighted before any write occurs. If two operations target the same chart, both patches are built from the same old snapshot. Because arrays such as `axes` are replacement fields, the later patch can restore values changed by the earlier operation while both report success. I reproduced this with an X-axis title update followed by a Y-axis title update: the second payload contains the old X-axis title. Please reject duplicate `(sheet, chart_id)` targets or fold/apply their patches sequentially.

skills/lark-sheets/references/lark-sheets-batch-update.md
```diff
@@ -2,30 +2,37 @@
...
+
 **不可放进 `--operations` 的写 shortcut**（`shortcut` 枚举不含它们，强行写入会被校验拒）：`+cells-set-image`（需本地上传图片）、`+styles-put` / `+dropdown-update` / `+dropdown-delete` / `+cells-batch-clear`（自身已是批量入口，不可再嵌套）、`+dim-move`。这些操作需在 `+batch-update` 之外单独调用。
 
 **行高列宽批量不走这里**：多行 / 多列不同尺寸用 `+styles-put` 的 `row_sizes` / `col_sizes`（可与样式同批），或 `+rows-resize --heights` / `+cols-resize --widths` 的 map 形态（见 `lark-sheets-range-operations`）；map 形态不可作为 `--operations` 子操作嵌入（子操作里仍可用单区间形态 `range` + `height`/`width`）。
 
-**执行语义（fail-fast，不回滚）**：默认首个失败的子操作即中断剩余操作，但**已执行成功的子操作不回滚**——服务端报 "N succeeded, M failed" 时前 N 个已实际生效。修复失败项后**只重发失败起的剩余子集**，整批重发会把已成功的操作（如插行）重复应用。传 `--continue-on-error` 则遇失败仍继续执行剩余操作。正因如此，含结构变更（插删行列 / 移动）的批次失败后要先回读确认现状再续发。
+**执行语义（fail-fast；失败后哪些已生效取决于批次构成）**：默认首个失败的子操作即中断剩余操作。此前的子操作**是否已落盘不统一**：纯单元格 / 行列结构类写入在提交前只累计在内存，失败时整体不落盘（等效回滚）；而图表 / 透视表等对象类子操作执行时会**先把此前累计的写入提交落盘再创建对象**——批次含这类子操作时，失败前完成的部分（含其之前的普通写入）已实际生效、无法回滚。因此失败后**不要假设"全部回滚"或"全部保留"**：先看返回 `results` 里各子操作的状态，再回读现状（行列数 / 目标格 / `+chart-list` 等对象清单）确认已生效集合，只补发未生效部分——盲目整批重发会重复应用已生效操作（如插行 / 建图），盲目只发失败尾可能写到未生效的旧结构上。传 `--continue-on-error` 则遇失败仍继续执行剩余操作，已成功部分保留（返回 "N succeeded, M failed"）。
```
fangshuyu-768, line 30: [P1] Keep one authoritative batch recovery contract

This new recovery contract contradicts `BatchUpdate.Description`, its Tips, and the source comment, which still say that every succeeded sub-operation remains applied and instruct callers to resend only the tail. After a failure, these opposing instructions can either duplicate structural writes or skip rolled-back operations. Please verify the server behavior and synchronize the CLI metadata, tests, and Skill guidance around one authoritative contract.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1725 @@
...
+		if flag == "x-axis-label-angle" || flag == "y-axis-label-angle" {
+			out[key] = rt.Int(flag)
+		} else if flag == "x-axis-min" || flag == "x-axis-max" || flag == "y-axis-min" || flag == "y-axis-max" {
+			out[key] = rt.Float64(flag)
+		} else {
+			out[key] = rt.Str(flag)
+		}
+	}
+	if rt.Changed("stacked") {
+		out["stack"] = "normal"
```
fangshuyu-768: [P2] Preserve the explicit false value of --stacked

`Changed("stacked")` is checked without reading the boolean value, so `--stacked=false` still emits `stack=normal` and enables stacking. This is the opposite of the caller explicit value. Please map `true` to `normal` and `false` to `none`, or reject the false form explicitly if the alias is intended to be presence-only.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1725 @@
...
+		"properties": map[string]interface{}{
+			"snapshot": patch,
+		},
+	}
+	sheetSelectorForToolInput(input, sheetID, sheetName)
+	if err := validateInputAgainstSchema(rt, input); err != nil {
+		return nil, nil, nil, "", err
+	}
+	notice := ""
+	if normalized != dataRange {
```
fangshuyu-768: [P2] Report range merging only when a merge occurred

`normalized != dataRange` does not necessarily mean ranges were merged; trimming whitespace also changes the string. For example, `A1:B10, D1:E10` remains two aligned ranges, but the result claims they were merged into an enclosing rectangle. Please return an explicit `merged` signal from normalization, or compare the original and normalized range counts before emitting this notice.

skills/lark-sheets/references/lark-sheets-batch-update.md
zhengzhijiej-tech: Thanks. This is a historical, non-chart batch recovery contract issue surfaced by the upstream Skill sync in this branch, rather than part of the chart-specific implementation. We are intentionally not changing the general batch recovery behavior or metadata in this PR and will handle that contract unification separately.

shortcuts/sheets/lark_sheet_batch_update.go
zhengzhijiej-tech: Fixed in 73333869c. Both `+batch-update` and `+batch-chart-update` now reject duplicate semantic chart targets before any snapshot read or write, with an actionable error instructing callers to split them into separate batch calls.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Fixed in 73333869c. The compatibility alias now maps `--stacked=true` to `stack=normal` and `--stacked=false` to `stack=none`, with standalone and batch regression coverage.

shortcuts/sheets/lark_sheet_chart.go
zhengzhijiej-tech: Fixed in 73333869c. Range normalization now returns an explicit `merged` signal, so whitespace-only normalization no longer emits the enclosing-range notice. Regression tests cover aligned ranges with spaces and a genuine unaligned merge.

shortcuts/sheets/lark_sheet_chart.go
```diff
@@ -0,0 +1,1745 @@
...
+	if inQuote {
+		return nil, common.ValidationErrorf("unterminated quoted sheet name")
+	}
+	return ranges, nil
+}
+
+func parseChartDataRange(value string) (chartDataRange, error) {
+	item := chartDataRange{}
+	ref := strings.TrimSpace(value)
+	if bang := strings.LastIndex(ref, "!"); bang >= 0 {
```
fangshuyu-768: [P2] Normalize sheet qualifiers before grouping chart ranges

`item.sheet` retains the raw qualifier spelling, so `Data!A1:B10` and `'Data'!B1:C10` are grouped as different sheets even though they identify the same worksheet. Their overlap therefore bypasses the merge path: dry-run keeps two overlapping refs, while quoting both ranges correctly produces `'Data'!A1:C10`. This also makes dimension, header-count, and index validation depend on equivalent syntax. Please use the existing sheet-qualifier parser to derive a canonical name for comparisons while preserving the original spelling for rendering, and add a mixed quoted/unquoted regression test.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -80,61 +89,613 @@ var BatchUpdate = common.Shortcut{
...
+			continue
+		}
+		fv := newMapFlagViewForCommand(operation.shortcut, operation.input)
+		chartID := strings.TrimSpace(fv.Str("chart-id"))
+		sheetID := strings.TrimSpace(fv.Str("sheet-id"))
+		sheetName := strings.TrimSpace(fv.Str("sheet-name"))
+		if chartID == "" || (sheetID == "" && sheetName == "") {
+			continue // The operation translator reports missing required fields.
+		}
+		selectorKey := "id:" + sheetID
```
fangshuyu-768, line 585: [P1] Canonicalize sheet selectors before checking duplicate chart targets

This key treats `sheet_id` and `sheet_name` as different namespaces. A batch can target the same worksheet and chart once by ID and once by name, bypassing this guard, so both patches are still built from the same pre-write snapshot. The later patch can restore fields changed by the earlier operation. Please resolve both selectors to one canonical sheet identity before preflighting the updates, and add a regression test covering ID/name aliases.

xiongyuanwen-byted (approved):
LGTM

PR #2389, opened 2026-08-18
feat(sheets): surface a before_you_edit notice on sheet-mutating commands

anunwu-byted:
## Why

Agents driving `lark-cli sheets` routinely regress on style inheritance, Lark-vs-Excel formula semantics and row heights because the governing skill reference never enters their context. Prose in SKILL.md only reaches sessions that read it (and re-read it on follow-up turns); the command output is the one channel every session parses.

Measured on a 661-session evaluation batch: 81.7% of sessions invoke lark-cli, and the import/write/read commands decorated here cover 80.8% — including 100% of the multi-turn edit workload where document-side guidance never lands.

## What

- `internal/beforeyouedit`: process-level notice slot, mirroring `internal/deprecation`. Registers with the output package from `init()`, so it also surfaces in embedder distributions that assemble the command tree via `cmd.Build` without running the root entry point's `setupNotices` (where `output.PendingNotice` is normally wired).
- `internal/output`: `RegisterBuiltinNotice` + `GetNotice` now merges the wired `PendingNotice` hook with builtin providers.
- `shortcuts/register.go`: `wrapSheetsBeforeYouEdit` decorates 9 sheet commands with per-intent guidance (same Validate/Execute/OnInvoke decoration pattern as `wrapSheetsBackwardDeprecation`; pre-existing OnInvoke hooks are chained). Messages point at `lark-cli skills read lark-sheets references/...` — the embedded skill content, version-matched to the binary by `internal/skillscheck`.

## Verification

- `go test ./internal/beforeyouedit ./internal/output ./shortcuts` — pass (re-run on top of current main 679ebd52)
- `go vet` / `gofmt` clean
- Real binary dry-run shows the notice in the envelope: `lark-cli sheets +workbook-info --spreadsheet-token x --dry-run` → `"_notice": {"before_you_edit": {"command": "+workbook-info", "message": "..."}}`
- A unit test pins the `PendingNotice == nil` (embedder) path and the merge path.

## Notes

- JSON/jq/raw/partial-failure and the stderr error envelope all carry `_notice` already; `--format pretty/table/csv/ndjson` never carried it and still does not.
- No envelope schema lock exists; e2e assertions are key-based — additive field.

## Summary by CodeRabbit

* **New Features**
  * Added before-you-edit notices for sheet shortcut commands.
  * Notices now appear in JSON output with the relevant command and message.
  * Notices are supported across validation, execution, invocation, and dry-run flows.
  * Multiple notices can be combined without replacing existing entries.

* **Bug Fixes**
  * Ensured notices remain available when validation fails or no explicit pending notice is provided.

### DIFF
--- a/internal/beforeyouedit/notice.go
+++ b/internal/beforeyouedit/notice.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+// Package beforeyouedit carries a process-level "read this before you edit"
+// pointer for sheet-mutating commands. The notice is surfaced in JSON output
+// envelopes under "_notice.before_you_edit", mirroring internal/deprecation.
+//
+// Unlike the update/skills notices wired in cmd/root.go, this one registers
+// itself with the output package directly (see init), so it also surfaces in
+// distributions that assemble the command tree via cmd.Build without running
+// the root entry point's setupNotices — e.g. embedder binaries.
+//
+// A CLI process runs exactly one shortcut, so a single process-level slot is
+// sufficient: the command's hooks record the notice before producing output,
+// and the output layer reads it back when building the envelope.
+package beforeyouedit
+
+import (
+	"sync/atomic"
+
+	"github.com/larksuite/cli/internal/output"
+)
+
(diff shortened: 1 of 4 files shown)


anunwu-byted:
Withdrawing for internal review; will resubmit.

PR #2406, opened 2026-08-19
fix(auth): preserve stored token read errors

### LINKED ISSUE #1925: GetStoredToken swallows storage/decryption errors and reports credentials as missing
## Summary

`GetStoredToken` currently converts every credential-storage or decoding error
into a `nil` token. As a result, a stored user credential can be reported as
`missing` when the actual problem is that the process cannot access the macOS
Keychain-backed master key, decryption failed, or the stored JSON is malformed.

In a headless/background process this surfaced as:

- `lark-cli auth status --json --verify` reporting the user identity as
  `missing`
- user-authenticated commands failing with `token_missing`

The encrypted credential file still existed and was valid. The credential was
not absent; its master key was unavailable to that process.

## Environment

- macOS 12.7.6, Intel x64
- Observed with `@larksuite/cli` 1.0.70
- The relevant implementation is unchanged in the current v1.0.71 release
- Reproduced in fresh non-GUI/background processes, including a `launchd`
  process and agent/automation runners

## Reproduction context

These steps describe the original failure. The affected machine has since
applied the supported workaround below and was not reverted solely to reproduce
the failure again.

1. Log in interactively using the default macOS Keychain-backed credential
   storage.
2. Confirm that `lark-cli auth status --verify` reports the user as ready.
3. Confirm that the encrypted credential file exists under
   `~/Library/Application Support/lark-cli/` with restrictive permissions.
4. Start a fresh headless/background process in a context that cannot access
   the macOS Keychain.
5. Run:

   ```bash
   lark-cli auth status --json --verify
   ```

   or another command requiring user authentication.

## Actual behavior

The CLI reports the stored user credential as missing and downstream commands
return `token_missing`.

## Expected behavior

The CLI should distinguish at least these cases:

- No stored credential exists
- macOS Keychain is unavailable or access is denied
- Credential decryption fails
- Stored credential JSON is malformed or corrupted

Only the first case should be reported as `missing`.

## Root cause

In v1.0.71, [`GetStoredToken`](https://github.com/larksuite/cli/blob/v1.0.71/internal/auth/token_store.go#L41-L52)
returns `nil` for both any `keychain.Get` error and any `json.Unmarshal` error:

```go
jsonStr, err := keychain.Get(keychain.LarkCliService, accountKey(appId, userOpenId))
if err != nil || jsonStr == "" {
    return nil
}
var token StoredUAToken
if err := json.Unmarshal([]byte(jsonStr), &token); err != nil {
    return nil
}
```

This collapses not-found, operational, decryption, and corruption errors into
the same downstream `missing` result.

Issue #1038 and PR #1085 already cover the underlying macOS automation/Keychain
compatibility problem and provide `config keychain-downgrade`. This issue is
specifically about preserving and reporting the real storage error instead of
misclassifying it as a missing credential.

## Suggested fix

- Return a distinct not-found result only when the credential truly does not
  exist.
- Propagate or classify Keychain, decryption, and JSON-decoding errors.
- Do not map non-missing storage errors to `user.status=missing` or
  `token_missing`.
- Include the existing recovery hint for affected headless environments:

  ```bash
  lark-cli config keychain-downgrade
  ```

- Never include tokens, decrypted credential content, or key material in logs
  or error messages.

## Suggested tests

Add automated coverage for:

- Credential file absent: reports `missing`
- Keychain retrieval denied/unavailable: reports a Keychain/storage error
- Decryption failure: reports a decryption/corruption error
- Malformed stored JSON: reports a decoding/corruption error
- Headless/background execution with the supported file-backed master-key mode
- Verification that errors and logs never expose token or key material

## Workaround verified

Running the supported command below moves master-key access to the file-backed
mode:

```bash
lark-cli config keychain-downgrade
```

After applying it, authentication succeeded from multiple fresh background
processes and a `launchd` process, and user-authenticated API commands worked
without another OAuth login.

green3sf:
## Summary

Preserve stored user-token read failures so inaccessible or corrupt credentials are no longer reported as missing. This keeps the existing typed storage error and keychain recovery hint visible to commands and diagnostics.

## Changes

- Make `internal/auth.GetStoredToken` return both the stored token and any storage or decoding error, then migrate callers to handle the distinction explicitly.
- Propagate failures through automatic identity selection, auth/profile commands, mail scope validation, and the sidecar status endpoint.
- Report `storage_error` from identity diagnostics and avoid suggesting a new login when credential storage is the actual problem.
- Keep login replacement and logout cleanup recoverable when an old token is corrupt, while surfacing a warning when revocation cannot read the token.
- Add regression coverage for absent versus corrupt credentials, typed error propagation, false `no_token` prevention, recovery guidance, and sensitive-value redaction.

## Test Plan

- [x] `gofmt` reports no unformatted changed files
- [x] `git diff --check`
- [ ] Unit tests pass (GitHub Actions; project dependencies were intentionally not downloaded locally)
- [ ] Manual local verification confirms the `lark-cli auth status --json --verify` flow works as expected

## Related Issues

- Fixes #1925

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication commands now distinguish missing credentials from unreadable or corrupted credential storage.
  * Invalid or mismatched stored credentials are rejected safely without exposing sensitive token data.
  * Storage and token retrieval errors are surfaced consistently instead of being treated as missing tokens.
  * Logout safely skips revocation when credentials cannot be read and provides a warning.
  * Profile, mail shortcuts, identity diagnostics, and status checks now report credential-storage issues accurately.
  * The multi-tenant status endpoint returns an appropriate server error when credential lookup fails.

### DIFF
--- a/cmd/auth/check.go
+++ b/cmd/auth/check.go
@@ -74,7 +74,10 @@ func authCheckRunWithRecovery(opts *CheckOptions, projector *recovery.Projector)
 		return output.ErrBare(1)
 	}
 
-	stored := larkauth.GetStoredToken(config.AppID, config.UserOpenId)
+	stored, err := larkauth.GetStoredToken(config.AppID, config.UserOpenId)
+	if err != nil {
+		return err
+	}
 	if stored == nil {
 		output.PrintJson(f.IOStreams.Out, map[string]interface{}{"ok": false, "error": "no_token", "missing": required})
 		return output.ErrBare(1)
--- a/cmd/auth/check_test.go
+++ b/cmd/auth/check_test.go
@@ -9,9 +9,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/larksuite/cli/errs"
 	larkauth "github.com/larksuite/cli/internal/auth"
 	"github.com/larksuite/cli/internal/cmdutil"
 	"github.com/larksuite/cli/internal/core"
+	"github.com/larksuite/cli/internal/keychain"
 	"github.com/larksuite/cli/internal/output"
 	"github.com/larksuite/cli/internal/recovery"
 	"github.com/larksuite/cli/internal/surface"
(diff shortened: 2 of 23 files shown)


green3sf:
Closing this PR as I do not plan to continue working on this change for now. Thank you for your time.

PR #2441, opened 2026-08-21
feat(slides): use office_slide_file parent_type for imported office presentations

R0bynZhu:
Summary

Image uploads to a presentation hard-coded parent_type=slide_file at every entry point, but imported "office" presentations require parent_type=office_slide_file — the drive backend rejects the native value for them. This mirrors the office_sheet_file rule the sheets domain already applies (shortcuts/sheets/helpers.go); the token shapes are identical, because an imported office file is an imported office file whether it backs a spreadsheet or a deck.

An imported office presentation carries either a legacy synthetic token prefix (fake_office_ / local_office_) or a 28-character token whose interleaved product/region marker is OFL0X.

Changes
Replace the slidesMediaParentType const with slidesMediaParentType(token), backed by isOfficePresentation(token); both parent_type values stay named constants.
Route both call sites through that one helper, so the rule lives in a single place:
uploadSlidesMedia — the Execute path shared by +media-upload and the  placeholder pipeline behind +create / +add-slide / +update-slide.
appendSlidesUploadDryRun — so every preview shows the same value Execute will send.
As in sheets, the rule stays inside the slides domain rather than moving into common.UploadDriveMediaAllTyped, which mail/doc/drive/base/calendar also share.

Known gap, documented at the helper: when --presentation is a wiki URL, the dry-run only has a "<resolved_slides_token>" placeholder — the real token needs a get_node call the preview must not make — so such a preview shows slide_file regardless. Execute is unaffected; it resolves first. Sheets has the identical limitation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Media uploads now correctly identify imported Office presentations and use the appropriate upload type.
  - Native presentations continue to use the standard upload type.
  - Dry-run previews now accurately reflect upload handling for both presentation formats.
  - Image uploads through slide creation and updates use the correct presentation context.

- **Tests**
  - Added coverage for direct uploads and image placeholders across native and imported Office presentations.
  - Expanded dry-run end-to-end validation for upload requests and ordering.


shortcuts/slides/slides_media_upload.go
```diff
@@ -132,18 +186,25 @@ func uploadSlidesMedia(runtime *common.RuntimeContext, filePath, fileName string
...
+// token only exists after a get_node call the dry-run must not make. Such a
+// preview shows slide_file even if the resolved deck turns out to be an
+// imported office one.
 func appendSlidesUploadDryRun(d *common.DryRunAPI, filePath, parentNode string, step int) {
 	d.POST("/open-apis/drive/v1/medias/upload_all").
 		Desc(fmt.Sprintf("[%d] Upload local file (max 20 MB)", step)).
 		Body(map[string]interface{}{
 			"file_name":   filepath.Base(filePath),
-			"parent_type": slidesMediaParentType,
+			"parent_type": slidesMediaParentType(parentNode),
```
fangshuyu-768, line 207: [P2] Keep Wiki dry-runs honest about the unresolved parent type

When --presentation is a Wiki URL whose node resolves to an imported Office deck, parentNode is the unresolved placeholder here, so slidesMediaParentType returns slide_file. Execute resolves the node first and sends office_slide_file. That makes the dry-run preview describe a different request precisely on the new branch. Please emit an unresolved or conditional parent_type until the token is known, rather than a concrete native value.

shortcuts/slides/slides_media_upload.go
```diff
@@ -7,21 +7,74 @@ import (
...
+// already applies (shortcuts/sheets/helpers.go). The token shapes are the same
+// there: an imported office file is an imported office file whether it backs a
+// spreadsheet or a deck.
+//
+// NOTE: neither value is accepted by the multipart upload_prepare endpoint
+// (99992402 field validation failed), so slides image uploads stay capped at
+// 20 MB regardless of which one applies.
+const (
+	slideFileParentType       = "slide_file"
+	officeSlideFileParentType = "office_slide_file"
```
fangshuyu-768, line 36: [P3] Update the shipped Slides guidance with the new parent type

This changes the command contract, but skills/lark-slides/references/cli/lark-slides-media-upload.md still says uploads are fixed to slide_file and that it is the only accepted value. The add-slide, update-slide, and workflow/error-handling references repeat the same rule. These files are shipped agent guidance, so after this merge they will contradict runtime behavior for imported Office decks and can drive incorrect diagnosis or manual requests. Please update all four references to describe the token-dependent slide_file versus office_slide_file selection.

shortcuts/slides/slides_media_upload.go
```diff
@@ -7,21 +7,74 @@ import (
...
+// presentation may carry.
+var officePrefixes = []string{fakeOfficePrefix, localOfficePrefix}
+
+func isOfficePresentation(presentationToken string) bool {
+	for _, prefix := range officePrefixes {
+		if strings.HasPrefix(presentationToken, prefix) {
+			return true
+		}
+	}
+	if len(presentationToken) != 28 {
```
ethan-zhx, line 51: 这里改为 < 28

ethan-zhx (approved):

PR #2446, opened 2026-08-21
feat(slides): add media download shortcut

ethan-zhx:
## Summary

- Add `slides +media-download` shortcut to download Slides image media by `file_token`.
- Support direct Drive media download first, then automatically fall back to source-file preview download when direct download is forbidden.
- Save downloaded media through runtime FileIO with safe relative output handling, auto extension detection, and duplicate filename avoidance.
- Update `lark-slides` skill guidance to route image download scenarios to the new shortcut.

## Tests

- Added unit coverage for direct download, forbidden fallback, non-permission error behavior, and dry-run output.
- Not run locally: `go test ./shortcuts/slides` failed because `go` is not available in the current shell (`go: command not found`).

## Summary by CodeRabbit

* **New Features**
  * Added a Slides media download shortcut using a file token.
  * Supports custom output paths, dry-run previews, safe filenames, duplicate handling, and automatic extension correction.
  * Falls back to a source-file preview when direct media access is denied.
  * Reports the downloaded file path, size, content type, source, and output adjustments.

* **Documentation**
  * Added usage details to the Slides shortcuts reference.

* **Tests**
  * Expanded end-to-end coverage for Slides media upload and download workflows.


shortcuts/slides/slides_media_download.go
```diff
@@ -0,0 +1,221 @@
...
+
+// SlidesMediaDownload downloads a Slides media file token to a local image
+// file. It retries through the Drive source-file preview artifact when direct
+// export is denied, which is the path available for media embedded in Slides.
+var SlidesMediaDownload = common.Shortcut{
+	Service:     "slides",
+	Command:     "+media-download",
+	Description: "Download a Slides media file_token to a local image file",
+	Risk:        "read",
+	Scopes:      []string{"drive:file:download"},
```
fangshuyu-768: [P2] Declare a scope accepted by the direct media endpoint

The first request targets `/drive/v1/medias/:file_token/download`, but that endpoint does not list `drive:file:download` as an accepted scope; the existing `docs +media-download` owner uses `docs:document.media:download`, which matches the [endpoint contract](https://open.feishu.cn/document/server-docs/docs/drive-v1/media/download). With a minimally scoped token derived from this metadata, the direct request always returns 403 and `source=download` is unreachable, so every invocation silently degrades to the preview path. Please either declare an accepted direct-download scope and gate `drive:file:download` only when entering the fallback, or skip the direct request.

shortcuts/slides/slides_media_download_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if strings.Contains(err.Error(), "preview") {
+		t.Fatalf("non-permission error should not fall back: %v", err)
+	}
+}
+
+func TestSlidesMediaDownloadDryRunIncludesFallback(t *testing.T) {
+	f, stdout, _, _ := cmdutil.TestFactory(t, slidesTestConfig(t, ""))
+
+	err := runSlidesShortcut(t, f, stdout, SlidesMediaDownload, []string{
```
fangshuyu-768, line 153: [P2] Add the required shortcut E2E coverage

This is a new shortcut, so the repository contract requires both dry-run and live E2E coverage. The package-level test here stubs both HTTP calls and only searches the rendered output for substrings; it cannot catch root-command registration/manifest drift, incorrect scope preflight, or whether a real Slides media token supports the direct-to-preview transition. Please add a structured dry-run test under `tests/cli_e2e/slides` that pins both requests and `preview_type=16`, plus a self-contained live upload/download/byte-verification/cleanup flow (or document the concrete live-fixture blocker as required by AGENTS.md).

shortcuts/slides/slides_media_upload.go
```diff
@@ -177,8 +177,6 @@ var SlidesMediaUpload = common.Shortcut{
 		}
 
 		fileName := filepath.Base(filePath)
-		fmt.Fprintf(runtime.IO().ErrOut, "Uploading: %s (%s) -> presentation %s\n",
```
fangshuyu-768, line 180: [P3] Preserve the existing upload progress output

This PR also removes the stderr progress messages from `+media-upload` and the placeholder-upload path used by `+create`. Those existing human-facing behaviors are unrelated to media download, and uploads of up to 20 MB or multiple images now appear idle. Please restore both progress writes, or move an intentional behavior change into a separately justified and tested change.

fangshuyu-768 (approved):
Reviewed the latest head (b7583dc). No blocking findings; targeted checks and CI, including dry-run and live E2E, pass. The remaining --output/--output-dir diagnostic issue is non-blocking.

liangshuo-1 (approved):

liangshuo-1 (approved):

PR #2449, opened 2026-08-22
fix(wiki): keep node-get stderr machine-readable

liujinkun2025:
## Summary
Remove the unconditional `Fetching wiki node ...` progress line from `wiki +node-get`. Normal successful calls no longer emit that progress text, and API failures no longer have that shortcut progress line before root error rendering.

This PR intentionally leaves existing conditional warnings and the deprecated `--token` compatibility warning unchanged.

## Changes
- Stop writing the unconditional progress line before the `get_node` API request.
- Assert that the normal successful path emits no progress output.
- Assert that covered API failure paths emit no shortcut progress text before root error rendering.

## Test Plan
- [x] `go test ./shortcuts/wiki/... -count=1`
- [x] `make vet`
- [x] `make fmt-check`
- [x] `go mod tidy -diff`
- [x] `QUALITY_GATE_CHANGED_FROM=origin/main make quality-gate`
- [ ] `make unit-test` (the changed `shortcuts/wiki` package passes; the full target is blocked by existing local service-catalog/metadata mismatches in `cmd/schema`, `internal/schema`, and `internal/affordance`)

## Scope
- The hidden deprecated `--token` alias and its existing pflag warning are unchanged.
- The conditional `--space-id` verification warning is unchanged.
- Broader stderr notice/deprecation policy is outside this focused fix.

## Related Issues
- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an unnecessary error message displayed during wiki node retrieval.
  * Successful requests and terminal business errors now complete without unexpected terminal output.

### DIFF
--- a/shortcuts/wiki/wiki_node_get_test.go
+++ b/shortcuts/wiki/wiki_node_get_test.go
@@ -404,8 +404,8 @@ func TestWikiNodeGetMountedExecuteParsesURLAndFormatsOutput(t *testing.T) {
 	if _, ok := data["url"]; ok {
 		t.Fatalf("did not expect a url field in +node-get output, got %#v", data["url"])
 	}
-	if got := stderr.String(); !strings.Contains(got, "Fetching wiki node") {
-		t.Fatalf("stderr = %q, want fetching message", got)
+	if stderr.Len() != 0 {
+		t.Fatalf("stderr = %q, want no progress output", stderr.String())
 	}
 }
 
@@ -444,7 +444,7 @@ func TestWikiNodeGetMountedClassifiesTerminalBusinessErrors(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Setenv("LARKSUITE_CLI_CONFIG_DIR", t.TempDir())
 
-			factory, stdout, _, reg := cmdutil.TestFactory(t, wikiTestConfig())
+			factory, stdout, stderr, reg := cmdutil.TestFactory(t, wikiTestConfig())
 			reg.Register(&httpmock.Stub{
 				Method: "GET",
 				URL:    "/open-apis/wiki/v2/spaces/get_node",
@@ -482,6 +482,9 @@ func TestWikiNodeGetMountedClassifiesTerminalBusinessErrors(t *testing.T) {
 			}
 			if stdout.Len() != 0 {
 				t.Fatalf("stdout = %q, want no success envelope", stdout.String())
+			}
+			if stderr.Len() != 0 {
+				t.Fatalf("stderr = %q, want no output before the root error envelope", stderr.String())
 			}
 		})
 	}
--- a/shortcuts/wiki/wiki_node_get.go
+++ b/shortcuts/wiki/wiki_node_get.go
@@ -98,8 +98,6 @@ var WikiNodeGet = common.Shortcut{
 			return err
 		}
 
-		fmt.Fprintf(runtime.IO().ErrOut, "Fetching wiki node %s...\n", common.MaskToken(spec.Token))
-
 		data, err := runtime.CallAPITyped("GET", "/open-apis/wiki/v2/spaces/get_node", spec.RequestParams(), nil)
 		if err != nil {
 			return wikiNodeGetProblem(err)


shortcuts/wiki/wiki_node_get_test.go
```diff
@@ -483,6 +483,9 @@ func TestWikiNodeGetMountedClassifiesTerminalBusinessErrors(t *testing.T) {
 			if stdout.Len() != 0 {
 				t.Fatalf("stdout = %q, want no success envelope", stdout.String())
 			}
+			if stderr.Len() != 0 {
+				t.Fatalf("stderr = %q, want no output before the root error envelope", stderr.String())
```
fangshuyu-768, line 487: [P2] Preserve machine-readable errors for the legacy flag

`--token` is still a supported compatibility alias, but the pflag `MarkDeprecated` call writes its warning directly to `FlagSet.Output()` before execution. If the invocation then returns a typed config or API error, stderr becomes a deprecation line followed by the JSON envelope instead of a single parseable JSON object. These assertions only inspect the runtime `ErrOut` and always invoke `--node-token`, so they miss that path. Please route the deprecation notice through the structured notice/envelope mechanism, or otherwise suppress raw flag output on failures, and add a root-level regression test that invokes `--token` and validates the complete stderr as JSON.

shortcuts/wiki/wiki_node_get_test.go
```diff
@@ -404,8 +404,8 @@ func TestWikiNodeGetMountedExecuteParsesURLAndFormatsOutput(t *testing.T) {
 	if _, ok := data["url"]; ok {
 		t.Fatalf("did not expect a url field in +node-get output, got %#v", data["url"])
 	}
-	if got := stderr.String(); !strings.Contains(got, "Fetching wiki node") {
-		t.Fatalf("stderr = %q, want fetching message", got)
+	if stderr.Len() != 0 {
+		t.Fatalf("stderr = %q, want no progress output", stderr.String())
```
fangshuyu-768, line 408: [P2] Cover the remaining success warning path

`--space-id` still leaves a supported success path that violates this new invariant: when `get_node` returns a node without `space_id`, `Execute` writes `Warning: ...` directly to `ErrOut` at `wiki_node_get.go:121-123` and then returns success. In JSON/`--jq` mode, a runner that merges streams again receives non-composable output, which is the behavior this PR is meant to remove. This assertion only exercises a response containing `space_id`, so it cannot catch that branch. Please make the unverifiable assertion a typed failure or carry it in structured output, and add a case with `--space-id` plus a response lacking `space_id` that validates the complete stderr contract.

shortcuts/wiki/wiki_node_get_test.go
liujinkun2025: Thanks, this is a valid pre-existing gap for the deprecated `--token` alias. This PR does not change the alias registration or its documented warning behavior; the shipped `lark-wiki` reference explicitly documents that warning. Fixing it correctly requires a separate decision on structured deprecation notices, along with corresponding documentation and regression tests. I have narrowed this PR to removing only the unconditional `Fetching...` progress line. Would you be okay with handling the legacy alias behavior separately?

shortcuts/wiki/wiki_node_get_test.go
liujinkun2025: Thanks. This is an intentional conditional warning: when `--space-id` is requested but the API response omits `space_id`, silently succeeding would incorrectly imply that the assertion was verified. The current repository contract sends warnings to stderr and does not guarantee JSON composability after stdout and stderr are merged. This PR only removes the unconditional `Fetching...` progress output and leaves conditional warnings unchanged; the PR scope now states this explicitly. I therefore propose no change for this path.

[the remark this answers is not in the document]
liujinkun2025: Duplicate of the existing reviewer discussion: https://github.com/larksuite/cli/pull/2449#discussion_r3842362882. The conditional warning is intentionally unchanged under the narrowed PR scope.

shortcuts/wiki/wiki_node_get_test.go
```diff
@@ -483,6 +483,9 @@ func TestWikiNodeGetMountedClassifiesTerminalBusinessErrors(t *testing.T) {
 			if stdout.Len() != 0 {
 				t.Fatalf("stdout = %q, want no success envelope", stdout.String())
 			}
+			if stderr.Len() != 0 {
+				t.Fatalf("stderr = %q, want no output before the root error envelope", stderr.String())
+			}
```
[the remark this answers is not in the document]
liujinkun2025: Duplicate of the existing reviewer discussion: https://github.com/larksuite/cli/pull/2449#discussion_r3838125987. The deprecated alias behavior is intentionally unchanged in this focused PR and is proposed for separate follow-up.

fangshuyu-768 (approved):
Re-reviewed at 438e6adb. No actionable findings; the focused change and regression coverage look good.

PR #2459, opened 2026-08-24
fix(mail): align send_as sender selection

bubbmon233:
## Summary

- use default send_as metadata when resolving compose/reply senders
- keep reply sender priority aligned with explicit `--from`, explicit mailbox, original recipient match, default send_as, then primary fallback
- make `mail +draft-edit --from` update the draft From header while `--mailbox` selects the owning mailbox
- update mail skill docs for the sender-selection behavior

## Validation

```bash
go test ./shortcuts/mail -run 'TestResolveComposeSender|TestResolveReplySender|TestPickSendAs|TestBuildDraftEditPatch|TestResolveComposeMailboxID|TestRequireSenderForRequestReceipt|TestRequestReceipt' -count=1
```

## Summary by CodeRabbit

* **New Features**
  * Improved sender selection for sending, replying, forwarding, drafts, and receipt requests.
  * Added support for send-as addresses, aliases, default senders, and matching original recipients.
  * Sender display names now appear in message and receipt headers.
  * Draft editing can set and validate the final `From` address.
  * Send-receipt requests can specify a sender.

* **Documentation**
  * Updated command help and mail guides with the new sender-selection rules.

* **Bug Fixes**
  * Prevented unintended mailbox fallback behavior.


shortcuts/mail/mail_draft_edit.go
```diff
@@ -27,8 +27,8 @@ var MailDraftEdit = common.Shortcut{
 	AuthTypes:   []string{"user"},
 	HasFormat:   true,
 	Flags: []common.Flag{
-		{Name: "from", Default: "me", Desc: "Mailbox email address containing the draft (default: me). Prefer --mailbox for clarity; --from is kept for backward compatibility."},
-		{Name: "mailbox", Desc: "Mailbox email address that owns the draft (default: falls back to --from, then me). Takes priority over --from when both are set."},
+		{Name: "from", Desc: "Sender email address for the draft's From header. When using an alias (send_as) address, set this to the alias and use --mailbox for the owning mailbox."},
```
sysuljx, line 30: [P1] Do not repurpose the public `+draft-edit --from` flag in place. Its existing documented contract is the mailbox containing the draft, retained specifically for backward compatibility. With this change, an old invocation like `+draft-edit --from shared@example.com --draft-id ...` reads the draft from `me` and also rewrites its From header. Keep `--from` as the legacy mailbox alias and introduce an unambiguous flag such as `--set-from` for the header change, or continue using the typed `set_header From` patch operation.

shortcuts/mail/helpers.go
```diff
@@ -290,16 +287,12 @@ func resolveMailboxID(runtime *common.RuntimeContext) string {
...
+// Priority: --mailbox > "me". When sending via an alias (send_as), use
+// --mailbox for the owning mailbox and --from for the alias sender address.
 func resolveComposeMailboxID(runtime *common.RuntimeContext) string {
 	if mb := runtime.Str("mailbox"); mb != "" {
 		return mb
 	}
-	if from := runtime.Str("from"); from != "" {
-		return from
-	}
 	return "me"
```
sysuljx, line 296: [P1] Preserve the legacy `--from` mailbox-routing contract. Before this change, `resolveComposeMailboxID` used `--mailbox > --from > me`, so existing commands such as `mail +send --from shared@example.com ...` created the draft under the shared mailbox. Returning `me` here changes the API path to `/user_mailboxes/me/drafts` while only putting the shared address in the From header; that can fail authorization or use the wrong owning mailbox. Default send_as resolution does not require this break: keep the `--from` fallback for mailbox routing unless `--mailbox` explicitly separates ownership from sender identity.

shortcuts/mail/helpers.go
```diff
@@ -353,19 +346,169 @@ func extractPrimaryEmail(data map[string]interface{}) string {
...
+func resolvePrimarySenderInfo(runtime *common.RuntimeContext, explicitMailbox string) composeSenderInfo {
+	if mailbox := strings.TrimSpace(explicitMailbox); isExplicitMailboxSender(mailbox) {
+		return composeSenderInfo{Email: mailbox}
 	}
 	email, _ := fetchMailboxPrimaryEmail(runtime, "me")
-	return email
+	return composeSenderInfo{Email: email}
+}
+
+func resolveDefaultSendAs(runtime *common.RuntimeContext, mailboxID string) (composeSenderInfo, bool) {
```
sysuljx: [P1] Remove this unused resolver before merging. Both required `lint` and `deadcode` checks currently fail on `resolveDefaultSendAs` (`unused` / `unreachable func`), so the PR cannot become green while it remains. The active callers already go through `fetchSendAsAddresses` and `pickSendAsAddress` directly.

PR #2472, opened 2026-08-24
fix: preserve non-data payload keys in SuccessEnvelopeData

### LINKED ISSUE #2428: api command silently drops non-"data" response payloads (e.g. bot/v3/info returns empty data)
## Summary

`lark-cli api` (the raw HTTP escape hatch) reports success but returns an empty payload for any API whose response does not wrap its payload under the `data` key.

Example: `GET /open-apis/bot/v3/info` returns its payload under `bot`, and the CLI prints `{"ok": true, "identity": "bot", "data": {}}` — the bot info is silently discarded.

## Repro

```
$ lark-cli api --as bot GET /open-apis/bot/v3/info
{
  "ok": true,
  "identity": "bot",
  "data": {}
}
```

Raw HTTP with the same tenant token shows the server returned the full payload:

```
$ curl -H "Authorization: Bearer $TENANT_TOKEN" https://open.larksuite.com/open-apis/bot/v3/info
{"code":0,"msg":"ok","bot":{"activate_status":2,"app_name":"...","avatar_url":"...","ip_white_list":[],"open_id":"ou_..."}}
```

## Root cause

`internal/output/envelope_success.go` — `SuccessEnvelopeData`:

```go
data, ok := m["data"]
if !ok || data == nil {
    return map[string]interface{}{}   // drops everything when payload is not under "data"
}
```

The unwrap assumes every Lark API wraps its payload in `data`. That holds for most endpoints but not all (`bot/v3/info` uses `bot`). The behavior is currently pinned by `TestSuccessEnvelopeData_MissingDataUsesEmptyObject` and `TestSuccessEnvelopeData_NilDataUsesEmptyObject`.

There is also no `--raw` / `--verbose` flag on `api` to bypass the unwrap, so the lost payload is unrecoverable from the CLI.

## Suggested fix (tested locally)

When `data` is absent or nil, fall back to the envelope minus the transport fields instead of an empty object:

```go
func SuccessEnvelopeData(result interface{}) interface{} {
    m, ok := result.(map[string]interface{})
    if !ok {
        return map[string]interface{}{}
    }
    if data, ok := m["data"]; ok && data != nil {
        return data
    }
    fallback := make(map[string]interface{}, len(m))
    for k, v := range m {
        if k == "code" || k == "msg" || k == "data" {
            continue
        }
        fallback[k] = v
    }
    return fallback
}
```

This preserves the JSON output contract (no `code`/`msg` leak into business data) and keeps both existing tests green — their inputs reduce to an empty fallback. With the patch:

```
$ lark-cli api --as bot GET /open-apis/bot/v3/info
{
  "ok": true,
  "identity": "bot",
  "data": {
    "bot": {
      "activate_status": 2,
      "app_name": "...",
      "open_id": "ou_..."
    }
  }
}
```

Regression test added locally: `TestSuccessEnvelopeData_NonDataPayloadKeyPreserved` (fails if the fix is reverted).

A `--raw` flag on `api` to emit the untouched envelope would additionally help debugging.

## Environment

- lark-cli 1.0.88 (npm `@larksuite/cli`, darwin/arm64)
- Brand: lark (open.larksuite.com)

dajiaohuang:
## Summary

When an API response uses a non-"data" key (e.g., `/bot/v3/info` returns payload under `"bot"`), the previous implementation discarded the payload and returned an empty object.

## Root cause

`internal/output/envelope_success.go` — `SuccessEnvelopeData` assumed every Lark API wraps its payload in `data`:

```go
data, ok := m["data"]
if !ok || data == nil {
    return map[string]interface{}{}   // drops everything when payload is not under "data"
}
```

## Fix

When `data` is absent or nil, fall back to the envelope minus the transport fields (`code`, `msg`, `data`) instead of an empty object. This preserves the JSON output contract (no `code`/`msg` leak into business data) and keeps existing tests green.

## Testing

- Existing tests `TestSuccessEnvelopeData_MissingDataUsesEmptyObject` and `TestSuccessEnvelopeData_NilDataUsesEmptyObject` still pass (fallback is empty when no other fields present)
- Added regression test `TestSuccessEnvelopeData_NonDataPayloadKeyPreserved`

## Example

Before:
```
$ lark-cli api --as bot GET /open-apis/bot/v3/info
{"ok": true, "identity": "bot", "data": {}}
```

After:
```
$ lark-cli api --as bot GET /open-apis/bot/v3/info
{"ok": true, "identity": "bot", "data": {"bot": {"activate_status": 2, "app_name": "...", "open_id": "ou_..."}}}
```

Fixes #2428

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed successful responses so business payloads remain available when no nested `data` field is provided.
  - Preserved response content under alternate fields while excluding protocol metadata.
- **Tests**
  - Added regression coverage to verify payload preservation and removal of transport fields.


dajiaohuang:
Rebased the existing branch onto current upstream main to resolve the DIRTY merge state and kept the SuccessEnvelopeData fallback behavior. Focused validation: go test ./internal/output (239 passed). Current mergeability is clean; remaining blocked state is CI/review gating.

PR #2476, opened 2026-08-24
fix(mail): gate triage labels output behind flag

bubbmon233:
## Summary
- omit labels from mail +triage structured output unless --labels is passed
- preserve existing labels behavior when --labels is requested
- keep search metadata batch_get gated by --labels and leave table labels column behavior unchanged

## Tests
- go test ./shortcuts/mail -run 'TestMailTriageStructuredOutputLabelsFollowFlag|TestMailTriageDryRunQueryWithLabelsAddsBatchGet|TestMailTriageDryRunListPathUsesMessagesAndBatchGet'

## Summary by CodeRabbit

* **Bug Fixes**
  * Structured mail triage output now omits message labels unless the labels option is enabled.
  * When requested, labels are consistently returned as text across list and search results.
  * Search results now include label metadata only when explicitly requested, improving response efficiency.

### DIFF
--- a/shortcuts/mail/mail_triage.go
+++ b/shortcuts/mail/mail_triage.go
@@ -284,7 +284,7 @@ var MailTriage = common.Shortcut{
 		switch outFormat {
 		case "json", "data":
 			outData := map[string]interface{}{
-				"messages":   messages,
+				"messages":   triageStructuredMessages(messages, showLabels),
 				"mailbox_id": mailbox,
 				"count":      len(messages),
 				"has_more":   hasMore,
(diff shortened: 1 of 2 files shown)


shortcuts/mail/mail_triage.go
```diff
@@ -284,7 +284,7 @@ var MailTriage = common.Shortcut{
 		switch outFormat {
 		case "json", "data":
 			outData := map[string]interface{}{
-				"messages":   messages,
+				"messages":   triageStructuredMessages(messages, showLabels),
```
sysuljx, line 287: **[breaking-change] Preserve the existing structured-output shape**

On current `main`, both list and search JSON/data results always contain `messages[].labels` (an empty string or comma-separated IDs), even when `--labels` is omitted. Routing the output through this helper removes that key for existing invocations. That is an observable wire-contract change: jq/scripts now receive `null`/missing, and consumers with a pinned schema can reject the result. The repository’s Hard Contracts explicitly treats output shapes as compatibility commitments.

The flag description makes the intended behavior understandable, but it does not make the already-shipped shape safe to remove in a patch release. Could we preserve the field on the existing surface (or add an explicit opt-out/versioned output), unless this is intentionally released and documented as a breaking change with migration guidance?

sysuljx (commented):
Reviewed the structured-output compatibility and the list/search implementations. I found one breaking-contract concern inline.

PR #2531, opened 2026-08-27
fix(slides): relax office token length check from 28 to >=25

R0bynZhu:
The interleaved "OFL0X" product/region marker is read at fixed positions (1-based 5/10/15/20/25), so a token only has to be long enough to hold it. Pinning the total length to exactly 28 silently reclassified every other length as native.

28 is already stale: per #2509 the local-office format is "OFL0X + 21 random + 1 office type enum" = 27 characters, and sheets relaxed the identical guard to >= 25 in that PR. Slides was missed, so an imported office deck at the current length uploaded with parent_type "slide_file" instead of "office_slide_file".

The marker positions are unchanged. Relaxing the length is only safe because of them: a false positive is the dangerous direction, since the drive backend does not validate that parent_node actually names an office file, so a misclassified native deck uploads successfully and only surfaces later as an image that will not render. The interleaved native token cases (same length, different marker) are what keep the floor honest.

Tests: replaced two mislabelled rows with five verified ones covering 27, 29, 25, 24 characters and a 28-character token carrying the ppt office type enum. #2509's own labels were off by one or two characters and it never covered the 25/24 boundary; this does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of office-backed slide and spreadsheet media across supported token formats.
  * Corrected handling of token length boundaries and marker positions, including legacy and local-office formats.
  * Ensured media is assigned to the appropriate office-file or native media type more consistently.
* **Reliability**
  * Expanded validation coverage for edge cases, malformed tokens, and office-type variations.

### DIFF
--- a/shortcuts/common/office_token.go
+++ b/shortcuts/common/office_token.go
@@ -1,0 +1,82 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package common
+
+import "strings"
+
+// Legacy synthetic prefixes an imported "office" document token may carry.
+// Exported because callers legitimately need to name them — a test that spells
+// "fake_office_" itself would drift from this list the moment it changes.
+const (
+	FakeOfficeTokenPrefix  = "fake_office_"
+	LocalOfficeTokenPrefix = "local_office_"
+)
+
+// officeTokenPrefixes is the prefix set IsLocalOfficeToken checks.
+var officeTokenPrefixes = []string{FakeOfficeTokenPrefix, LocalOfficeTokenPrefix}
+
+// officeTokenMarker is the interleaved product/region marker an imported office
+// token carries.
+const officeTokenMarker = "OFL0X"
+
+// officeTokenMarkerOffsets are the byte offsets the marker occupies in an
+// interleaved token — positions 5, 10, 15, 20, and 25, 1-based. The array
+// length is tied to the marker so the two cannot drift apart silently: adding a
+// character to one without the other stops compiling.
+var officeTokenMarkerOffsets = [len(officeTokenMarker)]int{4, 9, 14, 19, 24}
+
+// officeTokenMinLen is the shortest token the marker can be read out of, one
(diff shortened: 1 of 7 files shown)


shortcuts/common/office_token.go
```diff
@@ -0,0 +1,82 @@
...
+
+// officeTokenMarker is the interleaved product/region marker an imported office
+// token carries.
+const officeTokenMarker = "OFL0X"
+
+// officeTokenMarkerOffsets are the byte offsets the marker occupies in an
+// interleaved token — positions 5, 10, 15, 20, and 25, 1-based. The array
+// length is tied to the marker so the two cannot drift apart silently: adding a
+// character to one without the other stops compiling.
+var officeTokenMarkerOffsets = [len(officeTokenMarker)]int{4, 9, 14, 19, 24}
```
xiongyuanwen-byted, line 27: The "cannot drift apart silently" guarantee only holds in one direction: Go zero-fills an under-specified array literal.

I checked this — changing the marker to `"OFL0XY"` compiles fine and yields `[4 9 14 19 24 0]`. `IsLocalOfficeToken` then compares `token[0]` against the marker's 6th byte, so essentially every real local-office token classifies as native — reintroducing exactly the wrong-`parent_type`, image-won't-render failure this PR fixes. Shortening the marker does fail to compile, so the guard is real, but only for one of the two edits it claims to catch.

Today's code is correct, so this is low severity. Two ways out: soften the comment, or derive the offsets instead of listing them —

```go
var officeTokenMarkerOffsets = func() [len(officeTokenMarker)]int {
	var offsets [len(officeTokenMarker)]int
	for i := range offsets {
		offsets[i] = 4 + 5*i
	}
	return offsets
}()
```

which makes the claim true in both directions. `TestOfficeTokenMinLenMatchesMarkerOffsets` does catch the zero-fill case incidentally today, but its failure message points at `officeTokenMinLen` rather than the real cause.

Roy-oss1 (approved):

liangshuo-1 (approved):

PR #2533, opened 2026-08-27
refactor(sheets): keep the success path off stderr

xiongyuanwen-byted:
## Why

Based on the non-failure stderr audit of the five business domains ([doc](https://bytedance.larkoffice.com/docx/CBHAd9tRLoGx8ZxzCm5cusrDnff)), applied to the sheets domain.

Every sheets shortcut that had something to say on a successful run said it on stderr: ignored sub-op locators, emulated dimension semantics, deprecated flag spellings, the dropdown option-error steer, and the upload/export stage lines. PowerShell's native-command handling and most agent harnesses treat non-empty stderr as failure, so a working call reported itself as an error — and the facts a caller actually needs sat outside the JSON they parse.

## What changed

Pure stage text (`Writing image`, `Waiting for export task`, `Created export task`, `Uploading media for import`) is deleted — it duplicates what the result already proves. Everything decision-relevant moved into the payload:

| Field | Carries |
|---|---|
| `data.warnings` | ignored sub-op locators, colliding freezes, the dropdown option-error steer (now also in `--dry-run`), markdown export falling back to the token as file name |
| `data.effective_operation` | `+dim-insert`'s anchor shift under `--inherit-style before`; the whole `(rows, cols)` state a `+dim-freeze` leaves behind |
| `data.deprecation` | `+cells-batch-set-style` and `+dim-freeze`'s legacy `--dimension/--count`, under a key of its own rather than mixed into `warnings` |
| `data.upload` | how `+media-upload` sent the file (`mode`, `size_bytes`) |
| `data.poll` | export/import poll runs that had to retry (`attempts`, `transient_failures`, `last_error`) — added only when a retry actually happened |
| `data.input_corrections` | inputs the CLI rewrote before the request ran; `+workbook-import` uses it for a mislabeled `.xls` that is really an `.xlsx` |

Clean calls keep their exact previous payload shape: every field above appears only when it has something to report.

`+workbook-export` and `+workbook-import` delegate to `drive.RunExport` / `drive.RunImport`, so those shared cores are cleaned up here too — the sheets contract could not hold otherwise. `drive +export` / `drive +import` get the same treatment. Also fixed along the way: the compatibility `+export --output-path` download branch now reports `file_token` and `ticket` (the token previously reached the caller only through the deleted stderr line), and `+workbook-export` rejects a locally opened Office file up front instead of running create + poll into an opaque backend failure.

## Known gaps

Two shared helpers a sheets caller can still reach stay noisy, and are named in the contract test's comment: multipart media upload over 20MB (`common.UploadDriveMediaMultipartTyped`) and the bot-identity auto-grant warning (`common.AutoGrantCurrentUserDrivePermission`, which duplicates its own `permission_grant` result). Both belong to helpers shared with base / doc / slides / wiki / mail; the audit's opt-in real-time progress channel (`ProgressReporter` / `--progress-file`) is also out of scope here.

## Tests

- contract test: a successful run leaves stderr empty — dim-freeze, dropdown-set, cells-batch-set-style, batch-update, plus `+workbook-export` and `+workbook-import` end to end through the shared cores
- per-field tests for `warnings` / `effective_operation` / `deprecation` / `upload` / `poll` / `input_corrections`, including that a request within limits gains no new field
- `+export --output-path` regression test for the export identifiers
- source-scanning guard (no allowlist) so new direct `ErrOut` writes cannot reappear in the sheets packages
- `go build ./...` and the full non-e2e suite pass

## Summary by CodeRabbit

* **New Features**
  * Command results now include warnings, deprecation notices, effective operation details, upload summaries, export identifiers, polling diagnostics, and input corrections.
  * Dry-run previews display relevant warnings.
  * Workbook export rejects locally opened Office spreadsheet files with guidance.

* **Bug Fixes**
  * Successful sheet and drive operations no longer write progress or warning messages to stderr.
  * Improved validation for unsupported workbook exports and corrected mislabeled import extensions.

* **Documentation**
  * Updated dropdown warning behavior in command reference documentation.


shortcuts/sheets/lark_sheet_sheet_structure.go
```diff
@@ -166,13 +166,23 @@ var DimInsert = common.Shortcut{
...
+		// shift; the executed call has to as well, or a caller diffing the
+		// request against what they typed reads it as an off-by-one.
+		if dimInsertAnchorShifted(runtime, input) {
+			out = annotateSheetsResult(out, "effective_operation", map[string]interface{}{
+				"requested_position": strings.TrimSpace(runtime.Str("position")),
+				"anchor_position":    input["position"],
+				"side":               input["side"],
+				"inherit_style":      "before",
+				"note":               "--inherit-style before is emulated: the request anchors one row/column earlier and inserts after it, which lands the new row/column at requested_position while copying the PRECEDING style.",
+			})
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in c80dd7df — `TestDimInsertReportsEmulatedAnchorInPayload` asserts `requested_position` / `anchor_position` / `side` / `inherit_style` for an `--inherit-style before` insert, plus the negative case: `--inherit-style after` rewrites nothing, so it must not gain the block.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -2208,6 +2212,10 @@ var WorkbookExport = common.Shortcut{
 		if p.Token, err = resolveSpreadsheetTokenExec(runtime); err != nil {
 			return err
 		}
+		// Re-check after the wiki hop: Validate only saw the node_token.
+		if err := errLocalOfficeExportUnsupported(p.Token); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in c80dd7df — `TestWorkbookExport_LocalOfficeBehindWikiNodeRejected` stubs `get_node` returning a `local_office_` obj_token and asserts the typed `validation/failed_precondition`, plus that the (registered, optional) `export_tasks` stub captured zero requests. Verified it fails when the Execute-side guard is removed.

shortcuts/drive/drive_export_test.go
```diff
@@ -1933,3 +1933,91 @@ func TestWrapExportContextErr(t *testing.T) {
...
+	}
+	poll, _ := envelope.Data["poll"].(map[string]interface{})
+	if poll == nil {
+		t.Fatalf("expected a poll summary after a retried poll, got %#v", envelope.Data)
+	}
+	if poll["transient_failures"] != float64(1) {
+		t.Errorf("poll.transient_failures = %v, want 1", poll["transient_failures"])
+	}
+	if last, _ := poll["last_error"].(string); last == "" {
+		t.Errorf("poll summary should carry the last transient error, got %#v", poll)
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in c80dd7df — the test now asserts `poll.attempts == 2` (one failure, then the ready status) alongside `transient_failures` and `last_error`.

shortcuts/sheets/result_annotations.go
```diff
@@ -0,0 +1,81 @@
...
+// because of the payload's shape, so a non-object result is wrapped as
+// {"result": <original>, <key>: <value>} rather than discarded. Object results,
+// which is every case in practice, are annotated in place and keep their shape.
+func annotateSheetsResult(out interface{}, key string, value interface{}) interface{} {
+	switch typed := out.(type) {
+	case map[string]interface{}:
+		typed[key] = value
+		return typed
+	case nil:
+		return map[string]interface{}{key: value}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Not taking the code change, but the contract it flagged was genuinely ambiguous — clarified in c80dd7df.

`out == nil` means the tool returned **no output at all** (`callTool` returns `nil, nil` for an empty `data.output`), not "a JSON null result". There is nothing to preserve, so `{"result": null, ...}` would invent a field naming a result that never existed, and every consumer would then have to distinguish "result is null" from "there was no result". The wrapper exists for array/scalar output, where the tool's own answer would otherwise be lost.

The doc comment lumped nil in with non-object output, which is what made this read as a bug; it now spells out all three shapes separately. `TestAnnotateSheetsResultShapes` pins them: object annotated in place, array preserved under `result`, empty result with no invented `result` key.

shortcuts/drive/drive_import.go
```diff
@@ -141,18 +157,14 @@ func RunImport(ctx context.Context, runtime *common.RuntimeContext, p ImportPara
...
 	ticket, err := createDriveImportTask(runtime, spec, fileToken)
 	if err != nil {
 		return err
 	}
 
 	// Step 3: Poll task
-	fmt.Fprintf(runtime.IO().ErrOut, "Polling import task %s...\n", ticket)
-
-	status, ready, err := pollDriveImportTask(runtime, ticket)
+	status, ready, pollSummary, err := pollDriveImportTask(runtime, ticket)
```
fangshuyu-768, line 167: [P2] Preserve recovery state after remote work has started

Once `createDriveImportTask` succeeds, the import is already running. If `pollDriveImportTask` returns an error, this branch returns before `ticket`, `input_corrections`, or `poll` are emitted; removing the preceding stderr line therefore drops the only stable recovery handle. The same success-only migration also affects `shortcuts/sheets/backward/lark_sheets_spreadsheet_management.go`: download/save errors return after `file_token` is known but before the new fields are written. Please attach the ticket/file token and resume/download command to every post-creation typed error (or return a structured partial result), and cover the all-polls-fail and download/save-failure paths.

shortcuts/sheets/lark_sheet_batch_update.go
```diff
@@ -96,17 +96,18 @@ var BatchUpdate = common.Shortcut{
...
 			return err
 		}
 		if len(plan.localFailures) > 0 {
 			out = mergeBatchUpdatePartialOutput(out, plan)
 		}
-		runtime.Out(compactBatchChartCreateOutput(out), nil)
+		// Ignored sub-op locators and emulated / colliding semantics decide
+		// whether a caller can safely retry, so they travel with the result
+		// instead of on stderr, where a success-path write reads as a failure.
+		out = appendSheetsWarnings(compactBatchChartCreateOutput(out), batchWarnings(runtime))
```
fangshuyu-768: [P2] Keep preflight advisories on failed or partial batch calls

If `callTool` returns an error after some operations were applied, execution exits before `batchWarnings` is attached. Those warnings state which locators were ignored and which freeze operations overwrite each other, so they determine the actual target and the safe retry set; they were previously emitted before the request. Please preserve them in the typed-error or partial-result path while keeping success stderr empty, and add a partial-failure regression test.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -2216,6 +2224,22 @@ var WorkbookExport = common.Shortcut{
...
 }
 
+// errLocalOfficeExportUnsupported rejects an export whose target is a locally
+// opened Office spreadsheet (isOfficeSpreadsheet: a "local_office_"/"fake_office_"
+// token, or an interleaved OFL0X one). Such a token names a local file the Lark
+// client is showing, not a cloud document, so the drive export task can only
+// fail on the backend — and it fails late, after the create + poll round trips,
+// with an opaque message. Refuse up front and say why instead.
+func errLocalOfficeExportUnsupported(token string) error {
+	if !isOfficeSpreadsheet(token) {
```
fangshuyu-768, line 2244: [P2] Do not treat every Office-style token as a local disk file

The existing helper contract says `fake_office_` and OFL0X identify imported Office spreadsheets (`helpers.go:54-68`); a user may only have a shared/cloud token and no local source file. This guard rejects every such token and then says the workbook already exists on disk, which is not a valid recovery action for that case. Please distinguish truly local tokens from imported Office resources, or keep the rejection but provide a recovery path that does not assume access to the original disk file, with separate token-class tests.

shortcuts/drive/drive_export.go
```diff
@@ -277,11 +284,16 @@ func RunExport(ctx context.Context, runtime *common.RuntimeContext, p ExportPara
...
-	fmt.Fprintf(runtime.IO().ErrOut, "Created export task: %s\n", ticket)
 
 	var lastStatus driveExportStatus
 	var lastPollErr error
 	hasObservedStatus := false
+	// Transient poll failures are swallowed and retried, so without a count in
+	// the result a caller cannot tell a clean export from one that limped to
+	// the finish line. That, not the per-attempt narration this replaces, is
+	// the part worth reporting.
+	pollAttempts := 0
```
fangshuyu-768, line 303: [P2] Retain TTY-only liveness during bounded polling

This path can now remain silent for about 45 seconds, while Drive import and the legacy Sheets export can be silent for about 58 and 30 seconds. Silence is correct for captured agent and CI runs, but interactive users get no indication that the CLI is alive. `RuntimeContext.StartSpinner` is already gated by `StderrIsTerminal` and is a no-op for pipes, CI, and captured output, so please wrap the long poll/upload sections with it instead of removing all interactive liveness.

shortcuts/drive/drive_import.go
xiongyuanwen-byted: Fixed in d6cca2e4 — right, the ticket was only visible because polling narrated itself.

`RunImport` now wraps a poll failure with the ticket and `drive +task_result --scenario import --ticket <t>`. The legacy sheets `+export` path got the same treatment: download and save failures happen after the artifact is ready, so they now carry `ticket`, `file_token` and `drive +export-download --file-token <t>`, and a poll timeout carries the ticket too.

Covered by `TestDriveImportPollFailureKeepsTicket` (all polls fail) and `TestSheetExportDownloadFailureKeepsExportIdentifiers` (artifact ready, download 500).

shortcuts/sheets/lark_sheet_batch_update.go
xiongyuanwen-byted: Fixed in d6cca2e4 — agreed, fail-fast without rollback is exactly when those advisories matter.

`batchWarnings` / `batchIgnoredLocatorNotes` are now computed before the call and, on error, attached to the typed error's hint via `attachSheetsWarningsToError` (category / subtype / code / log_id propagate unchanged, so the failure is still one JSON envelope on stderr). Applied to `+batch-update`, `+batch-chart-create` and `+batch-chart-update`.

`TestBatchUpdateKeepsAdvisoriesOnFailure` drives a "1 succeeded, 1 failed" response and asserts the ignored-locator note survives in the hint while the tool's own message stays the message.

shortcuts/drive/drive_export.go
xiongyuanwen-byted: Good call — fixed in d6cca2e4.

I had not registered that `StartSpinner` is already gated on `StderrIsTerminal` and returns a no-op closure otherwise, so this costs the machine contract nothing. Wrapped the drive export poll, the drive import poll and the import media upload, plus the legacy sheets export poll. `stop()` is idempotent and is called before every result write so the cleared line cannot interleave with stdout.

The contract tests capture stderr into a buffer (not a TTY) and still assert it is empty, so the no-op path stays pinned.

shortcuts/sheets/lark_sheet_workbook.go
xiongyuanwen-byted: Fixed in d6cca2e4 — the rejection stays (drive export only produces artifacts for native Lark documents), but the recovery no longer assumes a local file.

Split by token class:
- `local_office_` / `fake_office_` prefix — a synthetic token the client mints for a file opened from the user's own disk, so "the workbook is already a file you have" is accurate.
- interleaved `OFL0X` — an Office file stored in Lark, possibly only shared with the caller. Message now says "Office file stored in Lark", and the hint points at `drive +download --file-token <token>`, then `sheets +workbook-import` if they want a Lark spreadsheet they can export.

`TestWorkbookExport_LocalOfficeTokenRejected` is now table-driven per class and asserts the OFL0X hint does **not** claim a local copy exists.

shortcuts/sheets/success_stderr_contract_test.go
```diff
@@ -0,0 +1,359 @@
...
+// of overwriting each other, and that an empty warning list leaves the payload
+// byte-identical — the property that keeps clean calls' output shape unchanged.
+func TestAppendSheetsWarningsMergesAndSkips(t *testing.T) {
+	t.Parallel()
+
+	merged := appendSheetsWarnings(map[string]interface{}{"warnings": []string{"first"}}, []string{"second"})
+	mergedMap, _ := merged.(map[string]interface{})
+	if got, _ := mergedMap["warnings"].([]string); len(got) != 2 || got[0] != "first" || got[1] != "second" {
+		t.Errorf("warnings should accumulate, got %#v", mergedMap["warnings"])
+	}
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Done in d6cca2e4 — added the `[]interface{}` case (the shape a decoded payload actually carries) and asserted both values survive in order.

shortcuts/sheets/lark_sheet_workbook_export_test.go
```diff
@@ -137,3 +137,122 @@ func TestWorkbookExport_CreateRateLimitKeepsCallerRecovery(t *testing.T) {
...
+		},
+	}
+
+	_, err := runShortcutWithStubs(t, WorkbookExport, []string{
+		"--url", "https://example.feishu.cn/wiki/wikTestNODE", "--as", "user",
+	}, getNode, createTask)
+
+	requireProblem(t, err, errs.CategoryValidation, errs.SubtypeFailedPrecondition, "cannot be exported")
+	if len(createTask.CapturedBodies) != 0 {
+		t.Errorf("no export task may be created for a local Office file, got %d request(s)", len(createTask.CapturedBodies))
```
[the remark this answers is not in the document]
xiongyuanwen-byted: Not applicable as described, but it did surface a real doc bug — fixed in d6cca2e4.

`registry.go` appends to `CapturedBodies` on **every** match (`registry.go:132-133`), outside the `if !s.Reusable` branch above it; `Reusable` only controls whether the stub stays matchable. So a one-shot `export_tasks` request is recorded and `len(CapturedBodies) != 0` does fail the test — verified empirically by deleting the Execute-side guard, which makes this test fail with `got 1 request(s)`.

What was wrong is the field's doc comment ("records every captured request body when Reusable is set"), which is what this reading came from. Corrected it to state the actual behaviour and that `len(CapturedBodies)` is a sound negative assertion.

shortcuts/sheets/lark_sheet_workbook.go
fangshuyu-768: Thanks, the OFL0X branch fixes that token shape, but this does not resolve the full finding. The repository contract introduced in #1606 explicitly describes `fake_office_` as an imported Office spreadsheet, and #2063 describes OFL0X as the newer token format while retaining `fake_office_` / `local_office_` as legacy forms of the same Office classification. That evidence does not establish that either prefix means the caller still owns a disk copy, so `isLocallyOpenedOfficeToken` still returns a recovery action the caller may be unable to take. Please use recovery guidance that does not assume a local source file for all Office-backed tokens, or provide authoritative evidence for distinct storage classes and encode that distinction in the classifier/tests.

fangshuyu-768 (approved):
Reviewed current head d6cca2e4. Targeted Drive/Sheets tests, Sheets CLI E2E, go vet, and diff checks pass.

liangshuo-1 (approved):

PR #2538, opened 2026-08-27
fix(client): surface later-page failures in --page-all runs

### LINKED ISSUE #2477: --page-all 在第二页失败时仍返回成功结果
我在看分页实现时发现，第一页成功后，如果后续页面请求失败，`--page-all` 会停止分页，但仍然把前面拿到的数据当成成功结果返回。

我用 fake transport 稳定复现了两种情况：

- 第二页返回网络错误：`PaginateAll` 的 error 仍然是 `nil`
- 第二页返回非零 API code：合并结果仍然保留第一页的 `code: 0`，`CheckResponse` 检查不出错误

例如第二页返回 `code: 999999` 后，最终结果还是：

```json
{
  "code": "0",
  "msg": "ok",
  "data": {
    "items": [{"id": "first"}],
    "has_more": false
  }
}
```

看起来原因是 `paginateLoop` 在后续页面失败时只执行了 `break`，而 `mergePagedResults` 又以第一页作为最终结果的基础。

期望后续页面失败时返回非零退出状态。流式输出可以保留已经写出的数据，但不能让调用方误以为结果完整。

复现版本：`main@06798847`

和 #590、#1859 的区别是：那两个是 page limit 截断，这个是后续请求失败后仍然返回成功。

Emrys1105:
## Summary

`--page-all` stopped paginating when a page after the first failed, but returned the pages it already had as a complete, successful result: exit code `0` and an `ok: true` envelope. With a non-zero business `code` the merged result also reported `has_more: false`, so the output actively asserted completeness and callers had no signal at all that data was missing. Export, sync and reconciliation jobs silently consumed truncated data as if it were whole.

The pagination loop folded "I stopped" and "I finished" into the same shape — both returned a `nil` error. This aligns the loop with the semantics the project already uses elsewhere: `internal/pagination.Walk` returns the error on any failed page, and its callers propagate it.

## Changes

- Propagate transport failures from later pages in `paginateLoop` (`internal/client/client.go`) instead of breaking out and returning a `nil` error
- Return a typed error via `CheckResponse` when a later page carries a non-zero business `code`, keeping the first-page branch untouched so the existing raw-response output contract still holds
- Correct the stale `PaginationOptions.Identity` field comment in `internal/client/pagination.go` (comment only; the field is now read by `paginateLoop`)
- Add failure-path regression tests in `internal/client/pagination_failure_test.go`, asserting typed error category and subtype for transport failures, unknown business codes and known business codes
- Lock empty stdout on a failed paginated run in `cmd/api/api_paginate_test.go` and `cmd/service/service_paginate_test.go`

Exit codes follow the existing `errclass` classification and add no new mapping: a transport failure exits `4`, an unparseable response exits `5`, and a business error exits by its category.

## Test Plan

- [x] `make unit-test` passed
- [x] `make vet` and `make fmt-check` passed
- [x] `go test ./internal/client/ ./cmd/api/ ./cmd/service/` passed — the three packages this change touches
- [x] Live check against a test tenant: a multi-page `--page-all` run still ends in an `ok: true` envelope with exit `0`, and a first-page business error still writes the raw response to stdout. Both behaviours are unchanged by this PR.
- [x] Failure-path verification: the pre-fix binary was driven against a local TLS-terminating proxy with per-request control over each page. On a page-2 business failure it returned `{"ok":true,...,"has_more":false}` with exit `0`; the fixed binary exits by error category with empty stdout. Covered deterministically by the new unit tests, which fail if either branch is reverted.

## Related Issues

Closes #2477


internal/client/pagination_failure_test.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+	}), nil
+}
+
+// errSimulatedTransport is a stable sentinel so the tests can assert the
+// transport error is carried through the pagination loop rather than merely
+// classified as a network failure — the loop must not swallow the cause.
+var errSimulatedTransport = errors.New("simulated transport failure")
+
+func transportFailure() (*http.Response, error) {
```
Emrys1105, line 47: ✅ Fixed in 13c3a4f8 for the transport-failure part: `transportFailure` now returns a package-level sentinel and both transport cases assert `errors.Is` alongside the category check, so a dropped cause fails the test. Verified by temporarily replacing the propagated error with a fresh one — both cases fail with "cause was not preserved".

🔍 Not actionable for the other two:

- The known business-code case already asserts the full typed contract. `TestPaginateAll_LaterPageKnownBusinessCodeMatchesPageOneClassification` uses `errs.ProblemOf` and checks `errs.SubtypeUserUnauthorized` (and the unknown-code case checks `errs.SubtypeUnknown`). Lines 164-169 are a different test — the `StreamPages` regression lock, which intentionally asserts category only, since that path was already correct before this change and the test exists to keep it that way.
- The command-layer cases assert stdout is empty rather than error metadata, on purpose. The defect being fixed is a successful-looking result leaking to stdout, so that is the contract those tests guard; the typed-error contract is covered one layer down in `internal/client`. Duplicating it there would blur what each layer is responsible for.

PR #2549, opened 2026-08-28
fix(client): surface later-page failures in --page-all runs

### LINKED ISSUE #2477: --page-all 在第二页失败时仍返回成功结果

Emrys1105:
## Summary

`--page-all` stopped paginating when a page after the first failed, but returned the pages it already had as a complete, successful result: exit code `0` and an `ok: true` envelope. With a non-zero business `code` the merged result also reported `has_more: false`, so the output actively asserted completeness and callers had no signal at all that data was missing. Export, sync and reconciliation jobs silently consumed truncated data as if it were whole.

The loop folded "I stopped" and "I finished" into the same shape — both returned a `nil` error — and it recognised only two failure signals: a transport error, or a JSON object whose `code` read as a non-zero float. A page that failed without one, most commonly a gateway 5xx, was accumulated as a normal page. A page whose `code` could not be read at all — absent, `null`, a string, `0.5`, `1e-324` — read as `0` through the float, and the loop paginated from it, re-issuing the request for its cursor. `api --page-all` is not restricted to `GET`, so that re-issued a `POST` or `DELETE`.

This aligns the loop with semantics the project already uses. A failed page is classified exactly as `HandleResponse` classifies a plain `api` response, in the same order. Whether a page is a step in the pagination — a continuation an earlier page promised, or a page to paginate from — is decided by one strict check: its business `code` is exactly `0`. That check is asymmetric by page position on purpose, and it is the one decision that was missing.

## Changes

All production changes are in `internal/client/client.go` unless noted.

- Fail the run on a later page's transport error instead of breaking out and returning a `nil` error
- Classify a failed page in the order `HandleResponse` uses for plain `api`: `CheckResponse` first, so a readable business `code` wins over the HTTP status (a `403` carrying `230027` is an authorization error), then an HTTP status of 400 or more. Because the paginated path now runs the same `CheckResponse` call and the same status check the plain path does, a business or HTTP failure classifies the same way — same category, same `error.code` — whether or not `--page-all` is set. Page 1 keeps its output contract: it is handed back whole with a `nil` error so the command layer classifies it and dumps the raw response
- Treat a page as a step in the pagination only if its business `code` is exactly `0`, as decided by `pageDeclaresSuccess`, which reads a JSON number as zero only if every digit of its mantissa is `0`. `CheckResponse` cannot make that call: it reads the code as a float, under which `0.5`, `1e-324` and an absent field all become `0`. The rule is asymmetric by page position, deliberately. A **later** page that did not declare success fails the run before it is emitted or accumulated — the page before it promised more data, and accepting this one, with or without a cursor and data of its own, would end the run as complete over that promise. A **first** page that did not declare success but carries a usable continuation token is refused before output with a non-zero exit. For `GET`/`HEAD`, the `hint` explains how to remove pagination/filtering and save the raw response with `--output`; for other methods it tells the caller to verify whether the first request changed remote state before retrying. `api --page-all` is not restricted to `GET`, so re-issuing the request for its cursor could replay a write, and an error rather than a stopped run is what gives the streaming formats, which have no envelope to carry `has_more`, a machine-readable signal. A first page that did not declare success and carries no usable continuation token is output on the existing contract: there is nothing to paginate, a code-less list is still a list, and plain `api` would have shown the same response
- **Deliberate change:** `main` paginated from a page whose `code` was absent or unreadable, re-issuing the request for its cursor and then reporting `has_more: false` over the merged result; it also accepted such a page as a later page. Both now fail with `invalid_response`. The one first-page case that still outputs — no cursor, so nothing to paginate — is where `main` special-cased a non-zero float (`0.5`, `1e999`) and withheld it from the streaming formats; such a page passes `CheckResponse` and the status check like any other, so it is now streamed like any other
- Add `callPage`, which keeps the `*larkcore.ApiResp` that `CallAPI` drops — nothing else in the loop can read an HTTP status — and decides whether the body is JSON by the same rule `HandleResponse` uses, now shared as `isJSONBody`. A body that declares a non-JSON Content-Type is never parsed: on a 4xx/5xx it is classified by status (a `text/html` gateway page exits `4` as plain `api` does, not `5`); on a 2xx it is reported as not a page, because plain `api` treats it as a download and a paginated run cannot. `main` parsed such a body anyway and followed the cursor it found in it
- Make `ParseJSONResponse` (`internal/client/response.go`) accept exactly one JSON value. It is shared by `HandleResponse`, `CallAPI`, `callPage`, the shortcut runner's `ClassifyAPIResponseWith` — so every shortcut's response classification — and the event runtime, and the stricter parse applies on each of those paths. It decoded the first value and ignored the rest, so `{...}garbage` and `{...} {"code":230027}` both passed as the first object, cursor included. Trailing whitespace is still accepted
- Resolve the identity the classification uses inside the loop — `opts.Identity`, then the request's own `As`, then `AsUser`, treating `AsAuto` as unresolved — because an empty identity is rewritten to `user` by `errclass` and hands a bot the wrong recovery text
- Report in `StreamPages` how much a streaming run emitted before it failed, counting only pages the output callback accepted
- Correct the `PaginationOptions` field comments in `internal/client/pagination.go`, and widen the `SubtypeInvalidResponse` comment in `errs/subtypes.go` to cover a response the CLI cannot accept as the envelope it expects (comments only)
- Tests: `internal/client/pagination_rules_test.go` states the rule once as a decision table — code shape × HTTP status × first/later page × `PaginateAll`/`StreamPages` — with every cell's expectation derived from `CheckResponse`'s own verdict, the status, and the page position, and asserts request count, callback count, category and `error.code`. Every shape runs with both supported cursor fields (`page_token` and `next_page_token`), as a terminal list, and with no data at all — the last being the shape a failed later page most often has — and a page beyond the cursor is always available, so the table proves the rule in both directions: a zero page's cursor is followed, any other page is refused or, on a first page that needs no pagination, output; the code shapes cover absent, every spelling of `0`, integers at 2^53 and `int64`, fractions, underflow, out-of-range and non-numbers. A parity test asserts that every one of those responses, plus bodies with trailing garbage or a second value, classifies identically through `HandleResponse` and through a first page; a declared non-JSON 2xx body is pinned as not a page; and the parser's one-value contract is pinned directly. `pagination_failure_test.go` keeps the Content-Type, identity, transport and stream-summary cases; `cmd/api/api_paginate_test.go` and `cmd/service/service_paginate_test.go` pin stdout — empty for a buffered format on a later-page failure, the pages already written for a streaming one, a first page that needs no pagination on each format's own contract — a code-less list streams as a list — a first page that carries a usable continuation token but did not declare success being refused with a non-zero exit and nothing on stdout, and a later page without data that did not declare success failing the run instead of merging as complete

Deliberately left to a follow-up: the business code is still read by `CheckResponse` and `errclass.intFromAny` through `float64`, so `230027.5` classifies as `230027` and `9007199254740993.0` reports `error.code` `9007199254740992`. Plain `api` and `--page-all` now behave identically there, which is this change's contract; making the decode exact means changing every command's classification and belongs in its own change with its own compatibility review. Also unchanged is the loop's continuation metadata — a non-boolean `has_more`, a `has_more` without a token, a repeating token — which is pre-existing behaviour on the success path.

Exit codes follow the existing `errclass` classification and add no new mapping: a transport failure or a `5xx` exits `4`, an unparseable or non-JSON page exits `5`, and a business error exits by its category. A page that passes `CheckResponse` and the status check but not the exact-zero gate exits `5` — unless it is a first page carrying no usable continuation token, in which case it is output as plain `api` would output it. The refusal does not consult `--page-limit`: a first page that carries a continuation token but did not declare success is refused even under `--page-limit 1`, where the loop would not have followed the token — the rule is about the page, not about how far the run was allowed to go.

## Test Plan

- [x] `make unit-test`, `make vet`, `make fmt-check` passed
- [x] `go test ./internal/client/ ./cmd/api/ ./cmd/service/` passed
- [x] Decision table: 20 code shapes × 3 statuses × 2 page positions × 2 entry points × 4 data variants (`page_token`, `next_page_token`, terminal list, no data) = 960 cells, with expectations derived from the rule rather than written per cell; parity test over the same shapes plus non-object, unparseable and trailing-content bodies against `HandleResponse`
- [x] Control-flow regressions: `--page-limit 1` still refuses an invalid first-page cursor before output; both cursor field names participate in the same rule; recovery hints are safe by default for unknown/write methods
- [x] Differential against `main`: a 462-case matrix (page position × HTTP status × Content-Type × 18 body shapes) compiled against both `main` and this branch for `PaginateAll` and `StreamPages`. No case issues more requests than `main`; no case succeeds where `main` failed; every difference is `main` accepting a failed page, following a cursor it should not have, or classifying a failure wrongly
- [x] Mutation-checked: reverting the transport return, exact-zero gate, page-position asymmetry, Content-Type handling, one-value parser, or float-free zero check makes the rule-specific tests fail. Additional control-flow mutations — dropping `next_page_token`, moving the page-limit stop before page admission, or giving write methods the `GET`/`HEAD` retry hint — also fail the new tests
- [x] Live check against a test tenant: a multi-page `--page-all` run still ends in an `ok: true` envelope with exit `0`, and a first-page business error still writes the raw response to stdout

## Related Issues

Closes #2477

### DIFF
--- a/cmd/api/api_paginate_test.go
+++ b/cmd/api/api_paginate_test.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/larksuite/cli/errs"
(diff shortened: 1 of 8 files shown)


Emrys1105:
Both remaining review threads are addressed in `4be90bb1`. Summarising here so the reasoning is in one place.

**`cmd/api/api_paginate_test.go` — assert the typed error contract**

Done, and it caught something. The command-layer cases now assert the classification rather than mere failure: `TestAPIPaginate_LaterPageTransportErrorEmitsNoStdout` keeps the simulated transport error in a variable and asserts `errors.Is(err, transportErr)` plus `CategoryNetwork`, and `TestAPIPaginate_LaterPageBusinessErrorEmitsNoStdout` asserts `CategoryAuthorization` and `SubtypeUserUnauthorized` for `230027`.

The two HTTP-status cases needed the same treatment for a different reason. A `502` body carrying no `code` is also an unreadable page, so the later-page guard caught it and both tests still passed with the status branch reverted — they were not load-bearing until the category assertion isolated them from the other branch.

**`internal/client/pagination_failure_test.go` — malformed-body cases for `callPage`**

The finding was right that the branch was untested; the conclusion went the other way. Driving each shape through `callPage` showed the SDK unmarshals the body inside `DoAPI`, so anything `ParseJSONResponse` would reject has already failed there with no response attached: a truncated body, an HTML `502`, `"code": "230027"` and `"code": {}` all return an `InternalError` before the parse branch runs. The status classification I had put there could never fire, so it is removed and `callPage` is now `CallAPI` plus the response and nothing else.

`TestPaginateAll_UnparseableLaterPageFailsInsideDoAPI` pins that reasoning instead, so the branch becoming reachable again breaks a test rather than going unnoticed. It also records the consequence: an unreadable body on a `5xx` reports as an invalid response rather than a network failure, which is what plain `api` reports for the same response — the two paths still agree.

That investigation also surfaced a real gap in the guard this PR had just added. `util.ToFloat64` returns `(0, false)` for anything that is not a number, and the guard keyed on the `code` field being *present*, so a page carrying `"code": null` passed as successful — the same silent truncation the guard exists to stop, one shape further in. It now keys on decodability, covered by `TestPaginateAll_LaterPageUndecodableCodePropagates`.

Every new branch was mutation-checked: reverting it fails exactly its own cases and no others.

Emrys1105:
Correction to my previous comment: the claim that `callPage`'s parse-failure branch "could never fire" is wrong, and so is the change I made on the strength of it. Both are fixed in `9353335d`.

**What I got wrong**

I concluded the branch was unreachable after driving several malformed bodies through `callPage` and watching every one fail earlier, inside `DoAPI`. Every fixture in that check declared `Content-Type: application/json`, and that is the one case where the conclusion holds. The SDK decodes only a body it believes is JSON. A gateway `502` served as `text/html`, or with no `Content-Type` at all, returns from `DoAPI` without an error, with the real status still on the response, and only `ParseJSONResponse` rejects it — landing in exactly the branch I had removed.

So the misclassification the original finding described was real: `--page-all` exited `5` on a `text/html` `502` while plain `api` exits `4`, because `HandleResponse` classifies that same response by status. The paginated path was the only one getting it wrong, on the most likely mid-pagination failure there is.

**What changed**

The status check is restored. The two paths are now covered separately, and the fixtures declare `Content-Type` explicitly rather than taking the helper's default — that default is what hid this:

- `TestPaginateAll_LaterPageNonJSONErrorIsClassifiedByStatus` — `text/html`, `text/plain` and absent `Content-Type`, asserting the category rather than mere failure.
- `TestPaginateAll_UnparseableJSONLaterPageFailsInsideDoAPI` — the JSON-`Content-Type` path, unchanged, still failing inside `DoAPI`.

`504` is deliberately absent from the first: the SDK maps it to a server timeout inside `DoAPI`, so a case for it would pass whether or not the branch exists. Reverting the branch now fails all four remaining cases and nothing else.

**Also in this commit**

`StreamPages` counted items before `onItems` accepted them, so a page the writer refused was still counted. The failure summary could report more items than reached stdout, in the one run where that number is what the caller is reading. Counted after acceptance now, covered by `TestStreamPages_FailureCountExcludesRejectedItems`.

Apologies for the noise — the earlier comment asserted the opposite with more confidence than the evidence supported.

PR #2559, opened 2026-08-29
feat(sheets): cut the top command-error clusters from the 08-18..24 eval batch

xiongyuanwen-byted:
## Why

A trace analysis over **14,818** `lark-cli sheets` calls attributed **2,036** command errors to 57 `(subcommand, flag)` groups; the top 6 cover **76%**. This PR addresses the four that are ours to fix, each at the layer that produced it.

Reproduced every group against `main` first — several were already prescribed correctly and are left alone; what follows is only what still failed.

### `--styles` vocabulary — 291 cases, the only group whose retry also failed

- **One shared prescription for the border family.** `foldBorderFamilyAliases` already absorbs `border` / `borders` / `border_<side>` / `border_<attr>`. What still reached the error path was the Lark OpenAPI's `border_type` (`FULL_BORDER` / `OUTER_BORDER`) and CSS's `border_width` — real vocabularies with no equivalent here, and `border_type` was the single top field in the group. Neither maps unambiguously onto a per-side style/weight/color triple (`FULL` vs `OUTER` differ on interior edges this payload cannot address), so they get a prescription, not a silent alias.
- **Prescribe the nested `{range, style:{…}}` envelope** — the OpenAPI's own request shape, copied one level too deep — plus `bg_color` / `fill_color` / `text_color`.
- **Prescriptions match on the key's letters alone**, so `border_type`, `borderType` and `border-type` are one mistake rather than three.
- **Collapse repeated issues in the `--styles` / `--writes` folds.** One wrong field name in a payload styling N cells produced N identical issues, each re-listing the full supported vocabulary — the fold meant to save round trips was burying its own answer. Six same-field cells went from 1,646 characters of repetition to one statement plus the other locations:

```
--styles has 7 issues (2 distinct): …cell_styles[0].border_type … [same at 5 more: …cell_styles[1].border_type, …[2], …[3], +2 more] | …cell_styles[6].bold …
```

### `--sheets` payload — 419 cases

- **Accept `dtypes` / `formats` as a positional array.** The same pandas habit that produces `columns` and `data` produces `df.dtypes.tolist()`; the payloads were otherwise correct and every retry just rewrote the same information as a map. Only a 1:1 match with `columns` is accepted — a length mismatch is rejected with the alignment rule, never guessed. The map form and the unknown-column check are unchanged, and the canonical form stays the map (wide acceptance layer, one documented shape).

### `+csv-put --file` — 35 cases

- **Read the value as a path.** `--file` is aliased onto `--csv` because agents reach for it, but the two names promise different things: rewriting only the name left the path to be written into the sheet as literal text, which the file-path guard then rejected — an error naming a flag the caller never typed. The read goes through the same `cmdutil.ReadInputFile` as `@file`, so the relative-path policy is identical (an absolute path is still rejected, toward stdin). A value naming nothing readable falls through untouched, so `--file` holding literal CSV keeps working.

### `--help` — 88 cases of `required flag(s) "title" not set`

- **Mark required flags in the sheets help.** `MarkFlagRequired` only sets a completion annotation cobra never renders, so `--title` was listed exactly like an optional flag. Scoped to this domain (rides the existing `withFlagErgonomics` PostMount chain); no other domain's help shifts. Sourced from flag-defs rather than the cobra annotation, because `+chart-create` and `+csv-put` deliberately clear it after mounting while the flag stays required on the real path — and a flag cobra has since put in a one-required group is left unmarked, since neither member of such a pair is individually required.

```
+workbook-create   --title      (required) Spreadsheet title
+chart-create      --properties (required) Full chart config JSON…      # relaxed annotation, still required
+csv-put           --csv        (required) RFC 4180 CSV text…
+csv-put           --start-cell Top-left A1 anchor…                     # one-required pair with --range, unmarked
```

## Not in scope

The largest single group — absolute paths passed to `--file` / `@file` (636 cases) — is deliberately untouched. The cwd-relative policy is a protocol decision (PR #2091), and that thread is being followed up separately.

## Testing

- New: 7 rows in `stylesPriorCorpus`, 4 collapse-fold unit cases, 4 positional-dtypes cases, 5 `--file` path cases, 4 required-help cases. One existing assertion was rewritten into two (distinct defects each render; identical defects collapse).
- `go vet ./...`, `gofmt`, `go test ./shortcuts/... ./cmd/...` clean; `lint/` source-contract scan byte-identical to `main`; `make quality-gate` passes.
- Pre-existing failures on `main`, unrelated and unchanged: `tests/cli_e2e/base` `TestBase_BasicWorkflow`, and the sheets e2e suite (the locally built binary's app lacks `drive:drive` / `space:folder:create`).

## Summary by CodeRabbit

- **New Features**
  - Help output now marks genuinely required flags with “(required)”.
  - `--file` can load CSV content from a readable file path.
  - Sheet data types and formats support positional arrays aligned with columns.
  - Added clearer guidance for common style-field and border-related names.

- **Bug Fixes**
  - Validation errors now consolidate repeated issues while preserving useful locations and counts.
  - Invalid column definitions and mismatched positional data produce clearer alignment errors.
  - CSV input now preserves intended behavior when switching between `--file` and `--csv`.

### DIFF
--- a/shortcuts/common/runner.go
+++ b/shortcuts/common/runner.go
@@ -1266,7 +1266,13 @@ func (ctx *RuntimeContext) InputResolvedFromSource(name string) bool {
 	return ctx.inputResolved[name]
 }
 
-func (ctx *RuntimeContext) markInputResolved(name string) {
+// MarkInputResolved records that the named flag now holds content read from an
+// external source rather than a value typed inline, setting the bit
+// InputResolvedFromSource reports. resolveInputFlags calls it for @file and
+// stdin; a domain calls it when it resolves a source itself (sheets reads the
+// path a +csv-put caller passed as --file), so the shape guards downstream
+// treat the result as content, exactly as they would for --csv @<path>.
+func (ctx *RuntimeContext) MarkInputResolved(name string) {
 	if ctx.inputResolved == nil {
 		ctx.inputResolved = map[string]bool{}
 	}
@@ -1312,7 +1318,7 @@ func resolveInputFlags(rctx *RuntimeContext, flags []Flag) error {
 			// strip a leading UTF-8 BOM so it can't corrupt the first CSV
 			// cell or break JSON parsing downstream.
 			rctx.Cmd.Flags().Set(fl.Name, StripUTF8BOM(string(data)))
-			rctx.markInputResolved(fl.Name)
+			rctx.MarkInputResolved(fl.Name)
 			continue
 		}
 
@@ -1349,7 +1355,7 @@ func resolveInputFlags(rctx *RuntimeContext, flags []Flag) error {
 			// strip a leading UTF-8 BOM so it
 			// can't corrupt the first CSV cell or break JSON parsing downstream.
 			rctx.Cmd.Flags().Set(fl.Name, StripUTF8BOM(string(data)))
-			rctx.markInputResolved(fl.Name)
+			rctx.MarkInputResolved(fl.Name)
 			continue
 		}
 	}
--- a/shortcuts/common/testing.go
+++ b/shortcuts/common/testing.go
@@ -43,7 +43,7 @@ func TestNewRuntimeContextWithBotInfo(cmd *cobra.Command, cfg *core.CliConfig, i
 // domain tests can exercise guards that branch on InputResolvedFromSource
 // without wiring the full resolveInputFlags path.
 func TestMarkInputResolved(rctx *RuntimeContext, name string) {
-	rctx.markInputResolved(name)
+	rctx.MarkInputResolved(name)
 }
 
 // TestNewRuntimeContextForAPI creates a RuntimeContext ready for HTTP tests:
--- a/shortcuts/sheets/csv_put_guard_test.go
+++ b/shortcuts/sheets/csv_put_guard_test.go
@@ -153,20 +153,42 @@ func TestResolveCSVPathFromFileAlias(t *testing.T) {
 
 	t.Run("reads the named file", func(t *testing.T) {
 		rctx := newCSVFileAliasRuntime("./data.csv")
-		resolved, err := resolveCSVPathFromFileAlias(rctx)
-		if err != nil || !resolved {
-			t.Fatalf("resolved = %v, err = %v; want the file to be read", resolved, err)
+		if err := resolveCSVPathFromFileAlias(rctx); err != nil {
+			t.Fatalf("resolve: %v", err)
 		}
 		if got, _ := rctx.Cmd.Flags().GetString("csv"); got != "a,b\n1,2\n" {
 			t.Errorf("--csv = %q, want the file contents", got)
 		}
 	})
 
+	t.Run("the contents are marked source-resolved", func(t *testing.T) {
+		// Contents read from a file may legitimately look like anything,
+		// including a path — the shape guards downstream must see the same bit
+		// they would for --csv @<path>, or a one-cell CSV holding "report.csv"
+		// is rejected as a caller who forgot the @.
+		if err := os.WriteFile("pathshaped.csv", []byte("report.csv\n"), 0644); err != nil {
+			t.Fatal(err)
+		}
+		rctx := newCSVFileAliasRuntime("./pathshaped.csv")
+		if err := resolveCSVPathFromFileAlias(rctx); err != nil {
+			t.Fatalf("resolve: %v", err)
+		}
+		if !rctx.InputResolvedFromSource("csv") {
+			t.Error("the rewritten value must be marked as read from a source")
+		}
+		if err := guardCSVValueIsNotFilePath(rctx); err != nil {
+			t.Errorf("path-shaped file contents must survive the guard, got: %v", err)
+		}
+	})
+
 	t.Run("an out-of-tree path is rejected toward stdin", func(t *testing.T) {
-		_, err := resolveCSVPathFromFileAlias(newCSVFileAliasRuntime("/tmp/data.csv"))
+		err := resolveCSVPathFromFileAlias(newCSVFileAliasRuntime("/tmp/data.csv"))
 		ve := requireValidation(t, err, "relative path")
 		if ve.Param != "--file" {
 			t.Errorf("param = %q, want the flag the caller actually typed", ve.Param)
+		}
+		if ve.Cause == nil {
+			t.Error("the underlying path error should be preserved as Cause")
 		}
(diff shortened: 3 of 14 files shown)


xiongyuanwen-byted:
Thanks — verified all four findings against the code before acting; two were real bugs, both now fixed in 224a416b.

**1. Alias provenance not cleared by a later canonical occurrence — valid, fixed.**

Reproduced: `--file a.csv --csv ./second.csv` silently read `second.csv` as a path, where a bare `--csv ./second.csv` correctly errors.

The suggested fix (clear the record on a canonical occurrence) does not work as stated, though: pflag runs the normalizer for `Lookup` and `Set` too — once with the canonical name immediately after the rewrite, and again on every later lookup, including the framework's own `resolveInputFlags` before `Validate`. Clearing there would wipe a legitimate `--file` attribution on a plain `--file x.csv` call.

Fixed by moving the commit point off the normalizer: it now only *stages* the spelling, and the flag's `pflag.Value` commits it on `Set`, which runs exactly once per real occurrence. Last occurrence wins in either order. Same shape as `flagalias.Bind`'s own `pendingSource`/commit, which this can't reuse because `Bind` advertises its aliases on the flag and in the exported manifest while these rewrites are deliberately silent. Regression tests cover both orders plus `--file` alone.

**2. Rewritten value not marked source-resolved — valid, fixed.**

Reproduced: a file containing `report.csv` was read correctly and then rejected by `csvPutInput`'s shape check. `RuntimeContext.MarkInputResolved` is now exported (the bit already existed for `@file` / stdin; a domain that resolves a source itself needs to set it) and `resolveCSVPathFromFileAlias` sets it. `--file <path>` is now equivalent to `--csv @<path>` all the way down, which also removed the special case that was skipping the guard from `Validate`. Regression test added for path-shaped contents.

**3. Typed-error assertions — valid, applied.**

Both sites now assert `Param` and a non-nil `Cause` alongside the message, including the shared prescription-corpus runner that covers the seven new rows.

**4. Expected-shape hint still says maps only — acknowledged, not changing.**

Deliberate. `style_vocab.go` states the domain contract this file follows: one canonical form is documented, the acceptance layer beneath it is wide and undocumented. The hint is a repair instruction, and the map form it prescribes is always correct — including for a caller who already wrote a valid positional array. Advertising the array form would make the wide layer part of the contract, which is what that rule exists to prevent.

xiongyuanwen-byted:
Self-review of the full branch turned up two more defects in the `--file` rule, both fixed in 962f2061.

**1. An unreadable `--file` path was answered under `--csv`.**

The fall-through kept only the "value names nothing" case in mind, but it also caught paths that name something unreadable. Reproduced through the mounted shortcut:

- `--file ./typo.csv` → `--csv value "./typo.csv" looks like a file path…`, Param `--csv`, hint offering `--csv @<path>` — the wrong flag named.
- `--file ./noread.csv` (mode 0000) → `…is an existing file… pass the same path with an @ prefix`, which cannot work: `@file` routes through this same reader and fails identically.

Now exactly one case falls through: a value that names nothing **and** is not path-shaped, i.e. literal CSV text, which `--file` accepted before this rule existed. Everything else answers under `--file`:

| input | answer |
|---|---|
| absolute / out of tree | `--file … must be a relative path` + stdin |
| path-shaped, names nothing | `--file "…" names no file under the current directory` |
| exists, unreadable / a directory | `--file cannot read file "…"` (hint points at stdin, not `@`) |
| literal CSV text | passes through to the `--csv` guard, unchanged |

**2. The alias spelling was staged with no `Parsed()` guard.**

The first commit had one; I dropped it when moving the commit point to the flag's `Value`, on the reasoning that `Value.Set` is the only committer. But staging still happens in the normalizer, and `chainFlagAliases` looks its own aliases up while installing — so with `PostMount` composed twice (which `installAliasProvenance` explicitly anticipates and handles), `Lookup("file")` replays through the already-installed normalizer at mount time and arms `pending`; the next real `--csv ./data.csv` occurrence then commits it and the value gets read from disk instead of meeting its guard.

Restored the guard. Confirmed the new double-mount regression test fails without it (`expected error, got nil`) and passes with it.

`go vet` / `gofmt` / `go test ./shortcuts/... ./cmd/... ./internal/...` clean; `lint/` contract scan byte-identical to `main`; `make quality-gate` passes.

xiongyuanwen-byted:
Second review pass — both valid, fixed in cbe3a8f4.

**Clear alias staging independently of `Parsed()`** — correct, and it reproduces.

`Parsed()` stays true once parsing has started, so my guard only covered a remount that happens *before* the first parse. Wrote the parse → remount → reparse test you asked for and it fails on the previous commit exactly as described: an explicit `--csv ./data.csv` is read from disk instead of meeting its guard.

Closed it from the other side rather than by dropping `Parsed()`: re-running `installAliasProvenance` (which every remount does) now resets staging, so a spelling staged by a remount's own alias lookups cannot survive into the next occurrence. `Parsed()` stays as the pre-parse half of the same invariant — together they cover both orders, and each half now has a test that fails without it.

**Assert `Cause` on the unreadable-file and directory branches** — applied; they now match the sibling subtest that already asserted it.

---

On CI: the `e2e-live` failure on the previous head is `tests/cli_e2e/base` `TestBase_RoleWorkflow` — `+role-create` succeeds, then the role is missing from `+role-list`, so every later subtest fails on an empty role ID. That is the live Base tenant, not this branch: the only file outside `shortcuts/sheets/**` in this PR is a rename of `RuntimeContext.markInputResolved` to `MarkInputResolved` plus its three call sites (`git diff main...HEAD -- shortcuts/common/` is 12 lines, all mechanical), which no Base command reaches. `results` is the aggregate job and follows `e2e-live`. Watching the run on the new head.

xiongyuanwen-byted:
CI is green on `cbe3a8f4` — [run 33295498157](https://github.com/larksuite/cli/actions/runs/33295498157), `e2e-live` included.

That settles the earlier `e2e-live` red: `tests/cli_e2e/base` `TestBase_RoleWorkflow` is flaky against the live tenant, not a regression here. Two things confirm it — the same suite passed on #2558 at 08:57 the same day, and in the failing log `+role-list` returned only the three template defaults (`Editor` / `普通用户` / `Reader`, `total: 3`) with byte-identical `role_id`s across independent re-runs, i.e. the listing was not seeing the base the role had just been created in. Worth a look from the Base side, but nothing this branch touches.

The one remaining red check in the rollup is the `results` job of run 33295421461, which was cancelled mid-flight by the concurrency group when the current run superseded it; the same check on the live run is green.

liangshuo-1 (approved):

PR #2567, opened 2026-08-31
feat(im): add concise message output

xiaoxiangyu-123:
## Summary

Add a command-scoped `--concise` option to IM chat and thread message lists. It renders compact Markdown for agent context without changing the generic `--format` pipeline or existing message conversion and enrichment behavior.

## Changes

- Add `--concise` only to `im +chat-messages-list` and `im +threads-messages-list`.
- Render chat/thread metadata, participants, messages and replies, reactions, downloaded resource paths, and pagination state as Markdown.
- Use `ChatSections` as the renderer input so future IM message commands can provide their own section boundaries.
- Omit `message_app_link` and other nonessential fields from concise output; JSON remains available for the complete data shape.
- Keep JSON, pretty, table, CSV, NDJSON, dry-run, pagination, conversion, and enrichment paths unchanged.
- Keep `--format concise` on the existing unknown-format path: warning, JSON fallback, and exit code 0.
- Reject `--concise` combined with explicit `--format`, enabled `--json`, or non-empty `--jq`; allow `--json=false` and `--jq=`.
- Scan the actual concise renderer view with the existing content-safety policy before writing output.
- Add concise examples to the two command references without changing Skill routing.

## Test Plan

- [x] `go test ./cmd ./shortcuts/im`
- [x] `make unit-test`
- [x] `make vet`
- [x] `make fmt-check`
- [x] `make build`
- [x] `QUALITY_GATE_CHANGED_FROM=origin/main make quality-gate` (existing warnings only)
- [x] Real user-identity tests from the SCM-built binary: chat/thread concise output, topic replies, empty results, pagination, resource downloads, reaction opt-out, dry-run, legacy formats, unknown-format fallback, and all output-flag conflicts

## Related Issues

Supersedes #2540.

### DIFF
--- a/cmd/root_integration_test.go
+++ b/cmd/root_integration_test.go
@@ -7,6 +7,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
+	"net/http"
 	"os"
 	"strings"
 	"testing"
@@ -14,6 +16,7 @@ import (
 	"github.com/larksuite/cli/cmd/api"
 	"github.com/larksuite/cli/cmd/auth"
 	"github.com/larksuite/cli/cmd/service"
+	extcs "github.com/larksuite/cli/extension/contentsafety"
 	"github.com/larksuite/cli/internal/apicatalog"
 	"github.com/larksuite/cli/internal/build"
 	"github.com/larksuite/cli/internal/cmdutil"
@@ -78,6 +81,11 @@ type typedErrorEnvelope struct {
 		Message string `json:"message"`
 		Hint    string `json:"hint"`
 		Param   string `json:"param,omitempty"`
+		Params  []struct {
+			Name   string `json:"name"`
+			Reason string `json:"reason"`
+		} `json:"params,omitempty"`
+		Rules []string `json:"rules,omitempty"`
 	} `json:"error"`
 }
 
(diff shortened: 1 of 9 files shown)


xiaoxiangyu-123 (commented):
Review conclusion: no merge-blocking functional or compatibility issue found. Concise is scoped to im +chat-messages-list and im +threads-messages-list through command-local help/completion and rendering. Existing JSON, pretty, table, CSV, NDJSON, unknown-format JSON fallback, --json, --jq conflict handling, pagination, and dry-run behavior remain covered by regression tests. One intentional historical behavior change exists in the shared thread expansion path: a thread omitted after the total reply budget is exhausted now receives thread_has_more=true. This is an additive optional JSON field and can also appear in +messages-mget because it shares the same expansion helper; no existing field is removed or retyped. Focused tests, race tests, affordance/schema tests, skill format validation, diff checks, and the latest CI run pass. The existing CodeRabbit inline note about asserting typed error metadata is valid but test-quality-only and non-blocking.

xiaoxiangyu-123 (commented):
Correction / requirement blocker: after comparing this PR with #2540 and the stated two-option requirement, #2567 still implements option 1 (a new command parameter/output mode). The new behavior is entered only through --format concise, while the existing/default JSON and pretty outputs are intentionally preserved. Therefore, if this replacement PR was opened to implement option 2 (simplify the existing output itself instead of adding a separate concise mode), the functional requirement is not met and this is merge-blocking. Compatibility within the implemented option-1 design remains sound; the problem is requirement alignment, not a runtime regression. Please clarify or rework the target output contract before merge.

liangshuo-1 (approved):

PR #2572, opened 2026-08-31
refactor(sheets)!: remove legacy sheets command surface

xiongyuanwen-byted:
> [!WARNING]
> **BREAKING CHANGE.** 42 sheets commands are removed. Callers still on the old
> names now get `unknown subcommand` with no deprecation notice.
> Migration table: [`skills/lark-sheets/references/lark-sheets-legacy-command-migration.md`](skills/lark-sheets/references/lark-sheets-legacy-command-migration.md)

## What

Deletes `shortcuts/sheets/backward` — the package that kept 42 pre-refactor sheets command names (`+create`, `+read`, `+write`, `+create-sheet`, `+media-upload`, ...) alive alongside the refactored ones. Monitoring puts their combined share below 5%.

~10.2k lines of the deletion is the package itself; the rest is cutting the machinery that carried it and porting the callers that still used it.

## Removed alongside the package

- `sheetsAliasReplacement` (the 42-entry alias→replacement map) and `wrapSheetsBackwardDeprecation`, which tagged each alias with a `_notice` deprecation envelope on execution.
- The deprecated cobra group and the custom `sheets --help` usage template that existed only to hide it. `applySheetsCompatGroups` becomes `applySheetsCommandGroups`: it still groups the `+`-shortcuts, so the OpenAPI metaapi subcommands (`spreadsheets`, ...) keep filing under cobra's stock "Additional Commands" heading. Verified `sheets --help` renders identically minus the deprecated block.

## No shared logic was owned by the deleted package

The `parent_type` mapping for sheets image uploads (`sheet_image` / `office_sheet_file`) was, by the deleted file's own header, *"a deliberate mirror of the equivalent note in `shortcuts/sheets/helpers.go`"*. The canonical implementation and `common.IsLocalOfficeToken` are untouched and keep their other callers.

Checked in both directions:

- every one of the package's 116 identifiers grepped repo-wide — the only hits are same-name-different-package coincidences (`sheets.SheetCreate` = `+sheet-create`, `sheets.SheetInfo` = `+sheet-info`);
- the 34 shared symbols it consumed (`common.` / `errs.` / `validate.` / `fileio.`) all still have other users, so nothing is left dead;
- the generated flag data (`data/flag-defs.json`, `flag_defs_gen.go`, ...) carries none of the removed commands — the deleted package defined its flags inline.

## E2E tests ported

`tests/cli_e2e/sheets` still drove the removed commands. Migrated to `+workbook-create` / `+workbook-info` / `+cells-set` / `+cells-get` / `+cells-search` / `+sheet-*`.

The sub-sheet dry-run assertions were rewritten rather than renamed: the old commands posted to `sheets/v2/sheets_batch_update`, the new ones invoke `modify_workbook_structure` over `sheet_ai/v2`. `+update-sheet` fanned out to `+sheet-rename` + `+sheet-hide` + `+dim-freeze`.

Every migrated command was run against a live workbook (created and deleted), which is where the new assertions come from rather than from guesswork:

- `+workbook-create` returns the token at `data.spreadsheet.spreadsheet_token`; `+workbook-info` lists sub-sheets at `data.sheets[]`, not `data.sheets.sheets[]`;
- `+sheet-rename` and `+sheet-hide` answer with a bare revision counter, so their effect is read back from `+workbook-info` (`sheet_name`, `is_hidden`);
- `+cells-search` returns `total_matches` / `matches[].address`, so the test asserts both the hit and the miss — a hit alone would still pass if the search silently matched everything.

## Docs that referenced the removed surface

- Three `skills/lark-drive` reference docs instructed agents to run `sheets +read` / `sheets +find`. These ship embedded in the binary via `content_embed.go`, so the instructions would have produced unknown-subcommand errors. Repointed at `+cells-get` / `+cells-search`, both verified to accept the same flags.
- `skill-template/domains/sheets.md` deleted: every sheets command it named was removed, and the cell payload shape it taught (`{"type":"formula","text":...}`) is rejected outright by `+cells-set`. Nothing consumes `skill-template/domains/`.
- Stale comments naming `backward.uploadSheetMediaFile` and `backward/helpers.go`.

## Left in place

`Shortcut.OnInvoke` and `internal/deprecation` now have no producers. Both are generic framework plumbing wired into the `_notice` envelope in `cmd/root.go` and carry their own tests, so removing them would touch the output contract; that is worth a separate change. The `OnInvoke` doc comment no longer claims a caller.

## Breaking change and migration

A stale skill calling `sheets +read` now gets the ordinary unknown-subcommand error instead of a working command plus a deprecation `_notice`. Edit-distance ranking does not bridge `+read` → `+cells-get`, so there is no runtime migration pointer.

To cover that, this PR ships a migration guide at `skills/lark-sheets/references/lark-sheets-legacy-command-migration.md`, indexed from the lark-sheets `SKILL.md` and reachable at runtime:

```bash
lark-cli skills read lark-sheets references/lark-sheets-legacy-command-migration.md
```

It carries the replacement for all 42 commands plus the differences that are **not** simple renames, which is where a rename-only table would mislead:

- the cell payload vocabulary changed — `{"type":"formula","text":"=SUM(C2:C5)"}` is rejected outright by `+cells-set`, which wants `{"formula":"=SUM(C2:C5)"}` (`--values` itself still works as an alias for `--cells`, so only calls carrying the old object shape break);
- response field paths moved — `+workbook-info` lists sub-sheets at `data.sheets[]` (was `data.sheets.sheets[]`) and no longer echoes the spreadsheet token; `+workbook-create` returns it at `data.spreadsheet.spreadsheet_token`;
- `+sheet-rename` / `+sheet-hide` answer with a bare `revision`, so results must be read back from `+workbook-info`;
- `+update-sheet` and `+update-dimension` fan out to several intent-specific commands;
- filter-view conditions are folded into the view's own flags, and `+media-upload` is folded into `+float-image-create`.

If we also want a runtime pointer, the sheets package has `unknownSubcommandHints` (a prescription table built for exactly this) — happy to add the removed names there in a follow-up.

## Test

- `go build ./...`, `go vet ./...` clean.
- `go test ./shortcuts/... ./cmd/... ./internal/...` passes.
- `LARK_CLI_BIN=<fresh build> go test ./tests/cli_e2e/{sheets,slides,drive}/... -run DryRun` passes.
- Live-verified against a real workbook: the full sub-sheet lifecycle, the cells read/write/search path, and image upload through both surviving surfaces (`+cells-set-image` returned a `file_token`, `+float-image-create` a `float_image_id`). All scratch resources deleted.
- The bot-identity live E2E tests fail in my environment for an unrelated pre-existing reason — the app lacks `drive:drive` / `space:folder:create`, so they die creating a drive folder before reaching sheets. The user-identity live test passes.

## Summary by CodeRabbit

* **New Features**
  * Introduced the updated Sheets command surface for workbook, cell, sheet, and dimension operations.
  * Added refreshed workflows using commands such as `+workbook-create`, `+cells-get`, `+cells-set`, and `+cells-search`.

* **Bug Fixes**
  * Removed obsolete Sheets aliases and legacy commands from registration, help, and completion listings.

* **Documentation**
  * Updated usage guidance and examples for current command names and output formats.
  * Added a migration guide mapping retired commands to their replacements.

---

BREAKING CHANGE: removes the 42 pre-refactor sheets commands (`+create`, `+read`, `+write`, `+append`, `+find`, `+set-style`, `+create-sheet`, `+update-sheet`, `+add-dimension`, `+set-dropdown`, `+media-upload`, `+create-filter-view`, ...). See `skills/lark-sheets/references/lark-sheets-legacy-command-migration.md` for replacements and the non-rename differences.


xiongyuanwen-byted:
Addressed all three review comments in e688292d — each was a real gap, thanks.

**1. Assert the appended row** (`sheets_crud_workflow_test.go`) — correct: the subtest checked only the ok envelope, and the later `+cells-search` covers row 2, so a `+cells-set` that reported success without persisting would have passed. Now reads `A4:C4` back and asserts the row landed.

Checked the new assertion is not vacuous against a live sheet: an unwritten row comes back as cells carrying `cell_styles` but no `value`, so `scalarsIn` does not contain `Charlie` and the read-back fails if the write does not persist.

**2. Omitted-title copy path** (`sheets_sheet_shortcuts_dryrun_test.go`) — correct: the comment on the empty-title suite asserts that an omitted `--title` means "let the server name the copy", but nothing exercised it, so the doc comment was not load-bearing. Added a case that pins `new_name` being *absent* from the payload rather than sent empty. It is non-vacuous in both directions, since the existing copy case pins `new_name == "Copy"` when `--title` is supplied.

**3. `tool_name` in the shared loop** — correct, and the sharper version of the point: all five sub-sheet commands hit the same `/tools/invoke_write` endpoint and are distinguished only by `operation`, so a case selecting a different tool with a matching operation would have passed. Moved the `modify_workbook_structure` assertion into the shared loop after the URL check and dropped the now-duplicated one from the create case.

On the **Docstring Coverage** pre-merge warning (73.08% vs 80%): leaving it. The functions it counts are test helpers and closures whose purpose is already carried by the block comments above them; adding per-closure docstrings to clear a percentage would add noise rather than explanation.

fangshuyu-768 (dismissed):
Reviewed the removal of the pre-refactor Sheets shortcuts, the migrated command references, and the updated regression coverage. No actionable issues found.

caojie0621 (dismissed):

xiongyuanwen-byted:
Both fixed in fbe281be — the first one was a real bug in the migration guide, which is the worst place to have one. Thanks.

**`--sheets` envelope** — correct. The `+append` row showed a bare array. Verified both forms against a dry-run:

```
--sheets '[{"name":"S1","mode":"append",…}]'
  → ok:false  "--sheets: top level must be the object {\"sheets\":[…]}, got a bare JSON array;
               wrap the sub-sheet list in a \"sheets\" key"

--sheets '{"sheets":[{"name":"S1","mode":"append",…}]}'
  → ok:true
```

Now shows the envelope form and says explicitly that a bare array is rejected.

**MD040** — tagged the diagnostic fence as `text`.

Also worth recording: the earlier `deterministic-gate` failure on this branch was the same class of problem — the gate rejects skill-doc examples that reference commands absent from the manifest, and my guide illustrated the *old* form as a runnable `lark-cli sheets +write …` line. Rewritten so the removed spelling is described in prose and only the current command appears as an example. `make quality-gate` is clean locally (rc=0, zero REJECTs).

liangshuo-1 (dismissed):

caojie0621 (approved):

liangshuo-1 (approved):

PR #2585, opened 2026-08-31
feat(slides): lint slide writes server-side, add --no-lint to opt out

R0bynZhu:
+create, +add-slide, +update-slide and +replace-slide are the four shortcuts that change slide content, so they are the four that can ask the backend to check the page before accepting the write. All four now send lint_xml=true by default and lint_xml=false when --no-lint is passed. The subject of the check is the page the write produces, not the payload it was handed: +replace-slide submits fragments, and a fragment that is correct on its own can still push a neighbour off the canvas.

The switch travels in the request body rather than the query string. A query parameter has to be declared in the gateway's own api meta before it is bound to a field, and the published definition of these endpoints does not list one — so an undeclared parameter is dropped, the field arrives unset, and the server reads it as "not requested". Verified against a live backend: pages that asked to be linted were written unlinted, with nothing anywhere to say so. Body fields ride along with the JSON already being sent and need no registration.

The value is sent explicitly in both directions rather than omitted when on. The parameter is newer than the registry, so the server-side default is not something this CLI can read anywhere, and a request that states the value keeps meaning the same thing if that default ever moves.

A refusal is passed through verbatim. It arrives as a JSON document in the message field, the shape the backend also uses for its nodeServer validation failures, and the same refusal reaches callers through `lark-cli api` as well, where nothing rewrites it — rendering it to prose here would give one refusal two formats depending on which command produced it. Each finding carries the numbers behind its own rule, and which numbers those are differs per rule, so nothing is decoded that is not used: the document is what the caller reads.

What the backend cannot say goes in the hint instead: how many findings refused the write, which pages those sit on, and --no-lint, which is a CLI flag the server has never heard of. Every finding in the report refused the write — the gate has no severity threshold, since a level that did not block would mean reporting a defect on a page and writing it anyway — so the page list is built from all of them. Sorting the findings into blockers and advice would tell the caller some of these pages are safe to leave, and cost them the retry.

Tests assert on the wire — the body the stub actually received — rather than on the builder's return value, so a command that stops calling its own builder still fails.

## Changes

- Change 1
- Change 2

## Test Plan

- [ ] Unit tests pass

## Related Issues

- None

## Summary by CodeRabbit

* **New Features**
  * Added server-side layout validation for slide creation, addition, updates, and replacements.
  * Added a `--no-lint` option to bypass validation when needed.
  * Improved error messages with affected pages, blocking issues, and remediation guidance.
  * Clarified partial-progress behavior when page creation fails.

* **Documentation**
  * Documented layout validation, error `4000153`, page preservation, and retry guidance across slide commands.

### DIFF
--- a/shortcuts/slides/slides_lint_param.go
+++ b/shortcuts/slides/slides_lint_param.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package slides
+
+import "github.com/larksuite/cli/shortcuts/common"
+
+// The server-side XML lint switch, carried by every shortcut that writes slide
+// content: +create, +add-slide, +update-slide and +replace-slide.
+//
+// The backend lints the page the write produces and refuses it when that page
+// fails. Note "produces", not "is handed": +replace-slide submits fragments, and
+// the subject of the lint is the assembled page, so a fragment that is valid on
+// its own is still refused when it pushes a neighbour off the canvas. The CLI
+// asks for the check on every call, so a caller who never touches the flag gets
+// the checked path; --no-lint is the escape hatch for the case the lint is wrong
+// about a page that must still ship.
+//
+// Sent explicitly in both directions rather than omitted when on: the parameter
+// is newer than internal/registry/meta_data.json, so the server-side default is
+// not something this CLI can read anywhere, and a request that states the value
+// means the same thing before and after that default ever changes.
+const noLintFlagName = "no-lint"
+
+// lintXMLBodyKey is the wire name the switch binds to, and it travels in the
+// request body rather than the query string.
+//
+// That is not a style choice. A query parameter has to be declared in the
+// gateway's own api meta before it is bound to a field, and the published
+// definition of these endpoints lists only xml_presentation_id, revision_id
+// and idempotency_key. Until that publish happens the gateway drops an
+// undeclared parameter on the floor: the request succeeds, the field arrives
+// unset, and the server reads it as "lint not requested". Verified against a
+// live backend — pages that asked to be linted were written unlinted, with no
+// error anywhere to say so. Body fields ride along with the JSON already being
+// sent and need no separate registration.
+//
+// snake_case matches every other body key these endpoints take: slide,
+// before_slide_id, parts, comment.
+//
+// Worth knowing when this is next touched: a wrong name here still fails
+// silently, for the same reason it did in the query. Tests and --dry-run only
+// prove what the CLI sent, not what was read.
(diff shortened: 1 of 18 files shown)


shortcuts/slides/slides_replace_pages.go
```diff
@@ -91,7 +92,12 @@ var SlidesReplacePages = common.Shortcut{
 				if runtime.Bool("continue-on-error") {
 					continue
 				}
-				return appendSlidesProgressHint(err, fmt.Sprintf("slides +replace-pages stopped at item %d/%d; %d page(s) completed before failure; old page is kept when create failed", i+1, len(resolved.Plan), countReplacedPages(results)))
+				// The lint refusal is enriched here for the same reason as on every
+				// other write path: --no-lint is a CLI flag the backend cannot name.
+				// The progress hint wrapping it carries the page, which the lint hint
+				// deliberately does not — inside one item the report always says
+				// slide 1.
+				return appendSlidesProgressHint(enrichSlidesLintError(err), fmt.Sprintf("slides +replace-pages stopped at item %d/%d; %d page(s) completed before failure; old page is kept when create failed", i+1, len(resolved.Plan), countReplacedPages(results)))
```
fangshuyu-768: [P2] Preserve lint recovery in partial batch results

With `--continue-on-error`, the branch above skips this enrichment, while `replaceOnePage` has already flattened the typed API error to `err.Error()` in `result.Error`. A `4000153` item therefore reaches the only partial-failure stdout result without its code or the `--no-lint` recovery hint, unlike the stop-on-error path. Please enrich and retain structured per-item error metadata before branching on `--continue-on-error`, and add a regression test for this path.

shortcuts/slides/slides_lint_param.go
```diff
@@ -0,0 +1,66 @@
...
+// prove what the CLI sent, not what was read.
+const lintXMLBodyKey = "lint_xml"
+
+// noLintFlag is the shared flag definition, so the three commands cannot drift
+// apart on the name or the wording.
+func noLintFlag() common.Flag {
+	return common.Flag{
+		Name: noLintFlagName,
+		Type: "bool",
+		Desc: "submit the XML without the server-side lint; by default the server lints every page and rejects the write when it fails",
```
fangshuyu-768, line 52: [P2] Do not report lint as performed without backend acknowledgement

This flag only sends `lint_xml=true`; an older or staged backend can accept or ignore that field and return normal success. In that compatibility state, this help text promises every page was linted even though the response provides no acknowledgement, so callers can trust a validation that never ran. Please describe the default as requesting server-side lint (or only claim it after backend capability/acknowledgement is guaranteed), and keep the command references consistent.

PR #2596, opened 2026-09-01
feat: support manifest-based distribution updates

Tantanz20020918:
## Summary

Allow an optional extension provider to select a stable distribution manifest while the CLI owns version checks, verified artifact downloads, and coordinated CLI and Skills updates. Without a manifest, updates still use the existing package-manager installation path. Explicit version queries now also refresh the update cache and its 24-hour timer.

## Changes

- Add the optional `DistributionProvider` contract and lifecycle guidance in `extension/transport`
- Add manifest parsing, artifact download, SHA-256 verification, archive preparation, and rollback-capable installation in `internal/distribution`
- Integrate manifest targets with the asynchronous update cache and `lark-cli update` check/install paths
- Make `lark-cli doctor` check the target from the configured update source; it reports version status, not the manifest URL or distribution identity
- Record the Skills source identity in the shared state file; this does not isolate separate installations sharing the same Skills directories
- Include `fix(event): accept scopes without token type metadata` (`fcc31adf2`): app-version scope entries with absent or null `token_types` are no longer discarded during Bot event scope preflight; explicit token-type lists remain authoritative

## Extension contract boundary

`DistributionProvider` is an optional capability on the existing process-wide transport provider so a wrapper has one aggregate registration point for outbound-network customization; it does not transfer update execution into the extension.

The provider performs only a fast, local lookup of the configured Manifest URL. It does not fetch or parse the Manifest, add authentication headers, download artifacts, verify checksums, unpack archives, compare versions, install files, or perform rollback. Those behaviors remain owned by lark-cli's internal distribution and self-update code.

The returned Manifest URL and artifact URLs contained in that Manifest are final download addresses. They bypass the extension URL rewriter and request interceptor, while retaining the CLI's built-in proxy and custom-CA handling. The current contract has no arbitrary request-header callback; authenticated sources should use an address that is directly fetchable by the built-in client, such as a signed URL, or rely on the configured network/proxy/CA environment.

## Test Plan

- [x] `make unit-test` passed
- [x] `make quality-gate` passed in the preceding validation round
- [x] local HTTP install-and-upgrade E2E passed in the preceding validation round
- [x] manual verification in the preceding validation round: `lark-cli update --check --json` followed by `lark-cli update --json` upgraded a local fixture from 1.0.1 to 1.0.2 and synchronized three Skills directories

- [x] Latest URL-branch follow-up merged; local wrapper E2E passed 19 install/update/SaaS scenarios and 10 login/routing/pruning checks against isolated loopback services. No live tenant API calls were used.
- [x] After the merge: full unit tests, incremental golangci-lint, vet, format checks, and dependency-file consistency passed.

## Related Issues

N/A


Tantanz20020918:
Confirmed:

1. #2581 has been fully superseded by #2595 and #2596. I have closed it and left a note linking the two focused replacements.
2. `DistributionProvider` only performs a fast, local lookup of the Manifest URL. It does not supply arbitrary authentication headers and does not fetch or parse the Manifest, download artifacts, verify checksums, unpack archives, compare versions, install files, or perform rollback; all of those remain owned by lark-cli. It is an optional capability on the existing aggregate transport provider to avoid introducing a second process-wide registry or breaking the existing `Provider` interface. Manifest and artifact URLs are final addresses and bypass extension rewriting/interception while retaining the built-in proxy and custom-CA policy.

I have added this contract boundary to the PR description. Thanks — please continue with the full code and security review.

PR #2606, opened 2026-09-02
feat: expand folder one level in message content (folder children browse)

jackie3927:
## Summary
Extend the IM shortcut suite with folder children browsing: message content rendering expands a `folder`-type message one level via the folder-children API, rendering first-level children (files and subfolders with child counts) directly in message output. The same expansion applies to folder attachments inside `post` messages and to folder replies in threads.

## Changes
- **Folder message expansion** (`folderConverter`): with Runtime + MessageID available, call `GET /open-apis/im/v1/files/:file_key/folder` (srctype=message, srcid=MessageID, recursive=false) and render one level.
- **Output shape** (key-first attributes): root `<folder key="..." name="..." child_count="...">` (child_count = all_count), `has_more="true"` when first-level items exceed the render cap (10) or all_count > items. Children render as `<file key="..." name="..."/>` and nested folders as `<folder key="..." name="..." child_count="..."/>` (no recursion; `child_count` on a nested folder is a depth hint, omitted when unknown). A genuinely empty folder renders `<folder key="..." name="..." child_count="0"/>`.
- **Prefetch**: list commands (chat-messages-list, messages-mget, messages-search, threads-messages-list) scan the page for folder messages **and post folder attachments** before the render loop and fan out the expansions concurrently (bounded to 8) — `PrefetchFolderChildren` caches XML keyed by message id (post attachments keyed message id + folder key). The per-item converter reuses the cache; a prefetch that already failed records a sentinel so the render loop degrades to the single-line tag without doubling requests.
- **Post attachments & threads**: post folder attachments and thread-reply folder messages use the same prefetch/cache path.
- **Graceful fallback**: without Runtime/MessageID or on API failure, output degrades to the single-line `<folder key="..." name="..."/>` — no crash. Failures emit a `folder_fetch_failed` warning on stderr.
- **Unit tests** (httpmock, no network): XML one-level output / unknown child_count omission / has_more / cap-10 / API failure & business-error fallback / empty folder / prefetch success + failure sentinel / converter-level inline + fallback / post folder-attachment expansion (inline + prefetch) / stderr warning prefix.

## Test Plan
- `go test ./shortcuts/im/...` green (convert_lib + shortcuts).
- Manual verification (BOE/PRE): a folder message expands to XML with root `child_count` and file/folder children (e.g. folder `file_v3_01155_...` in message `om_x100b66490b8678acb15859e1e1b66ab`).
- Manual verification: API failure / missing MessageID degrades to single-line folder output with stderr warning, no crash.

## Related Issues
None

### DIFF
--- a/shortcuts/im/convert_lib/content_convert.go
+++ b/shortcuts/im/convert_lib/content_convert.go
@@ -43,6 +43,13 @@ type ConvertContext struct {
 	// per-message inline GET", which keeps non-shortcut callers (events,
 	// ad-hoc tests) working unchanged.
 	MergeForwardSubItems map[string][]map[string]interface{}
+
+	// FolderChildren is an optional pre-fetched cache of folder-expansion XML,
+	// keyed by folder-message message_id. When set, the folder converter uses
+	// the cached entry instead of issuing its own GET; populated by callers
+	// via PrefetchFolderChildren before the FormatMessageItem loop. nil means
+	// "no prefetch — fall back to the per-message inline GET".
+	FolderChildren map[string]string
 }
 
 // converters maps message types to their ContentConverter implementations.
@@ -131,7 +138,7 @@ func FormatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext,
 	if len(senderNames) > 0 {
 		nameCache = senderNames[0]
 	}
-	return formatMessageItem(m, runtime, nameCache, nil, false)
+	return formatMessageItem(m, runtime, nameCache, nil, nil, false)
 }
 
 // FormatMessageItemWithMergePrefetch is like FormatMessageItem but threads a
@@ -140,8 +147,11 @@ func FormatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext,
 // can skip its own per-message GET. Shortcuts that iterate a page of raw
 // items should pre-fetch once and call this variant in the loop to avoid the
 // N × ~1s serial-merge_forward stall in the original code path.
+//
+// Kept as a nil-folder-prefetch compatibility wrapper for external callers;
+// internal list commands have migrated to FormatMessageItemWithFolderPrefetchOpts.
 func FormatMessageItemWithMergePrefetch(m map[string]interface{}, runtime *common.RuntimeContext, nameCache map[string]string, mergePrefetch map[string][]map[string]interface{}) map[string]interface{} {
-	return formatMessageItem(m, runtime, nameCache, mergePrefetch, false)
+	return formatMessageItem(m, runtime, nameCache, mergePrefetch, nil, false)
 }
 
 // FormatMessageItemWithMergePrefetchOpts is FormatMessageItemWithMergePrefetch
@@ -150,11 +160,24 @@ func FormatMessageItemWithMergePrefetch(m map[string]interface{}, runtime *commo
 // without local_path/size_bytes) is attached for the download enrichment stage
 // to fill. The other entry points are thin extractResources=false wrappers, so
 // default output is unchanged.
+//
+// Kept as a nil-folder-prefetch compatibility wrapper for external callers;
+// internal list commands use FormatMessageItemWithFolderPrefetchOpts when a
+// page may contain folder messages.
 func FormatMessageItemWithMergePrefetchOpts(m map[string]interface{}, runtime *common.RuntimeContext, nameCache map[string]string, mergePrefetch map[string][]map[string]interface{}, extractResources bool) map[string]interface{} {
-	return formatMessageItem(m, runtime, nameCache, mergePrefetch, extractResources)
+	return formatMessageItem(m, runtime, nameCache, mergePrefetch, nil, extractResources)
 }
 
-func formatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext, nameCache map[string]string, mergePrefetch map[string][]map[string]interface{}, extractResources bool) map[string]interface{} {
+// FormatMessageItemWithFolderPrefetchOpts is FormatMessageItemWithMergePrefetchOpts
+// with an additional folder-children prefetch cache (built via
+// PrefetchFolderChildren) keyed by folder-message message_id. Callers that
+// render a page that may contain folder messages pass both prefetch maps so
+// the folder converter reuses cached XML instead of N serial GETs.
+func FormatMessageItemWithFolderPrefetchOpts(m map[string]interface{}, runtime *common.RuntimeContext, nameCache map[string]string, mergePrefetch map[string][]map[string]interface{}, folderPrefetch map[string]string, extractResources bool) map[string]interface{} {
+	return formatMessageItem(m, runtime, nameCache, mergePrefetch, folderPrefetch, extractResources)
+}
+
+func formatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext, nameCache map[string]string, mergePrefetch map[string][]map[string]interface{}, folderPrefetch map[string]string, extractResources bool) map[string]interface{} {
 	msgType, _ := m["msg_type"].(string)
 	messageId, _ := m["message_id"].(string)
 	mentions, _ := m["mentions"].([]interface{})
@@ -173,6 +196,7 @@ func formatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext,
 			Runtime:              runtime,
 			SenderNames:          nameCache,
 			MergeForwardSubItems: mergePrefetch,
+			FolderChildren:       folderPrefetch,
 		})
 	}
 
--- a/shortcuts/im/convert_lib/folder_test.go
+++ b/shortcuts/im/convert_lib/folder_test.go
@@ -1,0 +1,398 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package convertlib
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/larksuite/cli/internal/cmdutil"
+	"github.com/larksuite/cli/internal/core"
+	"github.com/larksuite/cli/internal/httpmock"
+	"github.com/larksuite/cli/shortcuts/common"
+	"github.com/spf13/cobra"
+)
+
+// fetchFolderChildrenTree unit tests (httpmock, no real openapi).
+// Covers XML one-level output (folder key+name+child_count / file key+name /
+// sub-folder key+name+child_count / has_more) and fallback paths.
+
+// folderChildrenRoundTrip answers any GET /files/:key/folder request with a
+// one-child expansion (or a failure when fail is true). Built on the roundtrip
+// runtime with a static token resolver so the concurrent prefetch fan-out does
+// not race the credential provider under -race (same approach as merge tests).
+func folderChildrenRoundTrip(fail bool) convertlibRoundTripFunc {
+	return func(req *http.Request) (*http.Response, error) {
+		if fail {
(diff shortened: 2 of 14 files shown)


jackie3927:
Addressed the review (thanks for the detailed pass). Changes pushed to the PR branch.

## Fixed
- **gofmt/license**: extra blank line removed (misc.go), SPDX header added + import ordering fixed (folder_test.go).
- **child_count="0" misleading for sub-folders**: children_count now only emitted when > 0 (missing/unknown → attribute omitted); genuinely empty root folder keeps `child_count="0"`.
- **Path escaping**: folder key now goes through `validate.EncodePathSegment`.
- **Failures no longer fully swallowed**: transport error and business `code != 0` both emit a stderr `warning: folder_fetch_failed: ...` before falling back.
- **Attribute order**: `<file>/<folder>` render `key` first (consistent with media.go and the fallback branch).
- **Query-param assertions**: stub URLs now include the full query (`recursive=false&srcid=om_123&srctype=message`), so srctype/srcid/recursive are actually asserted.
- **First-level render cap**: at most 10 items rendered per folder; `has_more="true"` when more remain (product design — 10-item cap, not unbounded).
- **Post-message folder attachments**: `renderPostAttachments` now expands folder attachments one level via `fetchFolderChildrenTree` (was single-line `<folder>`).
- Tests: C4 (one level), unknown child_count, has_more, cap-at-10, transport failure, business `code!=0` failure, empty folder — all green (`go test ./shortcuts/im/...`).

## Intentional / notes
- **No prefetch fan-out yet**: folder-type messages are low-frequency (a page typically has 0–1), so the single inline RTT is acceptable; if folder usage grows we can follow the `PrefetchMergeForwardSubItems` pattern (merge.go: bounded-concurrency prefetch into ConvertContext).
- **`<file>` inside a folder is intentionally not visually distinguished**: files inside a folder are now downloadable via the same resource endpoints (the comment the review quoted is outdated), so agents downloading them succeeds.
- **all_count semantics**: resource side confirms `all_count` = direct first-level children count (not recursive descendants) — `has_more`/`child_count` logic is consistent.

jackie3927:
## Review follow-up (coderabbitai + human reviewers)

Addressed on top of the prefetch commit (`d7f8327a`):

1. **transport.go PRE endpoint routing (Major)** — removed in `fb777265`: the pre-gateway host redirect was a local pre-env test aid that leaked into the PR. `SecurityHeaderTransport` no longer rewrites `open.feishu.cn`/`accounts.feishu.cn`; transport.go is now identical to main.

2. **Exported formatter signatures (Major)** — restored in `d7f8327a`:
   - `FormatMessageItemWithMergePrefetch` / `FormatMessageItemWithMergePrefetchOpts` keep their original signatures (now nil-folderPrefetch wrappers → non-breaking for existing callers).
   - New `FormatMessageItemWithFolderPrefetchOpts(...)` carries the folder-prefetch cache; list commands that render folder pages use it.

3. **Converter-level & side-effect regression tests** (folder_test.go / text.go comments):
   - `TestFolderConverter_ConvertInline` — Convert issues the GET and renders expanded XML (fails if the Convert-level call is removed).
   - `TestFolderConverter_ConvertInlineFallback` — API failure → single-line fallback + asserts the `folder_fetch_failed` stderr warning.
   - `TestFolderChildrenTree_WarningOnStderr` — stable `folder_fetch_failed:` prefix asserted for transport/business failures.
   - `TestPostConverter_FolderAttachmentExpansion` — post message with a top-level folder attachment expands one level; asserts dispatched URL + expanded XML.

4. **misc.go typed models (Major, 10:56)** — deliberately kept the loose-map decode at this boundary: `DoAPIJSONTyped` returns a decoded `map[string]interface{}` (no raw body to re-unmarshal without a double marshal round-trip), and numeric fields already go through `json.Number` (`dec.UseNumber`) → `numToInt64` so unsupported shapes can't silently coerce wrong counts. Happy to switch to a typed struct if we add a raw-bytes `DoAPI` path — flagging here as a conscious trade-off.

All `go test ./shortcuts/im/...` green.

jackie3927:
## Round-2 review follow-up (`3ca3a676`)

**0) transport.go PRE hijack — already removed.** The comment reviewed `d098767f` (the prefetch commit), which briefly carried the pre-gateway host redirect. It was dropped in `fb777265`; the PR head's `transport.go` is identical to main (verified: `git diff main FETCH_HEAD -- internal/cmdutil/transport.go` is empty). Your concern is correct — thank you for the catch.

**1) Dead `data["code"]` check — removed.** Confirmed your probe: `DoAPIJSONTyped` unwraps the envelope and returns `data`, non-zero envelope codes surface as `err` (BuildAPIError). The re-check branch is deleted; `data == nil` now gets its own `empty data` warning instead of a misleading `<nil>`.

**2) Prefetch failure doubling — fixed with a sentinel.** `PrefetchFolderChildren` now records `""` for "tried and failed" (both the single fast path and the concurrent loop). The folder converter and post attachment renderer treat a `""` cache hit as "degrade to single-line, do not retry inline", so a page-wide failure (missing scope / DLP) issues exactly N prefetch GETs and zero render-loop retries. Tests: `TestPrefetchFolderChildren_FailureSentinel`, `TestFolderConverter_PrefetchFailureSkipsInlineRetry`.

**3) Post folder attachments now prefetched.** `PrefetchFolderChildren` also scans `post` messages' attachment zone (`files[]` with `is_folder`), caching under `message_id + "\x00" + folder_key` (a post may carry several folders). `renderPostAttachments` consults the cache first. Test: `TestPostFolderAttachmentPrefetch`.

**4) Thread-reply folders prefetched.** `ExpandThreadRepliesWithResources` Phase 2a now runs `PrefetchFolderChildren` over the same `allRawReplies` and renders replies through `FormatMessageItemWithFolderPrefetchOpts` (was nil before).

**5) Docs updated** (`skills/lark-im/references/lark-im-messages-mget.md`): folder messages expand one level (root `child_count`, `has_more="true"` past cap 10, nested-folder depth hint, empty folder `child_count="0"`), post folder attachments expand the same way, and the download sentence now covers folder children (verified downloadable in PRE — the `resource_extract.go:78` comment saying otherwise is stale; I can update it too if you prefer).

**Small items:** `folderPrefetchConcurrency` documented with rationale; empty `items` with `all_count > 0` now renders a self-closing `<folder ... has_more="true"/>` instead of an empty open/close pair; PR description rewritten (key-first, cap 10, prefetch, post/thread coverage).

**Open questions for you:** cap-10 escape hatch — no shortcut can list children 11+ today. Plan is a follow-up `+folder-children` shortcut (openapi already exists) — fine to defer out of this PR?

All `go test ./shortcuts/im/...` green.

jackie3927:
## Round-3 review follow-up (`c33f0df2`)

**Main issue — folder children download docs: you're right, fixed.** Your probe matches `extractPostResourceRefs`: it walks the *raw* content's top-level `files[]` (excluding `is_folder`) — folder children only exist after expansion in the render layer, so they never enter the `--download-resources` worklist. (My earlier "downloadable in PRE" claim came from a direct curl against the resources endpoint with a child `file_key`, which returns 200 — but that's not what the CLI command does.) Doc now states plainly: folder children are **browse-only**, not in the download worklist; standalone file messages and top-level post file attachments remain eligible. Making folder children actually downloadable would need the extractor to expand folders too (another download path) — deferring that as a follow-up rather than bolting it on here.

**Other items:**
1. **Godoc mishang** — fixed: `folderCacheKeySep` const doc now precedes the const with a blank line, `PrefetchFolderChildren` got a proper doc comment, and all three stale statements are gone (sentinel `""` semantics, `FormatMessageItemWithFolderPrefetchOpts`, composite message_id + folder-key cache). `merge.go` doc now mentions `FolderPrefetchOpts` too.
2. **Triple root-tag header** — deduped into `writeFolderOpen(b, key, name, childCount, forceChildZero, hasMore, selfClose)`; empty-folder early-return folded into the `shown == 0` branch (empty → `child_count="0"`, else all_count + has_more). Existing output unchanged (full folder test suite green).
3. **text.go double inline arm** — collapsed to a single chain via an `expanded` variable: cache hit → use; `""` sentinel → degrade; cache miss + Runtime → one inline fetch; else single-line. One `fetchFolderChildrenTree` call site now.
4. **Export wrappers** — kept as back-compat; godocs now say they exist for external callers and that internal commands migrated to `FormatMessageItemWithFolderPrefetchOpts`.
5. **Test gaps** — added `TestFetchFolderChildrenTree_NoItemsButAllCount` (self-closing `child_count="3" has_more="true"/>`) and `TestFetchFolderChildrenTree_EmptyData` (`empty data` stderr warning).

**Still open (previous round):**
- **all_count semantics** — first-level (direct) child count, confirmed against the backend folder-children handler (`buildHierarchyItems` renders exactly one level; all_count = that level's item count). `has_more`/root `child_count` are consistent with this.
- **cap-10 escape hatch** — no shortcut lists children 11+ today; planned follow-up `+folder-children`. Deferring out of this PR unless you want it in.
- **PR description** — updated last round to key-first/cap-10/prefetch/post (the round-3 comment may have reviewed the older body).

All `go test ./shortcuts/im/...` green (now 18 folder-related tests).

jackie3927:
## Correction on the folder-children download claim (`ec523660`)

Re-checked the download path after your round-3 probe — your experiment tested `--download-resources` (the **automatic** enrichment worklist), but that's a different mechanism from the manual command, and the doc now separates them:

- `+messages-resources-download` takes an explicit `--message-id` + `--file-key` (no auto-collection) and calls `GET /messages/:id/resources/:file_key` with exactly the key you pass. Folder children rendered inside `<folder>...</folder>` are real files addressed by their own `file_key`, and this endpoint serves them — verified earlier by direct download of a folder child in PRE (HTTP 200). The `resource_extract.go:80` comment ("cannot return one") refers to the **folder itself** (`is_folder` container, which is not a single file), not to its children.
- `--download-resources` only auto-collects **top-level** single-file refs from raw content (standalone file messages, post attachment `files[]` minus `is_folder`). Folder children exist only after render-time expansion, so they are never auto-added — that's why your `resources: [{key: file_real}]` probe shows only `file_real`. An agent wanting a folder child must pass its key explicitly to `+messages-resources-download`.

Doc updated to state this precisely (children downloadable via the explicit command; `--download-resources` auto-path excludes them; folders themselves are not single-file resources).

If you still believe the endpoint rejects child `file_key`s, happy to add a regression e2e once a stable folder message fixture exists in an env both of us can reach — the fixtures I had in PRE are now behind an app/message-visibility issue on my side.

jackie3927:
## Follow-up on the two later comments

Both are addressed (now one squashed commit `5937fdc5`):

1. **Prefetch goroutines write `runtime.IO().ErrOut` lock-free** — fixed. `fetchFolderChildrenTree`'s stderr warnings now go through a `folderWarn` helper guarded by a package mutex (aligning with merge's prefetch, whose warning is written inside its own `mu.Lock()`). The concurrent prefetch tests also now run on the roundtrip runtime with a static token resolver (same pattern as the merge tests) — the earlier `TestFactory`/httpmock setup additionally raced the credential provider's token resolution under `-race`. Local `go test -race ./shortcuts/im/...` is green.

2. **Docs only updated mget; chat-messages-list resource table has no folder row** — fixed. `lark-im-chat-messages-list.md` Resource Rendering table now has a Folder row (marker shape + download contract), and `threads-messages-list`/`messages-search` got matching one-line folder-expansion notes (all four commands render folder messages).

**On CI**: the main `CI` workflow (go test `-race`) hasn't shown up in the checks — for a fork PR it needs a maintainer to approve the workflow run before GitHub creates it (only CodeRabbit/sync-pr-labels are pending right now). Once approved it runs `go test -v -race -count=1` over `./shortcuts/...`, which the current head passes locally.

sammi-bytedance (approved):

liangshuo-1 (approved):

PR #2611, opened 2026-09-03
feat(sheets): accept the payload spellings the 08-29..31 reflow rejected

xiongyuanwen-byted:
## Why

The 08-29..31 reflow report ([doc](https://bytedance.larkoffice.com/docx/FVPJdqfpfonpjuxTwDecCantn9b)) counts 2839 failures across 31388 `lark-cli sheets` calls, 9.04%. A large share of them carry well-formed intent written in a spelling the flag does not document, so the retry rewrites the same data differently and the call succeeds on the second or third try. Each of the shapes accepted here has exactly one reading.

Numbers below are the report's own attribution counts, three platforms combined.

## What now succeeds

**`--sheets` on `+workbook-create` / `+table-put`**

| shape | count |
|---|---|
| numeric column whose cells arrived as JSON strings | 155 |
| row cut short at its last non-empty cell, now padded with nulls | 25 |
| `columns` written as one object per column | 35 |
| `columns` empty when the sheet carries no data either | 15 |
| bare sub-sheet list without the `{"sheets":[...]}` envelope | 10 |

`dtypes` also reads the writer's own type vocabulary (`number`, `date`). That one produced no error at all before: it fell through to the string default and silently wrote figures as text.

**`--styles`**: a quoted number under a numeric field (`"font_size": "16"`, 30), and `border_styles.outer` expanded to the four range edges (4).

**`+cond-format-create` / `+cond-format-update` `--properties`** (about 22 of the 62 attributed to attrs shape): `operator` renames to `compare_type` under the `{value}` / `{text}` shapes; symbol and abbreviation comparisons canonicalize (`>=`, `LT`, `LESS_THAN`); a numeric threshold becomes the string the schema declares; a two-element `between` list joins; a bare `attrs` object becomes the one-entry list. Rules whose own contract spells `operator` (timePeriod, iconSet, aboveAverage) keep it, and dataBar / colorScale numeric thresholds stay numbers.

**`+csv-put`** (19): answers to `--data` / `--content` / `--csv-file`, and a path-shaped `--csv` value naming a real file is read as `@<path>` would, reported in the result's `warnings` rather than silently. Requiring the prefix there was an inconsistency of that flag's own making, since every sibling path-valued flag takes the same value bare. A value that merely collides with a file name, `README.md`, still errors.

**`--font-line`** (8): takes the CSS text-decoration words (`strikethrough`, `strike`, `underlined`).

## What still fails, but with the fix in the message

`+cells-get` rejects an Excel multi-area range client-side and spells out the enclosing rectangle (54 rejections the backend answered with a bare `[90015206] invalid range`). `row_height` / `col_width` / `wrap` / `unmerge_cells` inside a `cell_styles` item now name where they really live, as do `--wrap-text` / `--text-wrap` / `--style` on `+cells-set-style`.

## What deliberately did not change

- Shrinking `--range` to fit a smaller `--cells` (122). `checkCellsMatchRange` documents why the computed range is not applied automatically.
- Chunking a `--styles` spec past the 100-operation cap (48). It would drop the atomicity of a single `batch_update`, and the current error already steers row banding to `+cond-format-create`.
- Merged-region conflicts (193). Server-side rejections whose prescription is already complete.

## Scope and tests

Everything is inside `shortcuts/sheets`. No shared-layer file is touched, per the domain-optimization constraint established in the 2026-07 rounds. Rejections that require guessing at intent still fail: a non-numeric string in a numeric column, a row wider than `columns`, a column-less sheet that does carry rows, `font_size: "large"`.

New tests pin every acceptance and its matching still-rejected counterpart, and the `--styles` prior corpus in `styles_acceptance_test.go` gains rows for the reflow findings. `go test ./shortcuts/...`, `make quality-gate` and the sheets dry-run e2e all pass.

## Summary by CodeRabbit

- **New Features**
  - Added broader flag aliases, command hints, and clearer spreadsheet guidance, including Windows-specific payload instructions.
  - Expanded CSV handling with automatic path-shaped file reading and warnings.
  - Added flexible table imports, metadata support, empty sheets, padding, and numeric/date coercion.
  - Expanded style and conditional-formatting vocabulary and input formats.
  - Added automatic retries for transient read failures and merged-cell recovery guidance.
  - Added multi-request support for large style updates.

- **Bug Fixes**
  - Improved dry-run behavior and cell-write range handling.
  - Corrected style, border, resize, and payload normalization.
  - Added guidance for stale conditional-formatting IDs.


xiongyuanwen-byted:
The report gained two sections after this PR was opened: a 58-row table covering every long-tail command, and a per-command breakdown of the "other" buckets its first pass had only counted. `dbedb3cf` works through them.

**One defect surfaced.** A cell carrying style fields at its top level (`{"value":"x","font_weight":"bold","border":{…}}`) passed every client check and reached the backend verbatim, which answered `[cells[0][0].border] unexpected property "border" is not defined`. 33 rejections, on payloads the `--styles` path accepts. Style fields now fold into the carrier that holds them and their values canonicalize on the same pass, so a boolean `word_wrap` no longer dies one step later on the schema's `expected type "string"`. Unknown keys still pass through.

**Now accepted**

| shape | count |
|---|---|
| `--writes` as `{"writes":[…]}` or a lone write object | 54 |
| `type` as the line-kind slot inside a border spec | 29 |
| a blank column heading | 13 |
| blank text in a date column | 7 |
| cell-style vocabulary in a conditional-format rule's style block | 9 |
| `custom` as a row/col sizing type when the op carries a size | — |

**Now prescribed** (the fix changes the shape, so the error carries it): `--position` on `+dim-delete` (18), a whole-column range where a cell style needs a rectangle (28), `--ranges` on `+cells-unmerge` (7), `--include` on `+csv-get` (3), the sheet selectors on `+styles-put`, `--payload` on `+table-put`, border keys on a styles item, and four invented subcommands (`+cells-get-style`, `+dropdown-list`, `+meta`, `+sheets-list`).

**Now renamed silently** (the command already has the input under another spelling): `--title` on `+workbook-import` (16), `--file` / `--outdir` on `+workbook-export` (15), `--replace` on `+cells-replace` (7), `--output` on `+csv-get`, `--sheet-name` on `+sheet-create`, `--new-title` on `+sheet-rename`.

The Excel multi-area `--range` also moved from `+cells-get`'s own `Validate` to the shared `--range` chain: 5 more landed on `+csv-get` and 4 on `+cells-set-style`, so the habit is not specific to reads. It is chained after the sheet-prefix rewrite, so a sheet name containing a comma cannot be read as several areas.

Still deliberately unchanged, unchanged from the original description: shrinking `--range` to fit `--cells`, chunking a `--styles` spec past the 100-operation cap, and the server-side merged-region conflicts. Network and 5xx failures (~150 across commands) stay untouched too: `RetryTransport` is already in the chain with `MaxRetries` at 0, and turning it on blanket-retries non-idempotent writes.

`go test ./shortcuts/...`, `make quality-gate` and the sheets dry-run e2e pass.

xiongyuanwen-byted:
`9c5fd51f` works through the review. Nine defects, plus the lint / deadcode failure: `payloadStdinForm` was unreachable, since only its `goos` variant is called.

**Silent wrong values.** A `between` threshold list of any length was joined into a comma string, so one or three thresholds became a plausible non-empty value that passed the local shape check and failed at the backend. The literal `"null"` decodes into any destination without an error and leaves it at the zero value — in a numeric column it marshalled back out as `0`, and under `font_size` it became a zero-point font. Both reject it now. `outer` and `all` in one border spec dropped `outer` even when the two named different boxes; a duplicate is still folded away, a conflict reaches the invalid-side check.

**Rejections the acceptance layer meant to absorb.** The multi-area `--range` prescription took its rectangle from the first and last area, so `A3,J3,G3` prescribed `A3:G3` and dropped J3. It now takes the min and max of both axes over every area. `type:"custom"` was rewritten to pixel only when the op carried `size`, so `height` / `width` — the alias the same parser accepts everywhere else — failed the enum check before it was ever resolved.

**False reports.** `+cells-set` reconstructed the stated range without the payload, so every bare `--range A1` with a multi-cell payload reported itself as narrowed. `+table-put --dry-run` skipped the style expansion in append mode, hiding a style-only write that the execute path performs.

**Contract.** `+styles-put` still promised one fail-fast batch request after it gained chunking, and its partial-failure advice called a whole-spec retry safe. Replaying an applied `merge_cells` is rejected as an overlap, so that was wrong for any spec carrying merges. The description, the tips and the warning now say a large spec can partially apply, and that the merge route is to read the sheet back and resend only what did not land.

The windows positional-argument annotation flattened the framework's error to text. It keeps it as the cause, and the annotation is split into its own function so its typed shape is asserted on every host — `chainPositionalArgsCause` installs nothing off windows, which is where CI runs.

**On the test asks.** Two landed differently from the suggestion. The batch sub-op precheck attributes to `--operations`, not `--cells`: that is the flag the caller passed, and the sub-op's own `--cells` is named in the message instead. The unrelated-failure subtest asserts `server_error`, not `invalid_parameters` — `callTool` pins the sheet-ai business code to that subtype.

Every behavior change carries a regression test; the two `+table-put` ones were checked against the pre-fix code and both fail there. `+styles-put`'s chunk test now reads the ordered 100 / 52 operation payloads off the plan and exercises a second-request failure through mocks.

`go test ./shortcuts/...`, `make quality-gate` (exit 0, 40 warnings against a base of 39), `golangci-lint --new-from-rev origin/main` and `deadcode -test` on the package all pass.

shortcuts/sheets/flag_ergonomics.go
```diff
@@ -140,6 +149,18 @@ var commandFlagAliases = map[string]map[string]string{
...
 	// verbatim as --cells — the name was the only thing wrong.
 	"+cells-set": {"values": "cells"},
+	// 08-29..31 reflow, long-tail table. Each of these names an input the
+	// command already has under one other spelling, with identical value
+	// semantics: the import name (16 rejections, all but one on windows),
+	// the export destination (13 + 2), and the replacement text (7). None is
+	// within the did-you-mean budget -- "title" shares no prefix with "name",
+	// and "replace" is 4 edits from "replacement".
+	"+workbook-import": {"title": "name"},
+	"+workbook-export": {"file": "output-path", "outdir": "output-path", "output-dir": "output-path", "output": "output-path"},
```
fangshuyu-768, line 159: **[P2] Preserve directory semantics for --outdir aliases** — `--outdir` and `--output-dir` do not have the same value semantics as `--output-path`: `applyWorkbookOutputPath` treats only an already-existing directory as a directory and otherwise treats the value as a file path. Thus `--outdir ./exports` when `exports` does not exist saves a file literally named `exports` in `.` instead of downloading into that directory. Track these spellings and handle them as directories, or leave them as prescriptions rather than silent aliases.

shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -603,6 +628,20 @@ func guardCSVValueIsNotFilePath(runtime *common.RuntimeContext) error {
...
+			// A path-shaped value naming a real file is a forgotten "@", and
+			// nothing else: the shape test already excludes prose, CJK text
+			// and plain filenames like "README.md" (see csvValueLooksLikePath).
+			// Read it and say so in the result rather than spending a round
+			// trip on the punctuation — the sibling path-valued flags
+			// (+workbook-import --file, this command's own --file alias) take
+			// the same value with no prefix, so requiring one here was an
+			// inconsistency of this flag's own making. 08-29..31 reflow:
+			// 11 of +csv-put's 29 rejections.
+			if csvValueLooksLikePath(raw) {
```
fangshuyu-768: **[P1] Require explicit file intent before reading --csv values** — `--csv` is documented as literal RFC 4180 text, but this branch changes any path-shaped value that happens to name a local file into that file contents. For example, a caller trying to write the literal `./data.csv` will upload the contents of a same-named local file, and the warning arrives only after the remote write. This makes semantics depend on cwd state and can disclose local data. Keep the existing rejection, or require explicit file intent (`@path`/`--file`) before reading.

shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -261,6 +280,7 @@ func cellsSetInput(runtime flagView, token, sheetID, sheetName string) (map[stri
 	if err := checkCellsMatchRange(cells, rangeStr); err != nil {
 		return nil, err
 	}
+	rangeStr, _ = fitCellsRange(cells, rangeStr)
```
fangshuyu-768: **[P2] Surface range narrowing for nested writes** — `cellsSetInput` is also used by `--writes` items and `+batch-update` sub-operations, but the narrowing note is discarded here and only the standalone `CellsSet.Execute` path reconstructs a warning from top-level flags. A nested `{range:"A1:D4", cells:[[...2x2...]]}` therefore ships `A1:B2` with no warning, even though the equivalent standalone call reports the changed extent. Preserve and aggregate the `fitCellsRange` note for nested operations and include it in the success result.

shortcuts/sheets/lark_sheet_object_crud.go
```diff
@@ -1199,3 +1232,293 @@ func filterDeleteInput(runtime flagView, token, sheetID, sheetName string) (map[
...
+				continue
+			}
+		default:
+			continue
+		}
+		existing, _ := style["font"].(string)
+		switch {
+		case existing == "":
+			style["font"] = word
+		case existing != word:
```
fangshuyu-768: **[P2] Do not normalize unknown font values into bold italic** — The `existing != word` branch assumes every other `font` value is the complementary valid style. An input such as `{"font":"normal","font_weight":"bold"}` is silently rewritten to `"bold italic"`, so invalid or contradictory input becomes valid with an italic style the caller never requested. Only combine the recognized `bold` and `italic` values; otherwise preserve the canonical value for schema rejection or report a conflict.

shortcuts/sheets/lark_sheet_object_crud.go
```diff
@@ -258,14 +261,35 @@ func newObjectUpdateShortcut(spec objectCRUDSpec) common.Shortcut {
...
+// like a transport problem rather than "that id belongs to a rule that no
+// longer exists". 08-29..31 reflow: 14 rejections across +cond-format-update
+// and +cond-format-delete, the two largest long-tail entries after the flag
+// renames. Any other failure passes through untouched.
+func annotateStaleObjectID(err error, spec objectCRUDSpec) error {
+	if spec.idFlag == "" {
+		return err
+	}
+	p, ok := errs.ProblemOf(err)
+	if !ok || p.Hint != "" || !strings.Contains(strings.ToLower(p.Message), "not found") {
```
fangshuyu-768: **[P2] Only add the stale-ID hint for object lookup failures** — This matches any not-found message on an ID-addressed update/delete, including backend errors such as `sheet "s" not found`. In that case the command tells the caller to refresh chart or conditional-format IDs even though the failing input is the sheet selector. Gate the annotation on the object type/ID being named, or explicitly exclude sheet/workbook not-found errors.

shortcuts/sheets/lark_sheet_styles_put.go
```diff
@@ -67,12 +79,35 @@ var StylesPut = common.Shortcut{
...
+				// Say what landed before naming the failure: each request is
+				// atomic on its own, so the sheet now carries the earlier
+				// chunks. Re-running the WHOLE spec is only safe when it holds
+				// no merges — a style stamp is idempotent, but replaying an
+				// already-applied merge_cells can come back as an overlap
+				// rejection before the failed chunk is even reached, which
+				// would leave the caller stuck on an error about work that
+				// already succeeded.
+				return attachSheetsWarningsToError(err, []string{fmt.Sprintf(
+					"--styles was sent as %d batch requests and request %d failed; requests 1-%d already applied. Style stamps are idempotent, so a spec of styles/sizes/freeze alone is safe to re-run as-is; if it carries cell_merges, read the sheet back first and resend only the merges that did not land — replaying an applied merge is rejected as an overlap",
```
fangshuyu-768, line 118: **[P3] Handle failure before any style chunk has applied** — When the first request of a multi-chunk spec fails, `i` is zero and this message reports `requests 1-0 already applied`. That is contradictory and gives misleading recovery guidance even though no earlier request landed. Handle `i == 0` separately and report that nothing was applied.

xiongyuanwen-byted:
`0f86a9d0` answers all six. One of them is a revert.

**P1, `--csv`.** You are right, and the leniency is gone. Reading a path-shaped value that happens to name a local file made the flag's meaning depend on the working directory: a caller writing the literal `./data.csv` into a cell uploaded a same-named local file, and the warning only arrived after the remote write. The existing prescription (`@<path>` or stdin) stands on its own again, and the substitution machinery is removed with it. That returns 11 rejections to being rejections, which is the right trade for not turning a cell write into a local-file read.

**`--outdir`.** Kept as an alias, but it now carries directory intent rather than inheriting `--output-path`'s file semantics. Both spellings say directory in their own name, so guessing from disk state was the bug. They ride the alias-provenance channel already built for `+csv-put --file`, and `applyWorkbookOutputPath` honors the flag without probing. `--file` and `--output` stay file-valued, which is what they mean.

**Nested range narrowing.** `fitCellsRange` already produced the note; `cellsSetInput` was throwing it away and the standalone path was reconstructing an approximation from its own flags. The builder returns it now, and all three callers report it: standalone, `--writes[i]`, and `operations[i]`. The reconstruction is deleted rather than kept alongside.

**`font` combining.** Only `bold` and `italic` combine. Anything else under `font` keeps its value for the schema to reject, so `{"font":"normal","font_weight":"bold"}` no longer acquires an italic nobody asked for.

**Stale-ID hint.** Now gated on the message naming the id the command addressed, plus an explicit skip for sheet and workbook not-found. A missing sheet no longer gets told to refresh rule ids.

**P3, first-chunk failure.** That case reports that nothing was applied, and drops the merge caveat, which does not apply to an untouched sheet.

Each fix has a regression test, and I checked all six fail when the fix alone is reverted. `go test ./shortcuts/...`, `make quality-gate` (exit 0, 40 warnings against a base of 39), `golangci-lint --new-from-rev origin/main` and `deadcode -test` on the package all pass.

shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -103,26 +103,39 @@ var CellsSet = common.Shortcut{
 			if err != nil {
 				return err
 			}
-			runtime.Out(out, nil)
+			runtime.Out(appendSheetsWarnings(out, notes), nil)
```
fangshuyu-768, line 111: **[P2] Preserve narrowing notes on write failures** — The narrowing notes are appended only after `callTool` succeeds. A `batch_update` can partially apply before returning an error, and a transport failure can leave a single write outcome unknown, so an applied narrowed write may be reported without its actual `A1:B2` footprint. Attach `notes` to the error in the `--writes` branch, and do the same with `narrowNote` in the standalone branch, using the existing `attachSheetsWarningsToError` helper.

shortcuts/sheets/lark_sheet_object_crud.go
```diff
@@ -1199,3 +1251,311 @@ func filterDeleteInput(runtime flagView, token, sheetID, sheetName string) (map[
...
+		switch {
+		case existing == "":
+			style["font"] = word
+		case existing == word:
+			// Already says it; the flat spelling is redundant.
+		case condFormatFontEnum(existing):
+			// The only two recognized values are bold and italic, so a
+			// different recognized one means both were asked for.
+			style["font"] = "bold italic"
+		default:
```
fangshuyu-768, line 1397: **[P2] Consume aliases already covered by bold italic** — When `font` is already `"bold italic"`, a matching `bold:true`, `italic:true`, or equivalent word alias reaches this default branch and remains in the outgoing style object. The dry-run payload therefore still contains an unsupported alias even though the canonical value already includes that effect. Treat `"bold italic"` as subsuming both recognized words and delete the redundant alias; preserve it only when the canonical `font` value is genuinely unknown.

shortcuts/sheets/lark_sheet_styles_put.go
```diff
@@ -67,12 +79,43 @@ var StylesPut = common.Shortcut{
...
+				// already-applied merge_cells can come back as an overlap
+				// rejection before the failed chunk is even reached, which
+				// would leave the caller stuck on an error about work that
+				// already succeeded.
+				if i == 0 {
+					// Nothing landed: the spec is untouched on the sheet, so
+					// the merge caveat below does not apply and saying
+					// "requests 1-0 already applied" would be nonsense.
+					return attachSheetsWarningsToError(err, []string{fmt.Sprintf(
+						"--styles was sent as %d batch requests and the first one failed, so NOTHING was applied; fix the spec and re-run it whole",
```
fangshuyu-768: **[P2] Do not assume the first batch request applied nothing** — The underlying `batch_update` is fail-fast and non-transactional: `flattenToolErrorMsg` explicitly handles responses such as “N succeeded, M failed”, so the first HTTP request can fail after some of its operations were applied. A transport timeout is ambiguous as well. This branch drops the merge caveat and tells the caller to re-run the whole spec, which can replay an applied `merge_cells` operation and fail before reaching the remaining work. Preserve partial/unknown-state guidance for `i == 0`, and only say nothing applied when the backend explicitly reports zero succeeded.

xiongyuanwen-byted:
`3d06a035`. All three, and the third is a correction to what I did last round.

**Narrowing notes on failure.** Right on both counts: a batch is fail-fast but not transactional, and a transport failure leaves a single write's outcome unknown. Both branches attach the note to the error now, so the footprint travels either way. A caller reconciling against the range they stated rather than the one written is the exact confusion the note exists to prevent.

**`bold italic` subsuming its aliases.** Also right. Last round stopped combining unrecognized values but routed this one to the same branch, so the flat spelling stayed in the outgoing style with no schema field to land in. It is dropped now. A genuinely unknown `font` value still keeps its flat sibling visible, which is the case that has to reach the schema.

**First-request failure.** I overcorrected last round. `i == 0` does not mean an untouched sheet, for the reason you give: operations before the failing one inside that request stay applied, and a timeout settles nothing. Only the backend's own `0 succeeded` earns the "nothing applied" message now, reusing the same test `flattenToolErrorMsg` already applies so the two cannot disagree about what a failed batch left behind. Every other first-request failure keeps the read-back route and the merge caveat, which is what the original text had before I split the case out.

Tests cover both sides of that split (`0 succeeded`, a partial `40 succeeded`, and a timeout), and I checked all three fixes fail when reverted individually. `go test ./shortcuts/...`, `make quality-gate` (exit 0, 40 warnings against a base of 39), `golangci-lint --new-from-rev origin/main` and `deadcode -test` all pass.

fangshuyu-768 (approved):

fangshuyu-768 (approved):
Reviewed the latest fixes at 3d06a035; the previously reported findings are addressed and the relevant checks pass.

PR #2635, opened 2026-09-07
feat(drive): resumable chunked downloads for +download and +pull

### LINKED ISSUE #2324: [drive] Large downloads restart from zero after transient failures; add Range-based resume
## Summary

`lark-cli drive +download` is unreliable for large Drive files in automation, and a transient transfer failure forces the next attempt to restart from byte 0. The Drive download API already supports HTTP `Range`, but the CLI does not expose resume/checkpoint behavior.

This makes a single temporary network/CDN failure very expensive for files around 1 GB or larger.

## Environment

- OS: Windows x64
- lark-cli: 1.0.83
- Authentication: user OAuth (`identity: user`)
- Resource: a regular file inside a Drive folder
- File type: RAR archive
- File size: 1,009,727,142 bytes (~963 MiB)

The real tenant URL, folder token, and file token are intentionally omitted because they belong to an internal corporate tenant. They can be shared privately with maintainers if needed.

## Command

```powershell
lark-cli drive +download `
  --file-token '<REDACTED_FILE_TOKEN>' `
  --output 'logs_20260811-170157.rar' `
  --overwrite `
  --json
```

## Observed behavior

1. The folder can be inspected and listed successfully.
2. File metadata can be queried successfully.
3. During the problematic run, downloading the only file in the folder failed twice. Total elapsed time was about 12 minutes 50 seconds.
4. No reusable partial content remained after failure, so every retry started from byte 0.
5. The temporary output provided no useful progress signal and the automation ultimately only knew that the download failed.
6. Running the same download command against the same token later succeeded and downloaded all 1,009,727,142 bytes in about 255 seconds.

This strongly suggests an intermittent transport/CDN failure rather than an invalid token, missing permission, insufficient disk space, or an unsupported file type.

For archive files, `drive +preview --type source_file` is not a fallback. The preview API returns:

```json
{
  "code": 90003033,
  "message": "file type does not support preview"
}
```

## Impact

In unattended agents and CI-style workflows, retrying a large file from byte 0 wastes substantial time and bandwidth. Multiple transient failures can prevent the downstream analysis from receiving any logs at all.

## Expected behavior / request

Please add resumable large-file downloads to `drive +download`, for example:

- `--continue` / `--resume` option.
- Persist a `.partial` file plus checkpoint metadata.
- Resume with `Range: bytes=<downloaded>-` after validating the remote file identity/size/version.
- Preserve partial data after transport errors or process interruption.
- Bounded retries with exponential backoff.
- Report downloaded bytes, total bytes, speed, and ETA on stderr.
- Preserve and return the underlying transport error in JSON.
- Optionally support configurable chunk size and integrity verification.

The official Drive download API documents partial downloads via the HTTP `Range` header and HTTP 206 responses:

https://open.feishu.cn/document/server-docs/docs/drive-v1/download/download

## Suggested acceptance cases

1. Interrupt a 1+ GB download after several hundred MB, rerun with `--continue`, and verify that only the remaining bytes are requested.
2. Simulate a connection reset and verify automatic bounded resume.
3. Verify that a changed remote file invalidates stale checkpoint data.
4. Verify that a server without Range support safely falls back to a full download.
5. Verify machine-readable progress and the original transport error in JSON output.

wufei-png:
## Summary

`drive +download` (and `+pull`) download each Drive file as a single non-resumable HTTP stream. Any transient transfer failure — observed in the wild as an HTTP/2 `RST_STREAM`/`INTERNAL_ERROR` mid-transfer (see #2324) — discards the partial temp file, and the next attempt restarts from byte 0. For multi-GB files this makes `+download` effectively unusable.

This PR makes drive downloads resumable and chunked:

1. **`extension/download`**: new `Options.StartOffset` resumes a multipart stream from an existing local offset instead of byte 0. `Stream.ContentLength` keeps reporting the remote total while the body reads only the remaining bytes. When `StartOffset > 0`, a server that does not honor `Range` fails instead of falling back to a full response (a full body cannot be spliced onto bytes already on disk).
2. **`extension/fileio` + `localfileio`**: optional `AppendingFileIO.AppendTo` appends a body to a partial file (bytes kept on failure) under the same output-path validation as `Save`.
3. **`drive +download`**: routed through `extension/download` with 64 MiB parts and per-part retries. New `--continue` flag resumes from `<output>.partial`:
   - complete partial → committed directly (no re-download);
   - stale (oversized) partial → discarded and restarted;
   - fresh download → any leftover partial is truncated.
   Progress is printed to stderr every 64 MiB. In `--continue` mode a failed run keeps the partial for a later resume; default behavior is unchanged (failures clean up).
4. **`drive +pull`**: uses the same chunked downloader (per-part retries) while keeping its existing `--if-exists` semantics.
5. **Preflight fix**: `+download` now preflights the `view` action instead of `export`. The `members/auth` API does not support `action=export` for plain uploaded files (`type=file`) and always answers `false`, which blocked every download of such files; `view` correctly reflects whether the download endpoint will accept the caller.

## Verification

- Unit tests: `extension/download` (StartOffset multipart / single-response / fallback rejection), `localfileio` (AppendTo), `shortcuts/drive` (`--continue` resume / commit / stale-restart / discard against a range-aware mock transport).
- All `shortcuts/drive`, `shortcuts/common`, `shortcuts/im`, `extension/download`, `internal/vfs/...` tests pass.
- End-to-end against a real tenant: 200 MB file, interrupted mid-transfer, resumed with `--continue`, final bytes match the source hash exactly.

Fixes #2324

## Summary by CodeRabbit

* **New Features**
  * Added resumable Google Drive downloads with `--continue`, chunked transfers, retries, progress reporting, and atomic completion.
  * Partial downloads are validated against remote file details before resuming, restarting, or finalizing.
  * Added support for appending downloaded content to existing local files.
  * Downloads now use view permissions for regular Drive files.

* **Bug Fixes**
  * Unsupported range requests now produce a clear error instead of silently downloading from the beginning.
  * Unsafe output paths and invalid resume offsets are rejected.
  * Partial files are preserved appropriately when resumable downloads fail.
  * `--continue` requires an explicit output path.

### DIFF
--- a/extension/download/README.md
+++ b/extension/download/README.md
@@ -15,6 +15,11 @@ Choose the representation contract deliberately:
 - `download.ImmutableSource`: permits multipart assembly without an ETag because
   the caller guarantees the source identifier pins the bytes.
 
+When resuming from a local offset, set `Transfer.StartOffset` together with
+the checkpoint's strong `Transfer.ExpectedETag`. The engine sends `If-Range`
+on the first resumed request and rejects a changed representation; the caller
+owns the partial artifact and may discard it before reopening from byte zero.
+
 External commands normally use the higher-level `command.Download`, which wires
 an authenticated OpenAPI transport and invocation-scoped FileIO while retaining
 the same engine options:
--- a/extension/download/download.go
+++ b/extension/download/download.go
@@ -80,6 +80,16 @@ type Options struct {
 	// PartSize is the maximum byte count requested per range. Zero selects
 	// DefaultPartSize.
 	PartSize int64
+	// StartOffset resumes the logical stream from an existing local offset
+	// instead of byte 0. The first range request starts at this offset, so the
+	// server must support Range (HTTP 206); when it does not and StartOffset is
+	// non-zero, Open fails instead of falling back to a full response, because
+	// a full response cannot be spliced onto bytes already written locally.
+	StartOffset int64
+	// ExpectedETag binds the first resumed range request to the representation
+	// identified by an existing checkpoint. The request carries If-Range and
+	// the first response must still report the same strong ETag.
+	ExpectedETag string
 	// MaxResponses bounds the total responses in one logical stream. Zero derives
 	// a bound from the declared object size and PartSize.
 	MaxResponses int
@@ -121,11 +131,14 @@ func Open(ctx context.Context, source Source, opts Options) (*Stream, error) {
 		return openFull(ctx, source.transport, opts, retryWait)
 	}
 
-	firstPart := ByteRange{Start: 0, End: opts.PartSize - 1}
-	resp, err := fetchWithRetry(ctx, source.transport, Request{Range: &firstPart}, opts, retryWait)
+	firstPart := ByteRange{Start: opts.StartOffset, End: opts.StartOffset + opts.PartSize - 1}
+	resp, err := fetchWithRetry(ctx, source.transport, Request{Range: &firstPart, IfRange: opts.ExpectedETag}, opts, retryWait)
 	if err != nil {
 		if !rangeProbeRejected(err) {
 			return nil, err
+		}
+		if opts.StartOffset > 0 {
+			return nil, resumeUnsupportedError(opts.StartOffset)
 		}
 		return openFull(ctx, source.transport, opts, retryWait)
 	}
@@ -135,16 +148,41 @@ func Open(ctx context.Context, source Source, opts Options) (*Stream, error) {
 
 	switch resp.StatusCode {
 	case http.StatusOK:
+		if opts.StartOffset > 0 {
+			resp.Body.Close()
+			return nil, resumeUnsupportedError(opts.StartOffset)
+		}
 		return singleStream(ctx, resp)
 	case http.StatusPartialContent:
 		return openPartial(ctx, source, opts, retryWait, firstPart, resp)
 	case http.StatusBadRequest, http.StatusRequestedRangeNotSatisfiable:
 		resp.Body.Close()
+		if opts.StartOffset > 0 {
+			return nil, resumeUnsupportedError(opts.StartOffset)
+		}
 		return openFull(ctx, source.transport, opts, retryWait)
 	default:
 		resp.Body.Close()
 		return nil, unexpectedStatus(resp.StatusCode)
 	}
+}
+
+// resumeUnsupportedError reports that a resumed download cannot proceed
+// because the server does not honor Range requests, so a full response cannot
+// be spliced onto bytes already written locally.
+func resumeUnsupportedError(offset int64) error {
+	return protocolError(
+		"cannot resume from byte %d: server does not honor Range requests; a full response cannot be spliced onto existing bytes",
+		offset).WithCause(errResumeUnsupported)
+}
+
+var errResumeUnsupported = errors.New("resumed range request is unsupported")
+
+// IsResumeUnsupported reports whether Open could not use the requested local
+// offset because the peer did not honor the resumed Range request. A caller
+// that owns the partial artifact may discard it and retry from byte zero.
+func IsResumeUnsupported(err error) bool {
+	return errors.Is(err, errResumeUnsupported)
 }
 
 func openFull(ctx context.Context, fetch Transport, opts Options, retryWait *retryWaitBudget) (*Stream, error) {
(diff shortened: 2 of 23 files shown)


wufei-png:
Thanks for the thorough review — all comments addressed in `35550e9`, except one that I'd push back on. Details:

**Fixed**

1. **Validator checkpoint (ETag + size)** — `+download` now writes a sidecar `<output>.partial.meta` (remote size + ETag) when a download starts and validates it before resuming: a missing checkpoint, a size change, or an ETag change discards the partial and restarts from byte 0. A resume can no longer splice bytes from two different remote representations. Covered by two new tests (`TestDriveDownloadContinueETagMismatchRestarts`, `TestDriveDownloadContinueMissingCheckpointRestarts`) plus the existing resume/commit/stale scenarios.
2. **DryRun consistency** — the dry-run plan now uses the `view` permission check (new `AddDriveFileViewPermissionDryRun`) matching the executed flow, and the scope-failure warning now reads "view permission check failed".
3. **forbidigo** — added reasoned `//nolint:forbidigo` directives to the `os.Remove`/`os.Rename` calls (FileIO exposes no remove/rename for partial artifacts) and to the local-file `AppendFromReader` open.
4. **Typed error assertions** — `extension/download` resume-rejection and validation tests now assert `errs.ProblemOf` metadata.
5. **Test helper reuse** — the four `--continue` tests share `driveDownloadRangeFactory`; the unused `rangeServingSource` helper was removed.
6. **localfileio tests** — `AppendTo` failure keeps already-written bytes (typed `WriteError`), and unsafe paths surface a typed `PathValidationError` with a wrapped cause.
7. **Docstrings** — added doc comments to the touched helpers.

**Declined (with reason)**

- The request to restore the 429 test assertions (`CategoryNetwork`/`SubtypeNetworkTransport` + the "stop immediate retries" hint). `+download` and `+pull` now transfer through `downloadtransport` with `WithReplaySafe` (the same transport IM uses), so a 429 is surfaced by the client layer as a typed `APIError` (`CategoryAPI`/`SubtypeRateLimit`) with the hint "retry with exponential backoff and jitter". The old "stop immediate retries" hint belonged to the previous single-stream path. The test keeps the meaningful contract: a typed HTTP 429 with an exponential-backoff hint. Happy to revisit if you'd prefer the drive transport to diverge from IM here.

Verified: `gofmt` clean; `shortcuts/drive`, `shortcuts/common`, `shortcuts/im`, `extension/download`, `internal/vfs/...` all pass; end-to-end resume against a real tenant with checkpoint validation and byte-exact output.

wufei-png:
Addressed the other four threads in `b87b54a` (full typed-error contracts in the resume/validation tests, `CategoryAPI`/`SubtypeRateLimit`/retryable assertions for the 429 case, `WriteError` cause preservation, and `internal/vfs` read in the localfileio test).

On the "remaining-content contract" thread (drive_download.go:419), I'd push back: `extdownload.Stream.ContentLength` is the **absolute** remote total, not the count remaining after `StartOffset`. It is taken from the `Content-Range` total field (`openPartial` sets `ContentLength: first.total`); only the body reader is bounded to the remaining bytes via `exactLengthReader(total - StartOffset)`. `TestOpenResumeFromStartOffsetMultipart` pins this contract by asserting `stream.ContentLength == len(payload)` for a resumed stream. So the checkpoint stores the absolute total (which the probe compares against on the next run) and `written != ContentLength - startOffset` compares the appended byte count correctly. Applying the proposed diff would store `startOffset + ContentLength` (e.g. 232144 for a 131072-byte file resumed at 30000) in the checkpoint and reject the valid partial on every later `--continue` run. The resume-then-resume regression path is covered by `TestDriveDownloadContinueResumesPartial`, which resumes from byte 30000 and re-validates the checkpoint on a second run.

wufei-png:
Review follow-up: I squashed the original six incremental PR commits together with the review fixes into one feature commit before pushing. The branch now contains a single cohesive commit: `34e01604 feat(drive): add resumable file downloads`; no separate fix-up commits remain.\n\nThe implementation keeps normal downloads on `FileIO.Save` and gates `--continue` behind a complete resumable backend. Validation passed: `make build`, `make vet`, `make unit-test`, `make fmt-check`, targeted download/drive tests, and current-source dry-run E2E.

wufei-png:
Superseded by #2663; closing this PR to reset stale automated review threads while keeping the final implementation unchanged.

PR #2641, opened 2026-09-07
feat(docs): await asynchronous document creation

SunPeiYang996:
## Summary
Keep `docs +create` synchronous to callers when document creation is promoted to a server-side task. Poll to completion, preserve the existing document response and resource follow-up, and return typed errors without repeating creation. Cancellation remains a network/transport error; timeouts guide callers to create first and append content in smaller batches.

## Changes
- Opt document creation into asynchronous promotion and poll task status with bounded waits and retryable GET backoff.
- Preserve response log IDs and typed failures; never report a processing task as command success. Timeouts recommend creating a document first and appending smaller batches, without exposing task-query recovery.
- Add internal, opt-in timing diagnostics on stderr without changing help, schema, or stdout.

## Test Plan
- [x] Focused document and runtime package tests pass on the development machine.
- [x] Live creation verified across 500–20,000 text blocks, including async completion and exact content checks.
- [x] Document dry-run E2E covers the conditional task query before image/file uploads.
- [x] Cancellation and timeout envelope regressions pass, including batching guidance and removal of task-query recovery.
- [x] Build, vet, formatting, and focused document/runtime/output tests pass.
- Full unit suite: all packages passed except an unrelated home-path symlink assertion in `internal/vfs/localfileio` (`TestPolicy_TildeStillReachesHomeFiles`).
- Full required CI results are tracked in this PR’s checks.

## Related Issues
- None

## Summary by CodeRabbit

- **New Features**
  - Document creation now supports asynchronous processing and monitors tasks until completion.
  - Temporary network and transport errors retry automatically without repeating creation.
  - Local images and other document resources continue uploading and linking after successful creation.
  - Optional diagnostic tracing is available through a debug setting.

- **Bug Fixes**
  - Successful asynchronous creation returns the standard document output format.
  - Timeout, cancellation, expiration, and task failures now provide clearer recovery guidance.

### DIFF
--- a/shortcuts/doc/docs_create_async.go
+++ b/shortcuts/doc/docs_create_async.go
@@ -1,0 +1,199 @@
+// Copyright (c) 2026 Lark Technologies Pte. Ltd.
+// SPDX-License-Identifier: MIT
+
+package doc
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/larksuite/cli/errs"
+	"github.com/larksuite/cli/shortcuts/common"
+)
+
+const (
+	docsCreateAsyncMaxWait             = 10 * time.Minute
+	docsCreateAsyncDefaultPollInterval = 3 * time.Second
+	docsCreateAsyncMinPollInterval     = 100 * time.Millisecond
+	docsCreateAsyncMaxPollInterval     = 10 * time.Second
+)
+
+type docsCreateAsyncTask struct {
+	TaskID      string                      `json:"task_id"`
+	Type        string                      `json:"type"`
+	Status      string                      `json:"status"`
+	Stage       string                      `json:"stage"`
+	PollAfterMS int                         `json:"poll_after_ms"`
+	Result      *docsCreateAsyncTaskResult  `json:"result"`
+	Failure     *docsCreateAsyncTaskFailure `json:"failure"`
+}
+
+type docsCreateAsyncTaskResult struct {
+	CreateDocument string `json:"create_document"`
+}
+
+type docsCreateAsyncTaskFailure struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
+type docsCreateAsyncEnvelope struct {
+	Task *docsCreateAsyncTask `json:"task"`
+}
+
+// waitForDocsCreateAsyncTask preserves the synchronous docs +create contract:
+// async success is projected back into the original document/warnings data
+// shape before permission and local-resource follow-up work runs. If the
+// bounded wait expires, the latest processing task remains a successful,
+// resumable result instead of being mistaken for a completed document.
+func waitForDocsCreateAsyncTask(runtime *common.RuntimeContext, initial map[string]interface{}) (map[string]interface{}, bool, error) {
+	envelope, err := decodeDocsCreateAsyncEnvelope(initial)
+	if err != nil {
+		return nil, false, err
+	}
+	if envelope.Task == nil {
+		return initial, false, nil
+	}
+
+	taskID := strings.TrimSpace(envelope.Task.TaskID)
+	if taskID == "" {
+		return nil, false, errs.NewInternalError(errs.SubtypeInvalidResponse,
+			"document create response included an async task without task_id")
+	}
+
+	deadline := time.Now().Add(docsCreateAsyncMaxWait)
+	latest := initial
+	task := envelope.Task
+	polledOnce := false
+	for {
+		status := strings.ToLower(strings.TrimSpace(task.Status))
+		switch status {
+		case "succeeded":
+			result, decodeErr := decodeDocsCreateTaskResult(task)
+			return result, false, decodeErr
+		case "failed", "expired":
+			return nil, false, docsCreateAsyncFailure(task)
+		case "", "processing":
+			// Continue below. An empty status is treated as processing so a
+			// temporarily sparse response does not trigger a duplicate create.
+		default:
+			return nil, false, errs.NewInternalError(errs.SubtypeInvalidResponse,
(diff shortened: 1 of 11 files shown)


shortcuts/doc/docs_create_v2.go
```diff
@@ -86,43 +88,61 @@ func dryRunCreateV2(_ context.Context, runtime *common.RuntimeContext) *common.D
...
+	trace.event("create_response", docsCreateDebugDetails{LogID: createLogID})
 	if err != nil {
 		return err
 	}
 	if docsAPIOperationFailed(data) {
 		return runtime.OutPartialFailure(data, nil)
 	}
-
+	trace.step("wait_task")
+	data, err = waitForDocsCreateAsyncTask(runtime, data, createLogID, trace)
```
fangshuyu-768, line 121: [P1] Preserve partial-failure semantics after async completion

waitForDocsCreateAsyncTask replaces data with the decoded create_document payload, but docsAPIOperationFailed is only checked on the initial acceptance response. If the task reaches succeeded while the embedded create response contains result="failed", this path proceeds with follow-up work and eventually emits ok:true, whereas the existing synchronous contract returns OutPartialFailure and preserves the warnings. Please apply the same partial-failure check immediately after the wait and add an async regression case.

shortcuts/doc/docs_create_async.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, errs.NewInternalError(errs.SubtypeInvalidResponse,
+			"document create response included an async task without task_id")
+	}
+	defer func() {
+		if err != nil {
+			err = docsCreateAsyncWaitError(err, taskID, logID)
+		}
+	}()
+
+	var delay time.Duration
```
fangshuyu-768, line 85: [P2] Honor the initial poll_after_ms

delay starts at zero, so a processing task returned by the create request is queried immediately even when that response supplies poll_after_ms. The interval is only applied after the first GET. This can violate the server-provided polling cadence and make a newly accepted task hit eventual-consistency or rate-limit failures. Please initialize the first delay from task.PollAfterMS, while still returning immediately for terminal initial states, and cover the timing behavior without adding a real sleep to the test suite.

shortcuts/doc/docs_create_async.go
```diff
@@ -0,0 +1,297 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+func docsCreateAsyncWaitError(err error, taskID, logID string) error {
+	if errors.Is(err, context.DeadlineExceeded) {
+		err = errs.NewNetworkError(errs.SubtypeNetworkTimeout, "timed out waiting for document creation task %s", taskID).WithCause(err)
+	} else if !errs.IsTyped(err) {
```
fangshuyu-768, line 221: [P2] Keep cancellation in the network error category

When the parent context is canceled, context.Canceled reaches this branch as an untyped error and is converted to internal/unknown. That misreports an ordinary interrupted wait as a CLI defect and changes the structured exit category. Other polling flows classify cancellation as network/transport while preserving context.Canceled as the cause. Please add an explicit cancellation branch and strengthen TestDocsCreateAsyncCancellationIsTyped to assert the category and subtype.

shortcuts/doc/docs_create_v2.go
SunPeiYang996: Checked the document creation service contract: content-write failures are returned as errors and become task.status="failed". A succeeded task serializes a fixed create_document payload containing document and warnings only; it has no result="failed" field. Therefore the succeeded + nested failed combination described here is not produced by the current service. Resolving this finding as not applicable to the create contract.

shortcuts/doc/docs_create_async.go
SunPeiYang996: Confirmed that the task query endpoint waits/polls on the server side, so the initial GET can be issued immediately. The CLI applies poll_after_ms when a query still returns processing, and retains backoff for retryable query failures. Keeping the immediate first query is intentional; resolving this finding.

shortcuts/doc/docs_create_async.go
SunPeiYang996: Fixed in a831b1d0. context.Canceled now returns network/transport and preserves the original cause for errors.Is; regression coverage asserts the category, subtype, and non-retryable behavior. Timeout/expiry/interrupted-execution errors now recommend creating a document first and appending smaller content batches, without task-ID/query recovery instructions. Other explicit failures retain their original reason. Focused document/runtime/output tests, build, vet, and formatting pass; the full suite has one unrelated home-path symlink assertion failure, recorded in the PR description.

shortcuts/doc/docs_create_async.go
```diff
@@ -0,0 +1,298 @@
...
+		case "succeeded":
+			return decodeDocsCreateTaskResult(task)
+		case "failed", "expired":
+			return nil, docsCreateAsyncFailure(task, logID)
+		case "", "processing":
+			// Continue below. An empty status is treated as processing so a
+			// temporarily sparse response does not trigger a duplicate create.
+		default:
+			return nil, errs.NewInternalError(errs.SubtypeInvalidResponse,
+				"document creation returned unsupported status %q", task.Status)
```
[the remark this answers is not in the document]
SunPeiYang996: Fixed in 0c1dafb7. Added TestDocsCreateAsyncUnsupportedStatus through the docs +create command, asserting internal/invalid_response, the create response log ID, no task recovery text, and empty stdout. Verified on the development machine that deleting the unsupported-status branch makes this test fail, then restored the branch and passed the async and hint-command tests. make vet and make fmt-check pass; go mod tidy leaves module files unchanged. The full unit suite still has the previously recorded, unrelated TestPolicy_TildeStillReachesHomeFiles home-path symlink assertion failure. Remaining CI gates are pending.

fangshuyu-768 (approved):

PR #2642, opened 2026-09-07
fix(base): verify null numeric record-list cells

AiniTang:
A rare matrix response from `base +record-list` can contain `null` for a populated Number/Currency cell while the record batch-get endpoint returns the stored value. Because the command currently accepts the matrix response as-is, downstream totals can be silently understated.

This change treats null numeric cells as ambiguous and verifies only those cells through `records/batch_get`. It reconciles by stable field ID, keeps confirmed legal nulls unchanged, and returns a typed `invalid_response` error when verification omits a record/field, reports `record_not_found`, or returns a conflicting revision. Verification is chunked to the API's 200-record and 100-field limits. Responses that do not expose revisions remain compatible; when both endpoints expose revisions, they must match.

Validation:
- `go test ./shortcuts/base/...`
- `make fmt-check`
- `make vet`
- `make unit-test`
- `QUALITY_GATE_CHANGED_FROM=origin/main make quality-gate`
- Live read-only smoke test against a 357-row table: both pages returned the same revision and a known populated currency cell returned its stored value

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `record-list` reliability by recovering missing numeric and currency values that may appear as `null`.
  * Preserved legitimate `null` values while validating and restoring silently missing number cells.
  * Added safeguards for incomplete, malformed, or inconsistent record responses.
  * Improved handling of large result sets by validating and repairing missing values in batches.

### DIFF
--- a/shortcuts/base/base_execute_test.go
+++ b/shortcuts/base/base_execute_test.go
@@ -2853,7 +2853,7 @@ func TestBaseRecordExecuteReadCreateDelete(t *testing.T) {
 			t.Fatalf("stdout=%s", got)
 		}
 		body := string(batchStub.CapturedBody)
-		if !strings.Contains(body, `"record_id_list":["rec_1"]`) || !strings.Contains(body, `"select_fields":["Cost"]`) {
+		if !strings.Contains(body, `"record_id_list":["rec_1"]`) || !strings.Contains(body, `"select_fields":["fld_cost"]`) {
 			t.Fatalf("batch_get body=%s", body)
 		}
 	})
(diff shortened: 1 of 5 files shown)


AiniTang:
Addressed both review findings in `a2c0053`:

- Currency nulls now follow the same batch-get verification path as Number nulls, with a regression test for value repair.
- A matrix with all required keys but malformed content now returns the established typed `internal/invalid_response`; only responses missing legacy keys keep the compatibility fallback.

Revalidated with focused Base tests, fmt, vet, the full race-enabled unit suite, quality gate, and a live read-only 357-row smoke test.

PR #2676, opened 2026-09-09
fix(wiki): migrate mutation lookups to node_by_token

liujinkun2025:
## Summary
Migrate the internal node lookups used by Wiki node creation, node-mode movement, and deletion from `get_node` to `node_by_token`. Resolve accepted Wiki node tokens and document object tokens to the identifiers required by each mutation, while retaining space checks and protecting shortcut targets from accidental deletion.

## Changes
- Use a small Wiki-local lookup helper for parent, source, and deletion lookups. Other domains and document-to-Wiki movement are unchanged.
- Create under the resolved parent node; move the resolved source node to the resolved parent. Explicit source and target spaces are checked against lookup results.
- Delete with the resolved Wiki node token or document object token according to `--obj-type`. Reject document-type deletion of a shortcut so its original document is not deleted. Declare both `wiki:node:create` and `wiki:node:retrieve` for deletion and its mandatory lookup. Preserve confirmation, child handling, and async completion behavior.
- Share the existing `node_by_token` error classifications with mutation callers, preserving node-get recovery hints and upstream error metadata.
- Update dry-run previews and Wiki guidance, and add request/error regression tests plus a self-contained live workflow with independent readback and shortcut protection checks.

## Test Plan
- [x] User/bot scope-preflight regressions: either missing scope stops before API requests; both scopes permit lookup and deletion.
- [x] Wiki package tests, including `go test ./shortcuts/wiki -race -count=1`.
- [x] Focused create/mutation dry-run E2E tests; live E2E package compile check.
- [x] User-identity live verification with isolated fixtures: Wiki/object tokens with inferred/explicit spaces, old/new lookup parity, create/move/delete readback, shortcut protection, child retention/cascade deletion, invalid inputs, and cleanup verification.
- [x] Temporarily reverted seven critical behaviors individually and confirmed the regression tests failed; restored all temporary mutations.
- [x] Canonical build, vet, formatting, skill format check, incremental golangci-lint, and committed-diff source guards.
- [x] `QUALITY_GATE_CHANGED_FROM=origin/main make quality-gate` after committing.
- [x] `go test -C lint ./... -count=1`.
- [x] `go mod tidy` leaves module files unchanged; dependency license check passes.

## Related Issues
- Follow-up to #2665.

## Summary by CodeRabbit

- **New Features**
  - Wiki node creation, moving, and deletion accept both Wiki node tokens and document object tokens.
  - Commands resolve tokens and use the appropriate node or document identifier.
  - Dry-run previews show node-resolution steps and resolved-token placeholders.

- **Bug Fixes**
  - Added validation for mismatched spaces, object types, and missing resolution data.
  - Lookup failures are classified consistently and prevent mutation requests.

- **Documentation**
  - Updated guidance on supported tokens, validation behavior, and required access scopes.

### DIFF
--- a/shortcuts/wiki/wiki_node_delete_test.go
+++ b/shortcuts/wiki/wiki_node_delete_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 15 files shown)


shortcuts/wiki/wiki_node_delete_test.go
```diff
@@ -23,6 +23,13 @@ import (
 	"github.com/larksuite/cli/shortcuts/common"
 )
 
+func TestWikiNodeDeleteDeclaredScopes(t *testing.T) {
+	t.Parallel()
+	if len(WikiNodeDelete.Scopes) != 1 || WikiNodeDelete.Scopes[0] != "wiki:node:create" {
```
fangshuyu-768: [P1] Declare the lookup scope for node deletion

+node-delete now always calls /wiki/v2/spaces/node_by_token, even when --space-id is supplied, but this test freezes the shortcut contract to only wiki:node:create. The existing +node-get wrapper declares wiki:node:retrieve for this same endpoint, and the shortcut runner treats declared scopes as authoritative for both preflight and permission recovery. A least-privileged token granted only the advertised delete scope can therefore pass preflight and then fail at the lookup before deletion, while the resulting error recommends only the already-granted scope. Add the lookup scope to WikiNodeDelete.Scopes and update this test and reference.

fangshuyu-768 (approved):

fangshuyu-768 (approved):

PR #2685, opened 2026-09-10
fix(im): label unavailable deleted message content

yifei-ambr:
## Summary
A recalled rich-text message can have a body that is no longer valid JSON. The formatter currently reports this as an invalid rich-text payload even though the deleted flag already identifies the actual state.

## Changes
- Render unavailable content as [deleted] when the message is marked deleted and its body is invalid or null
- Preserve existing conversion for active messages and deleted messages whose bodies remain valid
- Add focused regression coverage for invalid and null recalled content

## Test Plan
- [x] Unit tests pass: make unit-test
- [x] Static checks pass: make vet and make fmt-check
- [x] IM package tests pass: go test ./shortcuts/im/... -count=1
- [x] Quality gate passes: QUALITY_GATE_CHANGED_FROM=origin/main make quality-gate
- [x] Manual local verification confirms im +messages-mget renders a recalled post as [deleted] while preserving a neighboring active post

## Related Issues
- None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of deleted messages when their content is unavailable, invalid, or missing.
  * Displays `[deleted]` instead of exposing malformed or unavailable message content.

### DIFF
--- a/shortcuts/im/convert_lib/content_convert.go
+++ b/shortcuts/im/convert_lib/content_convert.go
@@ -188,16 +188,25 @@ func formatMessageItem(m map[string]interface{}, runtime *common.RuntimeContext,
 	rawContent := ""
 	if body, ok := m["body"].(map[string]interface{}); ok {
 		rawContent, _ = body["content"].(string)
-		content = ConvertBodyContent(msgType, &ConvertContext{
-			RawContent:           rawContent,
-			MentionMap:           BuildMentionKeyMap(mentions),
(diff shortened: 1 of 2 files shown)


yifei-ambr:
@sammi-bytedance The CLA is now signed, the review finding has been addressed, and the updated diff passes the repository quality gate and IM tests. Would you mind reviewing this focused IM formatter fix when convenient? Thank you.

PR #2687, opened 2026-09-10
Add mail sender allowlist and blocklist commands

infeng:
Add `mail +sender-list`, `+sender-get`, `+sender-set`, and `+sender-delete` to manage the current user’s sender allowlist and blocklist using the existing mail API resources. All four commands require a user access token and target the `me` mailbox.

List and get follow all pages, label records as `allow` or `block`, and use `mail:user_mailbox.message:readonly`. Get matches an exact sender after trimming whitespace and ignoring case, reports `found=false` only after both lists finish successfully, and reports conflicting membership explicitly. Set and delete use `mail:user_mailbox.message:modify`; server `failed_items` produce a typed error instead of a successful result. Delete requires confirmation. The change also adds command discovery metadata, dry-run support, and usage documentation.

Validation: 31 sender command test cases with mocked API responses passed, together with the affected affordance checks. Coverage includes pagination, exact lookup and misses, conflicting membership, write payloads and business failures, scope metadata, permission errors, bot rejection, validation, and deletion confirmation. No end-to-end testing was performed.

## Summary by CodeRabbit

- **New Features**
  - Added user-level mail sender allowlist and blocklist management.
  - Added commands to list, query, configure, and remove sender addresses.
  - Supports pagination, exact matching, dry runs, confirmations, and detailed error reporting.

- **Documentation**
  - Added usage guidance, permissions, workflows, API resources, and command references.

- **Tests**
  - Added coverage for pagination, validation, authorization, dry-run behavior, conflicts, and error handling.


infeng:
Debug/smoke test artifact from agentic-debug verification run — closing, not a real change.

PR #2710, opened 2026-09-11
feat(sheets): accept the parameter spellings two reflow windows say callers write

xiongyuanwen-byted:
Two datasets drive this branch. The 09-04..07 error reflow (1046 fingerprint buckets, 343031 estimated failed_pv) says which payload shapes the domain rejects; the 09-01..08 flag backflow (1092 command/flag pairs, 71748 rows) says which flag names callers actually write. Everything here was decided by replaying the case against a real binary rather than by reading the message text, which is also how the "already fixed" half was separated out: PR #2611 merged seven minutes after v1.0.94 was cut, so 21.8% of the reflow was fixed on main and unreleased at the time the data was collected.

The contract is the one `style_vocab.go` already states: one canonical form documented, a wide acceptance layer undocumented, and no rewrite whose meaning is in doubt. Every shape below has exactly one reading; the shapes that need a guess are listed at the end and still fail.

### Payload vocabulary

- **`border_type` and the border family** (46035 rejections, the largest single style field). It carried two vocabularies at once — the OpenAPI's side selector and the CSS-ish line word — which read as ambiguous from the field name and is decidable from the value. The reading that blocked it was also backwards: `set_cell_range` applies each side to every cell of the range, verified on a live sheet, so `FULL_BORDER` is exactly `border_styles.all`. The values that address a subset of a range's edges still have no expression here and now prescribe which subset they mean. `border_width`, `border_all` and the one-value shorthand (`border:"solid"`, `border:2`) join them.
- **openpyxl / xlsxwriter font flags** — `bold`, `italic`, `underline`, `font_name`. The concept is in the key and the on/off in the value, so each carries its own value reading and an unreadable one keeps its prescription.
- **`--styles` items** (3938) — the sheet selector answers to `sheet_name`, a section to the effect it has (`row_heights`, `column_widths`, `merges`) and to the singular; a section written as one object lifts into its one-entry list; the bare item list needs no envelope.
- **Column objects** (2559) — the heading is read from `title` / `header` / `label` / `column` / `field` / `key`.
- **`--properties` on cond-format** — the comparison is read from whichever key carries it (`operator`, `criteria`, `comparison`, …) when the value is a compare word, which is what keeps `timePeriod`'s own `operator` intact; the `font` slot folds an object, a list or a reordered string onto its enum.
- **Repeated column names**, and a `dtypes` key that differs from its column only in spacing or casing.

### Shapes

- **`--cells` is sized from its payload**, both directions. The command overwrites by default and a bare single-cell range has always been an anchor, so enforcing the extent in one direction only was the least defensible of the three available positions. The note that already reported narrowing now reports both and names the shipped range. `--start-cell` answers as the anchor spelling its CSV sibling already uses.
- **A row longer than `columns`** loses its trailing empty cells first, and what is still longer gets blank headings rather than invented names. **A sheet with no columns** takes the `--values` treatment: positional type-less columns, no header row, exactly the block that arrived.
- **A whole-column or whole-row style range** is closed against the sheet's own grid on the execute path, which is where the grid is readable.
- **A payload wider or taller than a new sheet's grid** is not refused. 200 columns and 50000 rows are what the create call accepts, not a limit on what a sheet may hold: column 201 and column 260 written into a 200-column sheet both landed and read back, growing its grid to `A1:IZ200`, and row 60000 grew a 20-column sheet the same way. The one refusal the backend does issue counts cells rather than rows or columns (`exceeded the maximum number of cells per sheet`, at 50001×260), and the cell budget already bounds that a million cells below it. Three commits here added local ceiling checks before that was probed; they have been taken back out, and `sheetCreateDims` keeps the two numbers as what they are — clamps on the create call.
- **A missing sheet selector** is resolved against the workbook: one sheet is the sheet meant, several are listed by name. This moved where the requirement is settled — `resolveSheetSelector` now decides only the mutual exclusion, because it runs in Validate, offline, long before anything could look a sheet up.

### JSON and flag names

- **A payload written in another JSON convention** is read on a fallback parse: a string that lost its quotes (4939 rejections — the parser reports the run's first byte, and CJK leads with 0xE4..0xE9, so nothing is wrong with the encoding), single quotes, Python literals, a trailing comma, an unquoted key. A bracket that does not match and a payload cut short are refused, and now say at which byte.
- **Flag spellings fold** by letters and digits alone, so `--sheetName`, `--sheet_Name` and `--sheet.name` are one name — and only ever onto a flag that exists.
- **A flag that exists on a sibling** is named ("`--csv` is a flag of +csv-put"), ranked above an edit-distance neighbour that is not a near-typo.
- **Domain-wide renames** where the surface spells a locator in full and the caller writes the short form: `--sheet` 765 rows over 21 commands, `--spreadsheet` 541 over 17, `--ranges` 525 over 18, the output family 450 over 19. Three more carry a value that changes shape: a bare A1 range into `--ranges`, a bare line word into `--border-styles`, a lone scalar into `--cells`.
- **A defensive `--yes`** on a command that asks for no confirmation is accepted and ignored (437 rows). The framework reads it only on high-risk writes, so the request was already honoured.
- **Missing required flags** keep cobra's wording and gain what the flag takes plus the command's own example. `+cond-format-create` stops demanding `--ranges` / `--rule-type` from a caller who wrote the whole rule in `--properties`.

### What still fails, deliberately

A bracket that does not match, a payload cut short, and a digit-led bare run (`123中文` is equally one string that lost its quotes and two cells that lost their comma). A row wider than its columns where the extra values are real. A sheet that asks for a header row over columns nobody named. `OUTER_BORDER` and the interior-edge values. `--dry-run` keeps every rejection that a real run resolves by reading the workbook, because a preview sends nothing.

### Measurements

Reflow buckets, classified by message pattern (±a few points at the boundaries): fixed 62.7% of 343031 failed_pv, of which 21.8% was #2611 awaiting release; 6.0% now carries a better message while still failing; 20.3% is blocked on two things outside this repo; 9.5% remains.

Flag pairs, probed against the binary before and after over the 275 pairs carrying 97% of the rows: parse silently 29.0% → 36.8%, answered by a curated prescription 50.6% → 52.4%, left with nothing but a did-you-mean 11.8% → 2.7%.

### Verification

`make vet fmt-check`, `make quality-gate` and `go test ./cmd/... ./internal/... ./shortcuts/... ./extension/...` are clean apart from `TestPolicy_TildeStillReachesHomeFiles`, which fails on this machine before the branch (the home path resolves through a symlink here).

Beyond unit tests, every change was run end to end against spreadsheets created for the purpose, reading the cells back to check the values and the shapes rather than the return code: 56 cases across eight of the ten commits, plus a separate pass for the flag aliases. One case failed and was a test-ordering artifact worth recording — a workbook that earlier cases had grown to several sheets can no longer exercise single-sheet selector resolution.

### Follow-ups

The path allowlist rejects 11.2% of the reflow, 66% of it from two agent-runtime session directories where user attachments land. That policy is shared with drive and im, so it is written up for its owner rather than changed here.

`--writes` and `--cells` flag descriptions, and one sentence of the skill text, now describe behaviour that changed; they live in the spec repo and follow separately.

## Summary by CodeRabbit

- **New Features**
  - Added automatic sheet selection for single-sheet workbooks.
  - Added intuitive flag aliases, spelling variants, and clearer guidance for invalid or missing options.
  - Expanded support for loose JSON, shorthand styles, border settings, flexible table formats, and cell data.
  - Cell writes now fit ranges to payload size and support default sheet selectors.
  - Added whole-row and whole-column style ranges.

- **Bug Fixes**
  - Improved validation and error messages for malformed payloads and conflicting sheet selectors.
  - Improved handling of uneven cell rows and flexible table inputs.


shortcuts/sheets/json_repair.go
```diff
@@ -0,0 +1,262 @@
...
+			return i, true
+		}
+	}
+	return 0, false
+}
+
+// unescapeSingleQuoted turns the body of a single-quoted string into the text
+// it denotes, so strconv.Quote can re-escape it for JSON. Only \' differs from
+// the JSON reading; every other escape is left for Quote to render verbatim.
+func unescapeSingleQuoted(body []rune) string {
```
zhengzhijiej-tech: [P2] Preserve escape semantics inside single-quoted strings. This helper only consumes `\'`; sequences such as `\n`, `\t`, and `\\` are passed to `strconv.Quote`, so a Python/JavaScript-style string like `'a\n'` is repaired into a JSON string containing the two literal characters `\` and `n`, rather than the newline denoted by the source spelling. Because this fallback advertises support for Python/JavaScript quoting, that silently changes cell data. Please either decode the supported escape set before quoting or reject escapes whose semantics are not preserved.

shortcuts/sheets/json_repair.go
```diff
@@ -0,0 +1,262 @@
...
+	if first >= '0' && first <= '9' || first == '-' || first == '+' || first == '.' {
+		return false
+	}
+	return !strings.ContainsAny(token, `"'\`)
+}
+
+// trimTrailingComma removes a comma already written to b when only whitespace
+// separates it from the bracket about to be written, reporting whether it did.
+func trimTrailingComma(b *strings.Builder) bool {
+	trimmed := strings.TrimRight(b.String(), " \t\r\n")
```
zhengzhijiej-tech: [P1] Avoid rebuilding the full output prefix for every trailing comma. `b.String()`, `TrimRight`, `Reset`, and `WriteString` copy nearly all bytes accumulated so far. A deeply nested payload with a trailing comma at each level therefore repeats an increasingly large copy and can degrade to O(N^2) time and allocation. Please use a slice-backed buffer that can be truncated, or track the last non-whitespace position without materializing the builder each time.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -897,11 +897,22 @@ func parseValuesRows(runtime flagView) ([][]interface{}, error) {
 	if raw == "" {
 		return nil, nil
 	}
+	// --values decodes here rather than through parseJSONFlag, so it takes the
+	// same loose-JSON repair; strict input never reaches it.
+	if !json.Valid([]byte(raw)) {
```
zhengzhijiej-tech: [P1] Avoid the unconditional extra scan and allocation for valid `--values` payloads. `json.Valid([]byte(raw))` copies and scans the entire input, after which `json.Decoder` scans it again and builds the decoded value. These payloads can be large, so the normal strict-JSON path pays roughly twice the parsing bandwidth plus the byte-slice allocation. Please decode once first and invoke loose repair only after the strict decode fails, while preserving `UseNumber` and the EOF check.

shortcuts/sheets/flag_ergonomics.go
```diff
@@ -872,3 +1139,189 @@ func chainEnumNormalization(cmd *cobra.Command) {
...
+			return nil
+		}
+		names := make([]string, 0, len(missing))
+		for _, f := range missing {
+			names = append(names, strconv.Quote(f.Name))
+		}
+		msg := fmt.Sprintf("required flag(s) %s not set", strings.Join(names, ", "))
+		verr := common.ValidationErrorf("%s", msg)
+		for _, f := range missing {
+			verr = verr.WithParam("--" + f.Name)
```
zhengzhijiej-tech: [P1] Preserve every missing flag in the structured error metadata. `WithParam` assigns the scalar `Param` field, so this loop overwrites the previous value and leaves only the final missing flag even though the message names them all. Agents consuming the error envelope cannot recover the complete set. Build `errs.InvalidParam` entries and attach them once with `WithParams` (or otherwise append to `Params`) so all missing flags survive serialization.

zhengzhijiej-tech (commented):
Reviewed the current head (`fd20910`). I left seven non-duplicate inline findings below. I did not repeat the existing threads covering the three chart selector preflight regressions and the conditional-format font normalization issue. Two other reported concerns were not reproducible on this head: current aliases with the same squashed spelling resolve to the same target, and camelCase sibling-flag lookup succeeds through the sibling command's normalizer.

shortcuts/sheets/flag_ergonomics.go
```diff
@@ -872,3 +1139,189 @@ func chainEnumNormalization(cmd *cobra.Command) {
...
+	// A cut inside a parenthetical leaves an open bracket; drop it with the
+	// half-sentence it opened rather than printing "(A1 notation, e.g".
+	if strings.Count(usage, "(") > strings.Count(usage, ")") {
+		if i := strings.LastIndex(usage, "("); i > 0 {
+			usage = usage[:i]
+		}
+	}
+	usage = strings.TrimRight(strings.TrimSpace(usage), ".,;")
+	if len(usage) > 140 {
+		usage = strings.TrimSpace(usage[:140]) + "…"
```
zhengzhijiej-tech: [P2] Truncate usage text on a rune boundary. `len` and `usage[:140]` operate on bytes, so a description containing multibyte text can be cut in the middle of a UTF-8 sequence and emit an invalid hint. Please truncate by runes or adjust the byte boundary with `utf8.RuneStart` before appending the ellipsis.

shortcuts/sheets/helpers.go
```diff
@@ -894,3 +1004,121 @@ func requireJSONArray(runtime flagView, name string) ([]interface{}, error) {
...
+// The lookup costs one request and only on the path that was going to fail
+// anyway. Validate and DryRun keep the offline resolver: dry-run prints the
+// request it would send without sending anything, and reaching the network to
+// fill in a flag would break that.
+func resolveSheetSelectorExec(ctx context.Context, runtime *common.RuntimeContext, token string) (sheetID, sheetName string, err error) {
+	sheetID, sheetName, err = resolveSheetSelector(runtime)
+	if err != nil || sheetID != "" || sheetName != "" || strings.TrimSpace(token) == "" {
+		return sheetID, sheetName, err
+	}
+	names, listErr := workbookSheetNames(ctx, runtime, token)
```
zhengzhijiej-tech, line 1044: [P2] Preserve the workbook-structure failure instead of converting it into a selector validation error. Authentication, permission, network, and not-found failures all enter `listErr != nil` below and are reported as if the caller merely omitted `--sheet-id` / `--sheet-name`; adding a selector will not fix those cases and hides the retryability and error code from automation. Please return or wrap `listErr`, and reserve the missing-selector error for a successful structure read that cannot identify a sheet.

shortcuts/sheets/lark_sheet_workbook.go
```diff
@@ -1057,14 +1068,22 @@ const (
...
 func parseWorkbookCreateStylesItems(v interface{}) ([]map[string]interface{}, error) {
-	root, ok := v.(map[string]interface{})
-	if !ok {
+	// A bare list at the top level can only be the items the envelope would
+	// have held — the same reading --sheets takes of its own bare list, and
+	// the two flags travel together in one call often enough that accepting
+	// it on one and not the other is its own trap.
+	var rawItems interface{}
+	switch shaped := v.(type) {
+	case []interface{}:
```
zhengzhijiej-tech, line 1091: [P2] Keep the type error consistent with the newly accepted root shapes. This branch now accepts a top-level array, but the `default` error below still says `--styles` must be an object shaped as `{"styles":[...]}`. Invalid scalar input therefore receives a false description of the supported contract. Please state that either a bare array or the object envelope is accepted.

chendaxin-tk (commented):
Static follow-up review. I skipped findings already raised in existing threads (including the sheet-delete target inference, ragged-table panic, selector lookup error masking, fuzzy dtype collision, grid-error masking, and single-quote escape corruption). No live-resource or end-to-end tests were run.

shortcuts/sheets/lark_sheet_write_cells.go
```diff
@@ -1002,80 +1018,111 @@ func dropdownHighlightWarnings(runtime flagView) []string {
...
+// payload is their own data and the command overwrites by default, but that is
+// the sentence the note has to say out loud.
 func fitCellsRange(cells []interface{}, rangeStr string) (string, string) {
 	target, err := parseCellRange(rangeStr)
 	if err != nil {
 		return rangeStr, ""
 	}
 	rows, cols, ok := cellsExtent(cells)
-	if !ok || rows > target.rows || cols > target.cols || (rows == target.rows && cols == target.cols) {
+	if !ok || (rows == target.rows && cols == target.cols) {
```
chendaxin-tk, line 1118: [P1] Preserve an explicitly bounded range instead of widening it to the payload. With `--range A1:A1 --cells '[[1,2],[3,4]]'`, this path sends `A1:B2` under the default overwrite behavior, modifying B1, A2, and B2 even though the caller explicitly selected only A1:A1. The warning is returned after the request, so it cannot protect those cells. Bare `A1` can retain anchor semantics, but an explicit `A1:A1`/`A1:B2` extent should either match the payload or fail before writing.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -559,9 +640,12 @@ func (in *tableSheetIn) normalize(idx int) (tableSheetSpec, error) {
 			spec.Columns[j] = tableColumnSpec{Name: name, Type: "string", Format: "@"}
 			continue
 		}
-		if seenCol[name] {
-			return tableSheetSpec{}, common.ValidationErrorf("--sheets[%d] %q: duplicate column name %q", idx, in.Name, name)
-		}
+		// A repeated heading is a real table shape, not a mistake: a sheet is
```
chendaxin-tk, line 654: [P1] Keep positional dtype/format metadata positional when headings repeat. The newly accepted `columns:["a","a"]` can carry `dtypes:["string","float64"]`, but `columnLabels.resolve` converts that array to a name-keyed map, where the second `a` overwrites the first. Both columns then become numeric, so a first value such as `"001"` is coerced or rejected instead of remaining text. Either retain metadata by column index or reject distinct positional metadata for duplicate names.

shortcuts/sheets/helpers.go
```diff
@@ -894,3 +1004,121 @@ func requireJSONArray(runtime flagView, name string) ([]interface{}, error) {
...
+func workbookSheetNames(ctx context.Context, runtime *common.RuntimeContext, token string) ([]string, error) {
+	out, err := callTool(ctx, runtime, token, ToolKindRead, "get_workbook_structure", map[string]interface{}{
+		"excel_id": token,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var names []string
+	for _, entry := range sheetEntriesFromStructure(out) {
+		if name, _ := entry["sheet_name"].(string); strings.TrimSpace(name) != "" {
```
chendaxin-tk: [P1] Read the established `title` field as a fallback here. The real/tested `get_workbook_structure` shape is `[{"sheet_id":"sh1","title":"Sheet1"}]`; `lookupSheetIndex` and `tableGetSheetMeta` already support both `sheet_name` and `title`. This adapter currently returns no names for that normal response, so omitted-selector resolution fails even for a sole sheet. `workbookSheetGrids` has the same omission at line 1057, which also prevents whole-row/whole-column style ranges from being bounded.

shortcuts/sheets/flag_ergonomics.go
```diff
@@ -277,20 +421,65 @@ var intuitiveFlagHints = map[string]map[string]string{
...
 		target, ok := usable[name]
 		if !ok {
+			// The separator-and-casing fold: --sheetName, --sheet_Name and
+			// --sheet.name are one spelling of --sheet-name, and the wire
+			// vocabulary (camelCase JSON keys) is where every one of them
+			// comes from. Only a fold onto a name this command actually
+			// carries applies, so a flag that does not exist still reaches
+			// the unknown-flag error with its did-you-mean rather than being
+			// renamed into something the caller never asked for.
+			if folded := squashed[squashFlagName(name)]; folded != "" && folded != name {
```
chendaxin-tk: [P1] Stage alias provenance when this fold resolves a semantic alias. `--csvFile ./data.csv` reaches canonical `csv` through the squashed index, but this early return skips `tracker.pending`, so `resolveCSVPathFromFileAlias` treats the path as inline CSV instead of reading the file. Likewise `--outDir ./newdir` loses directory semantics and may create a file named `newdir`. Resolve the folded spelling back through `usable` (or otherwise record its source) before returning the canonical target.

shortcuts/sheets/lark_sheet_object_crud.go
```diff
@@ -374,7 +374,7 @@ func newObjectDeleteShortcut(spec objectCRUDSpec) common.Shortcut {
 			if err != nil {
 				return err
 			}
-			sheetID, sheetName, err := resolveSheetSelector(runtime)
+			sheetID, sheetName, err := resolveSheetSelectorExec(ctx, runtime, token)
```
chendaxin-tk, line 377: [P1] This execution-time resolver is unreachable for the omitted-selector case. The factory's `Validate` callback still calls `objectDeleteInput` with empty IDs/names, and `requireSheetSelector` rejects before Execute runs. Thus an otherwise valid `+chart-delete` against a one-sheet workbook still fails with `specify at least one...`. Use the preflight/deferred-selector validation for this factory; the update factory and several hand-written merge/resize/range validators have the same mismatch.

shortcuts/sheets/lark_sheet_object_crud.go
```diff
@@ -1418,12 +1428,90 @@ func normalizeCondFormatStyle(style map[string]interface{}) {
...
+	raw, present := style["font"]
+	if !present {
+		return
+	}
+	var bold, italic bool
+	switch v := raw.(type) {
+	case map[string]interface{}:
+		for key, val := range v {
+			on, readable := val.(bool)
+			if !readable || !on {
```
chendaxin-tk: [P1] Do not silently discard unrecognized members of a composite font object. For `font:{"bold":true,"underline":true}`, this loop recognizes `bold`, ignores `underline`, and later replaces the whole object with schema-valid `font:"bold"`; the requested underline disappears without an error. Only replace the object when every present member is losslessly understood, or map the remaining effects (for example to `text_decoration`) before emitting the enum.

shortcuts/sheets/style_vocab.go
```diff
@@ -664,9 +791,43 @@ func normalizeCellsFlagValue(v interface{}) interface{} {
...
+		}
+		width = max(width, len(row))
+	}
+	if width == 0 {
+		return
+	}
+	for i, rowRaw := range rows {
+		row, _ := rowRaw.([]interface{})
+		for len(row) < width {
+			row = append(row, map[string]interface{}{})
```
chendaxin-tk, line 834: [P1] Check the rectangularized cell count before materializing every missing cell. A small ragged JSON payload with one 10,000-cell row followed by 9,999 empty rows reaches this loop and allocates roughly 100 million empty maps (plus the interface slots) before any matrix-budget validation, which can OOM-kill the CLI. Compute `len(rows) * width` with overflow/cap checks first, then reject oversized matrices before padding.

shortcuts/sheets/style_vocab.go
```diff
@@ -1017,5 +1235,108 @@ func foldBorderFamilyAliases(in map[string]interface{}, path string) error {
...
+	soft func(side, attr string, val interface{}),
+) error {
+	word, isStr := v.(string)
+	if !isStr {
+		return setSideLoose("all", v, "border_type")
+	}
+	key := strings.ToUpper(strings.ReplaceAll(strings.TrimSpace(word), "-", "_"))
+	key = strings.ReplaceAll(key, " ", "_")
+	if side, ok := borderTypeSideValues[key]; ok {
+		soft(side, "style", "solid")
```
chendaxin-tk, line 1354: [P1] Apply the explicit line attributes to the side selected by `border_type`. Earlier normalization turns `border_style:"dashed"` into an all-sides dashed entry; then `border_type:"LEFT_BORDER"` adds a solid left entry here. Expansion therefore yields left=solid and top/bottom/right=dashed, rather than the unambiguous requested dashed left border. The side selector and line attributes need to be combined before the all-sides shorthand is expanded.

shortcuts/sheets/lark_sheet_write_cells.go
xiongyuanwen-byted: Keeping the widening, and recording why rather than quietly disagreeing.

The extent used to be enforced in one direction only: a payload smaller than the stated range was narrowed to it, a larger one was rejected. That asymmetry is what this branch removed, after weighing exactly the case you name. Three things decided it. The command overwrites non-empty cells by default (`--allow-overwrite`), so "cells the caller did not name" are already writable on the normal path. A bare single-cell range has always been an anchor, so the range was never a strict assertion. And the narrowing direction — which touches the same contract — was already silent before this branch.

What it costs is real and is the sentence I would keep in the review: a payload built one row short now writes quietly where it used to fail here. That is why `fitCellsRange` reports both directions in the result's `warnings`, naming the shipped range and saying which way it moved, and why the widening note points at `--allow-overwrite=false`.

The alternative you describe — bare `A1` stays an anchor, an explicit extent must match or fail — is the pre-branch behaviour, and it accounts for 21425 rejections in the 09-04..07 window across both directions. I do not think that trade is mine to reverse unilaterally in either direction, so I have left the behaviour as it is and flagged this thread to the owner. If the call goes the other way, the change is one branch in `checkCellsPayloadShape` plus the tests that pin it.

xiongyuanwen-byted:
Pushed `7ef84f45`, which takes 27 of the 28 findings. Each one was reproduced against a built binary before the fix and re-run after; `TestReviewRegressions` in `param_tolerance_round2_test.go` pins the ones that were behavioural, and every fix was also exercised end to end against a spreadsheet created for the purpose, reading the cells back.

**The panic.** Confirmed exactly as described — `data:[[1,2],[3]]` took the process down on both the no-columns and the widened-columns paths, because `fitColumnsToRows` widens the column list after `padShortRows` has run and both the validator and the writer index a row by column position. Both paths re-pad against the count they settle on.

**Ordering.** `wrapBareListValue` now runs before the loose-JSON repair; `normalizeCondFormatFontValue` before the flat-word loop; a `border_type` side selector now carries the line attributes the caller spelled out, instead of leaving a dashed box with a solid left edge.

**Swallowed failures.** A workbook-structure read that fails on authentication, permission or transport used to surface as "specify at least one of --sheet-id or --sheet-name" from the selector resolver and as "unsupported range form" from the style bounder. Both propagate now. A sheet-prefixed whole-column range is split before the bounder's patterns are applied.

**Rewrites that could pick a winner.** Two stray dtypes keys folding onto one column, and positional dtypes over a repeated heading, both refuse rather than resolve by map order. A font object carrying an effect this enum has no room for is left for the schema instead of being replaced. A single-quoted `\n` is decoded rather than passed through as a literal backslash and n; an escape that cannot round-trip leaves the payload unrepaired.

**Reach.** The deferred-selector change only took effect where `Validate` ran through `validateViaInput` — the object-CRUD factories and the three chart commands validate through their own callbacks and are now wired to the same helper. `+sheet-delete` is excluded from resolution entirely: it takes the whole sub-sheet and has no narrower target to state.

**Resource and protocol.** The padding cap is checked against the padded size before the rectangle is materialised; `--values` decodes once and repairs only on failure; the trailing-comma trim truncates its buffer rather than rebuilding the prefix per level; `liftFlatCellsRow` no longer trips `asasalint`.

Three test assertions that could not observe what they pinned were tightened: the abbreviation guard now reads the clause rather than the whole hint, the repair comparison is byte-for-byte, and the border cases assert all four sides.

One finding is answered rather than applied — the explicit-extent widening, replied to in its own thread. It is a decision that was taken deliberately with the numbers in front of it, and reversing it is the owner's call, not mine.

xiongyuanwen-byted:
Pushed `78c9e9f0` for the two findings from the re-review, both of them consequences of the previous round.

Deferring the missing selector in `Validate` only helps if execution answers it, and the object-create factory reads its selector from flags the spec names itself, so it could not go through `resolveSheetSelectorExec` and was still handing the empty pair to the input builder. The workbook lookup is split out (`resolveOmittedSheetSelector`) so both readers share it. Verified against a one-sheet workbook: `+cond-format-create` with no selector now resolves and applies.

The side selection replaced the selected side's map rather than merging into it, so `border_left_color` next to `border_style` and `border_type:"LEFT_BORDER"` lost the colour. Only the attributes that side lacks are copied from `all` now; read back from a live sheet, the left edge carries dashed and #FF0000 together.

`TestObjectCreate_RequiresSheetSelector` moves to `--dry-run`, which is where that rejection still belongs.

Both are covered in `TestReviewRegressions`. Everything else from the first pass stands; the one finding I answered rather than applied is still the explicit-extent widening, in its own thread.

shortcuts/sheets/lark_sheet_table_io.go
```diff
@@ -590,6 +686,141 @@ func (in *tableSheetIn) normalize(idx int) (tableSheetSpec, error) {
...
+		padShortRows(s)
+		return
+	}
+	for len(s.Columns) < widest {
+		s.Columns = append(s.Columns, tableColumnSpec{})
+	}
+	// Widening the column list leaves every SHORTER row short of it, and the
+	// writer indexes a row by column position. padShortRows ran in normalize,
+	// before this; it has to run again against the count this settled on.
+	padShortRows(s)
```
fangshuyu-768, line 827: [P1] Check the rectangularized cell count before padding. `fitColumnsToRows` calls `padShortRows` before `validate` reaches `checkCellBudget`. A compact payload with one 8,000-cell row followed by 2,000 empty rows therefore expands to about 16 million interface slots before the 1M-cell limit can reject it; larger shapes can OOM the CLI. Precompute the total projected rows × widest-column count across all sheets using overflow-safe arithmetic, reject over-budget payloads before either padding path, and only then materialize the rectangle. Please cover both the no-columns and widened-columns forms.

caojie0621 (approved):

PR #2723, opened 2026-09-14
feat: make Meta API methods discoverable from help and schema

heshuaichen-dev:
## Summary

An agent that knows a Meta API method exists still could not reliably run it. Domain help stopped at resource rows that named no methods, `schema` answered either one method or a dump too large to read, method help never showed the request body, and a mis-separated path was rejected with no correction. Evaluation recorded the path-separator rewrite done correctly zero times across six rounds.

This PR makes the method surface navigable in one hop at every depth: `schema` renders by target depth (service index, method index, full envelope), domain help lists methods flattened in their runnable form, method help carries the request-body contract, and every rejection answers with the form that runs.

## Changes

- Render `schema` by target depth and add `--jq` in `cmd/schema/schema.go`; index shapes and sanitization live in `internal/schema/index.go` and `internal/schema/types.go`
- Flatten domain help to a method-level listing and embed the request-body contract in method help, in `cmd/service/affordance.go`, `cmd/service/bodyhelp.go` and `cmd/service/service.go`
- Answer mis-separated and unresolvable paths with the runnable command form, in `cmd/root.go`, `cmd/schema/schema.go` and `internal/apicatalog/path.go`
- Keep the bare service index at a no-shard-parse cost, gated on strict mode and plugin-declared restrictions, in `cmd/schema/schema.go` and `cmd/build.go`
- Point skills and templates at method help before the full schema, in `skills/lark-*/SKILL.md` and `skill-template/`
- Pin output budgets and the assembly scope in `cmd/output_budget_test.go` and `cmd/schema/schema_assembly_scope_test.go`

## Test Plan

- [x] `make unit-test` — 129 packages, race enabled
- [x] `go vet ./...` and `gofmt -l .` clean; `go mod tidy` leaves `go.mod` / `go.sum` unchanged
- [x] `golangci-lint run --new-from-rev=origin/main` (v2.1.6) — 0 issues
- [x] `go test ./...` green, including `tests/plugin_e2e` concealment and degraded-subsystem paths
- [x] `node scripts/skill-format-check/index.js` passed for the `skills/lark-*` edits
- [x] No dry-run or live E2E added: the diff touches no `shortcuts/`, so the shortcut E2E matrix in AGENTS.md does not apply. The new behavior is pinned by unit tests in `cmd/schema`, `cmd/service` and `internal/schema`, by output-budget assertions in `cmd/output_budget_test.go`, and by `cmd/schema/schema_assembly_scope_test.go`, which fails if the bare service index starts parsing service shards
- [x] Manual check against a real tenant: `lark-cli schema`, `lark-cli schema im`, `lark-cli schema im chat.members get`, `lark-cli calendar --help`, `lark-cli --help drive`, `lark-cli help calendar event.attendees.batch_delete`

## Related Issues

N/A

## Summary by CodeRabbit

- **New Features**
  - Domain help now lists runnable API methods in a flattened format.
  - Method `--help` includes copyable request-body templates and field guidance.
  - `schema` provides structured service and method indexes, full contracts, and `--jq` filtering.
  - Schema output now includes runnable command forms.

- **Improvements**
  - Invalid or misspelled commands provide structured errors and actionable corrections.
  - Hidden commands remain usable while keeping help listings focused.
  - Resource descriptions are clearer and sanitized.

- **Documentation**
  - Updated English and Chinese guides and skill instructions to prioritize method help, using schema for deeper contract details.


cmd/schema/schema.go
```diff
@@ -6,7 +6,9 @@ package schema
 import (
 	"context"
 	"errors"
+	"fmt"
```
heshuaichen-dev, line 9: ✅ Fixed in f9559648.

Confirmed reproducible before the fix — `schema im.chat<U+202E>members.get` echoed the raw bidi bytes straight into the envelope. `safeSeg` was applied in the shortcut and all three `ErrService` paths but not in `ErrResource` / `ErrMethod` / `ErrPath`; all three now go through it.

Added `TestResolveError_SanitizesEveryRejectedSubject` covering bidi and ANSI in both the resource and method positions. It passes its parts pre-split (the way `apicatalog.ParsePath` hands them over) because a single dotted argument resolves as a service name and would have exercised the branch that was never unsafe — and it asserts which branch each case reaches, so it cannot drift back onto an already-sanitized path and still pass. Verified revert-failing against each branch independently.

cmd/service/service.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/larksuite/cli/internal/output"
 	"github.com/larksuite/cli/internal/recovery"
 	"github.com/larksuite/cli/internal/registry"
+	"github.com/larksuite/cli/internal/schema"
```
heshuaichen-dev, line 26: ✅ Fixed in 68123bf0.

Confirmed as a regression this branch introduced, measured against `origin/main`:

```
origin/main:  lark-cli __complete im ""  → 14 native resource entries
before fix:   same command                → 0
after fix:    same command                → 14 (byte-identical to main)
```

The domain command now carries a `ValidArgsFunction`. Cobra appends its results to the visible children, so the 24 `+shortcut` completions are unaffected, and the second hop was never broken — once a resource is named, Cobra finds that hidden command and completes its own method children normally.

A group is offered only when a method under it is still reachable, mirroring `flattenedAPIMethods`, which skips a leaf a policy layer took away; completing a group whose every method is concealed would hand back a path that rejects even `--help`.

skills/lark-task/SKILL.md
```diff
@@ -19,14 +19,14 @@ metadata:
 1. 先将用户意图与下方 Shortcut 表精确匹配。只有表中明确列出的 shortcut 才可直接选择；参数不确定时读取对应 reference 或运行该 shortcut 的 `--help`。
 2. 没有精确匹配、或无法确认当前版本是否支持时，先运行 `lark-cli task --help`，以当前 CLI 输出的命令列表为准。
 3. help 中存在匹配 shortcut 时，使用 help 列出的完整 shortcut token（例如 `+create`）运行 `lark-cli task <shortcut> --help`，再按真实 flag 执行。
-4. help 中没有匹配 shortcut 时，不得尝试相似的 `+<verb>`；从 help 中选择原生 resource，运行 `lark-cli task <resource> --help` 确认 method，再运行 `lark-cli schema task.<resource>.<method>` 获取参数结构，最后调用 `lark-cli task <resource> <method> ...`。
+4. help 中没有匹配 shortcut 时，不得尝试相似的 `+<verb>`；从 help 中选择原生 resource，运行 `lark-cli task <resource> --help` 确认 method，再运行 `lark-cli task <resource> <method> --help` 看参数与请求体骨架（只有深层嵌套结构或需要完整契约时才查 `lark-cli schema task.<resource>.<method>`），最后调用 `lark-cli task <resource> <method> ...`。
```
heshuaichen-dev, line 22: ✅ Fixed in 9a8da5e7.

You were right that the two sections disagreed, and there were three spots, not one: step 4 of the workflow, the tasklist-search note, and the worked example at the bottom all still said to run `schema` before every native call. The API resources section is now the only rule — method `--help` first, `schema` only for deep nesting or the full contract.

I checked the other seven skills this PR touches for the same contradiction; `lark-task` was the only one carrying those instructions.

.gitignore
```diff
@@ -50,6 +50,14 @@ cmd/api/download.bin
 app.log
 /sidecar-server-demo
 /server-demo
+# `make examples-build` runs `go build` with no -o, so each example lands in the
```
heshuaichen-dev, line 53: ✅ Fixed in 5777663b — added root-anchored `/audit-observer.exe` and `/readonly-policy.exe` alongside the extensionless entries.

cmd/service/service.go
heshuaichen-dev, line 26: ✅ Fixed in 80c4ad23 — the completion is now attached at every resource level, not just the domain.

Two corrections to the finding, since they change what the fix had to be:

`chat.members` is a single flat command name, not `chat` → `members`; `lark-cli <domain> chat <TAB>` has no `chat` group to land on. And no committed service nests resources at all — parsing all 15 `internal/registry/catalog/services/*.json`, every resource entry carries only `methods`, so the scenario is unreachable with today's catalog.

The structural point still holds, though: `meta.Resource` carries its own `Resources`, `apicatalog.walkResources` recurses through them, and `registerService` builds and hides one command per `ResourcePath` segment — so completing only at the domain would leave intermediate groups uncompletable the moment a catalog does nest. Since no real service exercises it, the test supplies the shape with a synthetic `meta.Service` and asserts both hops; verified revert-failing. Real-catalog completion is unchanged at 14 entries for `im`.

heshuaichen-dev:
## The `e2e-live` failure is pre-existing on `main`, not introduced here

`e2e-live` — and the `results` gate that aggregates it — fails on a single test, `tests/cli_e2e/base TestBaseDashboardBlockNPSWorkflow`. It is counted three times because CI re-runs it twice. The other 24 checks pass.

It is a server-side rejection rather than a build or logic error:

```
code 800010701 "Invalid discriminator value."
fields[1].type: Use one of these allowed discriminator values:
"text", "number", "select", "datetime", ... "checkbox", "button"
```

The test posts `[{"name":"Name","type":"text"},{"name":"Satisfaction","type":"rating"}]`, and `rating` is not in the set the API currently accepts.

Why this is unrelated to this PR:

- This PR changes no base code or fixtures — the diff against `main` touches nothing under `shortcuts/base/` or `tests/cli_e2e/base/`.
- The same job fails on this PR's merge base `39aaf9fc`, on the same test (run `34575777411`).
- `main` has been red on its last three runs, since 2026-09-11.
- The only commit between the last green run (`b8b21da3`) and the first red one is `b67a4e85` (#2562), which is also the commit that added both this test and the `rating` value.

I have deliberately left the test alone. It belongs to the base domain and to #2562, and choosing a replacement field type is a product decision about what the NPS fixture should assert — not something to settle inside a help/schema change.

Suggest reading this check as blocked on a `main` fix rather than on this PR.

